PR #1, opened 2017-01-01
Url fix

FransBouma:
source/d3d11 pointed to d3d10 in readme. 

Thanks for opening the source of this great project!


crosire:
Whoops :)

PR #4, opened 2017-01-02
Two fixes: mouse cursor change on splitter bar and node expansion bug on techniques list

FransBouma:
Two changes/fixes:

- When the mouse is now hovered/above the splitter bar, the mouse cursor is changed to the vertical resize cursor. This makes it easier to see when the splitterbar is active. I tried several things to make it better, but it appears the techniques tree has a padding below it of several pixels, which is unavoidable it seems. This makes it look as if the splitter bar starts right below the horizontal bottom border of the techniques tree, but that's not the case. This small change makes it easier: the user can now see where the splitter bar is and makes moving the splitter bar easier

- Fix for an issue where a technique was moved due to reordering with the mouse: If a technique was reordered in the list in such a way that the filename (e.g. Dof.fx) appeared twice or more times in the tree, the nodes all got the same ID. This had the effect that only the first node with the filename 'Dof.fx' was expandable in the top pane with the techniques: all other nodes with 'Dof.fx' were not expandable. This change fixes that by assigning a unique ID to the treenodes containing the filename. 

The ordering goes wrong in some occasions with drag/drop, likely due to some issue inside imgui. perhaps it's a good idea to add a separate tab with only the order of the techniques, so it's separated from the technique tree itself as it's a separated concern. But that's for another day :)


FransBouma:
The treenode issue popped up because I wanted to define a specific order of the shaders. After looking into it a bit more, it looks like that order isn't preserved between sessions (it is within a session when you close / reopen the UI). Is that the case? I couldn't see a way to specify it in the ini files as well, so my guess is you can re-order it through drag/drop but you have to do that every time.

crosire:
I'm not sure I understand the question. What shader order do you refer to? I assume you mean the technique order. The reordering for that one is done in https://github.com/crosire/reshade/blob/master/source/runtime.cpp#L1992. The `save_preset` call shortly after makes sure that change is saved to the INI as well and it should therefore be preserved between sessions. The order is loaded from an INI file here: https://github.com/crosire/reshade/blob/master/source/runtime.cpp#L827 (beautiful code, I know, but it does the job :innocent:).

FransBouma:
Yeah the technique order, sorry about that. Ok, then I think something's off, as I have for testing the deband.fx and dof.fx installed, moved one of the dof techniques above the deband technique, but after restarting the game, the order was back to first deband, and then the dofs from dof.fx. I couldn't find any order specification in the ini files as well. Will see if I can find out what's going on.

crosire:
The "Techniques" value in the INI file implicitly specifies the order.
E.g. `Techniques=DOF,Deband,SomeOtherDOF,EvenMoreDOF` means the "DOF" techniques comes first, then "Deband", etc.

FransBouma:
> However, due to this, the order of disabled techniques is not preserved! It only works on active techniques.

Ah, then I understand why the ini file didn't contain the order I specified.

IMHO this is rather confusing. The thing is: you likely want to enable techniques with a key sometime, or at some point through the UI. However doing so then also requires you to reorder the effect in such a way that it is ran at the right order of the other effects. Say I have deband, smaa, lumasharpen active. Then I want to enable DOF for some screenshot. I have to re-order it first to make sure it's run before smaa and after lumasharpen. That doesn't make sense, sorry. Perhaps for techniques which are always active, but not all techniques are always active: for screenshotters, we enable alot of techniques for a shot which are disabled afterwards as playing a game requires different effects than a screenshot does ;)

In v2 I have reordered the techniques in a particular order which works (for me at least) very well. That's completely separate from when an effect is active or not. With the preserved order, I can enable/disable effects at will and not have to worry about the order, as that's already been setup, once. 

Regardless, are the two fixes/changes OK? Should I make the ordering persistent regardless of the effect being active? (I dont' see why that would hurt anyone, i.e. it won't break presets, as all techniques will be sorted below the active ones after the change anyway)

crosire:
Yeah, the pull request is fine. The ordering could use some further discussion, but that's topic for another pull request or issue.

FransBouma:
> The ordering could use some further discussion, but that's topic for another pull request or issue.

Agreed. Will think about it some more and open an issue when I have it thought out properly. :) Cheers

PR #9, opened 2017-01-09
Add basic config editor to end of setup

mgziminsky:
Added a very basic config editor to the end of the setup to make it easier to select an existing preset and the search paths.


setup/Wizard.xaml.cs
```diff
@@ -26,7 +26,9 @@ public partial class WizardWindow
 		PEInfo _targetPEInfo = null;
 		string _tempDownloadPath = null;
 
-		public WizardWindow()
+        private string ConfigFilePath => Path.ChangeExtension(_targetModulePath, ".ini");
```
crosire: There perferably should be tabs instead of spaces here.

setup/ReShade Setup.csproj
```diff
@@ -37,6 +39,12 @@
     <Prefer32Bit>false</Prefer32Bit>
   </PropertyGroup>
   <ItemGroup>
+    <Reference Include="Microsoft.WindowsAPICodePack, Version=1.1.2.0, Culture=neutral, processorArchitecture=MSIL">
```
crosire: I'm not all too happy about adding additional dependencies and would much prefer a pure WPF implementation.

setup/Wizard.xaml.cs
mgziminsky: Sorry, will fix those

setup/ReShade Setup.csproj
mgziminsky: I completely understand, and I spent the better part of a day trying to find a native solution. Unfortunately, WPF does not provide a folder selection dialog, and the FolderBrowserDialog from win forms doesn't support multi-select. I could build a simple custom listview dialog, but that's a lot more work for a bad end result.

Everything I read online said to either roll your own, or to use that library. Given how much better the version provided by the dependency is, I opted for that. If you are strongly against adding dependencies, I'll make a simple custom one, but it will be a bare minimum implementation.

FransBouma: Does this help? http://stackoverflow.com/a/12909759/44991

mgziminsky: Unfortunately no. The file dialog isn't the problem in this case, it's the lack of a folder dialog. The code sample in that link is for enabling multiple select in file dialogs (also it's using win forms, but it's basically the same in WPF anyway), although it does link to examples for building your own custom folder dialog. 

The point I was trying to make in my previous comment was that I'd prefer not to build a custom one, although I will if insisted, since the result will be significantly worse than the native version exposed by the dependency. It would also add a non-trivial amount of additional code to the project that would need to be maintained should anybody ever decide to expand on the editor.

setup/ConfigEditor.xaml.cs
```diff
@@ -0,0 +1,126 @@
...
+		public ConfigEditor(string configPath)
+		{
+			InitializeComponent();
+			_configFilePath = configPath;
+
+			_load();
+			Closing += (object sender, CancelEventArgs args) => _save();
+		}
+
+		private void _writeValue(string key, string value)
```
FransBouma: methods in .NET should use PascalCasing for their names. starting them with '_' is not really great. Perhaps a thing to look into next time ;)

setup/ReShade Setup.csproj
FransBouma: Ah I see. Indeed a problematic situation. MS released the code for that on their (now defunct) code.msdn.microsoft.com site, and someone copied it to github. See: https://github.com/aybe/Windows-API-Code-Pack-1.1 There are other repos with the same code. It could help with adding the dialog as source to the project so no external dependency is needed :)

mgziminsky: I'll see about copying the relevant source into the project instead, but is that really better? The main solution already has 4 git submodules linked into it, is adding a couple nuget dependencies really that much worse.

I can understand the desire to avoid dependencies for simple things that can be easily done without them, but that's really not the case here. All of the other options are quite a bit more work and/or add a lot of unimportant bloat to the project.

setup/ConfigEditor.xaml.cs
mgziminsky: I'll go back through and fix the styling. It's been a while since I've worked with .NET and just went with my general preferences since there wasn't a style guide.

It'd be nice if somebody could add a basic style guide to the README, or somewhere, documenting at least indentation, naming conventions, and braces placement.

setup/ReShade Setup.csproj
FransBouma: I didn't want to imply you had to copy the whole windows api code pack :) Just the class which produces the folder browser, that's all. I don't think that's a lot of bloat. The thing with adding a nuget package is that the dlls have to be shipped with the tool, or you have to ilmerge them together into the exe. It's currently a single exe that's shipped, I think with the dlls added to that you need extra instructions how to deal with the zip file etc. But I'm not the person who has the end-say over this, so if crosire is fine with it, I don't mind, I was merely pointing into the direction where a solution might be found ;)

setup/ConfigEditor.xaml.cs
FransBouma: No biggie, it was just something I noticed. No need to spend a lot of time on :)

setup/ReShade Setup.csproj
mgziminsky: Yep, I understood what you were suggesting. Also, I totally understand the desire to keep everything a single exe, which is what the additional nuget dependency in the second commit is for :) It handles merging the dependent DLLs into the target executable automatically as part of the build.

mgziminsky:
@crosire, can you give your final say on what you want me to do with this in order to get it merged?

After trying out the different options discussed earlier, I still feel using the NuGet dependencies is the simplest and cleanest option.
I can add the dependency sources directly to the project instead, but I need the full source for the Core and Shell modules, I can't just pull in a couple files. I could potentially link them in as submodules, but I don't see why having the source directly would be better either way.
The one thing I really don't want to have to do though, is build a custom folder dialog from scratch.

crosire:
I don't really know. I'll put this on the back burner for a little while and decide later.

mgziminsky:
So, this has been sitting here for over a year now. Any chance of getting it merged soon?

crosire:
Yeah. Thinking about it again, adding additional dependencies won't be the end of the world. I'd appreciate if you could rename `common.props` to `Common.props` though (see `git mv`) to match naming convention in that directory.

PR #10, opened 2017-01-09
Use ISO 8601 for timestamps in the logfile

bacondither:


crosire:
Looks reasonable. Thanks!

PR #11, opened 2017-05-06
Fix for games that use DXGI Present testing

Kaldaien:
Some newer games will issue `IDXGISwapChain*::Present* (...)` using the flag `DXGI_PRESENT_TEST`.

When these calls are encountered, they are not actually presenting anything. This is a status inquiry similar to D3D9's cooperative level to check for device reset, and may be used to test swapchain latency, a partially covered window and various other  things prior to drawing anything.

These calls ***should not*** be passed through ReShade's render pipeline, they should be ignored and passed directly to the DXGI runtime. Framerate counting, performance and stability will suffer in games like Deus Ex: Mankind Divided if software fails to ignore these calls.

### DIFF
--- a/.vs/ProjectSettings.json
+++ b/.vs/ProjectSettings.json
@@ -1,0 +1,3 @@
+{
+  "CurrentProjectSetting": "x86-Debug"
+}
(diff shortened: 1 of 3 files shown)


crosire:
Good catch!

PR #12, opened 2017-05-21
Fix wglCreateContextAttribsARB flow in debug builds

CookiePLMonster:
This fixes a startup issue in KOF XIV when ReShade is compiled in Debug as described here:

https://reshade.me/forum/troubleshooting/3194-the-king-of-fighters-xiv-cbt-black-screen-issue

technically, `attribs` could stay as an array of 7 entries for Release builds but IMO it doesn't matter.

### DIFF
--- a/source/opengl/opengl.cpp
+++ b/source/opengl/opengl.cpp
@@ -3145,7 +3145,10 @@ HGLRC WINAPI wglCreateContextAttribsARB(HDC hdc, HGLRC hShareContext, const int 
 
 	int i = 0, major = 1, minor = 0, flags = 0;
 	bool core = true, compatibility = false;
-	attribute attribs[7] = { 0, 0 };
+#ifdef _DEBUG
+	bool debugFlagAdded = false;
+#endif
+	attribute attribs[8] = { };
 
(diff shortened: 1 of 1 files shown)


crosire:
Could you check if things still work with latest commit e108b00721f68f64bb6f23ebcb0cb2f1f79494ce? According to the specs (https://www.khronos.org/registry/OpenGL/extensions/ARB/WGL_ARB_create_context.txt) you can specify the same attribute multiple times in the array in which case the last value specified is used. That's why technically things should have worked already. If that commit breaks things again than that part of the specs is apparently not working in the driver you use. Then it would be best to not skip the `WGL_CONTEXT_FLAGS_ARB` and `WGL_CONTEXT_PROFILE_MASK_ARB` values while looping over the existing attributes and only write them afterwards instead of writing them to the array two times.

CookiePLMonster:
Yes, it still works. So looks like the issue was about lack of null terminator in the attribs array. It's required as specified by the very website you linked:

`The list consists of a sequence of <name,value> pairs terminated by the value 0.`

I figured not sending the same attribute twice would do no harm, but as you pointed out it really shouldn't matter.

crosire:
Great! Yeah, looks like I forgot about the null-termination when I wrote that code.

PR #13, opened 2017-05-28
Fix for KOF CBT black screen + more

CookiePLMonster:
I changed quite a few things:

1. Changed uniform storage variables to GL*ptr variables - they are created from natural word size variables (used to be truncated) and used as parameters where GL functions expect GLptr variables - so figured this will make 32-bit and 64-bit OpenGL implementation more consistent.
2. Added more move semantics.
3. Merged and altered a fix from @bacondither, as mentioned in KOF topic on ReShade forums:
https://github.com/bacondither/reshade/commit/81c607e49269f843c7b02f494b62de0e9a9d9463
I preserved the check, albeit with fixed conditions.

I also planned to port some of the changes made in d3d8to9 over the last couple of months - mainly interlocked refcounting. However, I guess I'll push this separately given 3. is a quite major bugfix, making ReShade usable in more games.

Cheers!

### DIFF
--- a/source/filesystem.hpp
+++ b/source/filesystem.hpp
@@ -23,7 +23,7 @@ namespace reshade::filesystem
 	public:
 		path() { }
 		path(const char *data) : _data(data) { }
-		path(const std::string &data) : _data(data) { }
+		path(std::string data) : _data(std::move(data)) { }
 
 		bool operator==(const path &other) const;
 		bool operator!=(const path &other) const;
--- a/source/opengl/opengl_effect_compiler.cpp
+++ b/source/opengl/opengl_effect_compiler.cpp
@@ -428,7 +428,7 @@ namespace reshade::opengl
 
 			glBindBuffer(GL_UNIFORM_BUFFER, previous);
 
-			_runtime->_effect_ubos.push_back(std::make_pair(ubo, _uniform_buffer_size));
+			_runtime->_effect_ubos.emplace_back(ubo, _uniform_buffer_size);
 		}
 
 		return _success;
--- a/source/opengl/opengl_effect_compiler.cpp
+++ b/source/opengl/opengl_effect_compiler.cpp
@@ -2226,7 +2226,7 @@ namespace reshade::opengl
 
 		auto &uniform_storage = _runtime->get_uniform_value_storage();
 
-		if (_uniform_storage_offset + _uniform_buffer_size >= uniform_storage.size())
+		if (_uniform_storage_offset + _uniform_buffer_size >= static_cast<ptrdiff_t>(uniform_storage.size()))
 		{
 			uniform_storage.resize(uniform_storage.size() + 128);
 		}
--- a/source/com_ptr.hpp
+++ b/source/com_ptr.hpp
@@ -20,6 +20,11 @@ public:
 	{
 		reset(copy._object);
 	}
+	com_ptr(com_ptr<T> &&move) : _object(nullptr)
+	{
+		std::swap(_object, move._object);
+	}
+
 	~com_ptr()
 	{
 		reset();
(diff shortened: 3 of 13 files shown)


source/filesystem.hpp
```diff
@@ -23,7 +23,7 @@ namespace reshade::filesystem
 	public:
 		path() { }
 		path(const char *data) : _data(data) { }
-		path(const std::string &data) : _data(data) { }
+		path(std::string data) : _data(std::move(data)) { }
```
crosire: I'm curious as to how these are different?

source/opengl/opengl_effect_compiler.cpp
```diff
@@ -2226,7 +2226,7 @@ namespace reshade::opengl
 
 		auto &uniform_storage = _runtime->get_uniform_value_storage();
 
-		if (_uniform_storage_offset + _uniform_buffer_size >= uniform_storage.size())
+		if (_uniform_storage_offset + _uniform_buffer_size >= static_cast<ptrdiff_t>(uniform_storage.size()))
```
crosire: I don't see the reasoning behind this. `ptrdiff_t` is signed, `size_t` is not and the overall size can't be signed either, so this only crates problems but doesn't solve any?
EDIT: I see this was done to not run into a warning due to signed/unsigned mismatch on the comparison. It would problably make more sense to make `_uniform_storage_offset` and `_uniform_storage_size` of type `size_t` instead? Because they are used in various `size_t` expressions already anyway.

source/filesystem.hpp
CookiePLMonster: The latter is move semantics friendly. Let's compare three cases:

1.
```c++
std::string string = "FOO";
path myPath(string);
```
Before the change, `path` constructor does one copy (from `data` to `_data`). After the change, `path` constructor does one copy (to `data`) and one move (from `data` to `_data`). Move constructing overhead is most likely negligible, plus it's not unlikely the compiler will notice it and optimize the move out.

2.
```c+++
std::string string = "FOO";
path myPath(std::move(string));
```

Before the change, there is no constructor which takes an rvalue so it degrades to lvalue and becomes identical to case 1. - thus doing a copy. After the change, `data` is move constructed and so is `_data` - so we get to use move semantics properly.

3.
```c++
path myPath(std::string("FOO"));
```

Before the change, the temporary is constructed, copied (from `data` to `_data`) and destructed. After the change, the temporary is constructed and moved.

Of course, another way to solve it is to just add another constructor taking `std::string&&` instead of trying to unify the two. What I learnt recently though is that it's perfectly reasonable to pass arguments by value instead of const reference when the constructor is supposed to copy them. It's especially useful if you have multiple arguments and you'd like to accept them as either lvalue or rvalue - instead of resorting to hacks or creating overloads for every possible combination of parameters, you just pass them by value and move to class fields.

source/opengl/opengl_effect_compiler.cpp
CookiePLMonster: Sure thing. I did this just to suppress the warning and didn't check if either offset or size are used as if they had a sign or not (fairly sure one of them gets initialized to -1, though!). You're natually much more familiar with this code so if you think making those two unsigned instead is a better idea, I have no objections.

source/filesystem.hpp
crosire: Ah, right, I only saw the pass-by-value, but didn't realize that works out with move semantics anyway.

source/opengl/opengl_effect_compiler.cpp
crosire: The -1 initialization is a similar but different field in `runtime_objects.hpp`. I don't know why I made them `int`s, but I guess I was just testing this out and later kept the code without refactoring it properly.

PR #14, opened 2017-05-29
Only load enabled effects in performance mode

mgziminsky:
Significantly improves startup time in performance mode with a full shaders checkout.
This makes it more feasible to share a single shaders directory for all games since unused effects won't get loaded.

I did change the preset file format slightly though, so newer versions wouldn't properly load with an old preset file. The solution would be to simply turn off performance mode once, re-enable the desired effects, then turn performance mode back on.

I'll remove the cleanup commit if you want. It was originally part of some changes that were relevant to the PR, but the final solution didn't use them. I left it in because I still think it improves the code, but it doesn't really matter.

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -426,10 +426,36 @@ namespace reshade
 
 		_effect_files.clear();
 
-		for (const auto &search_path : _effect_search_paths)
+		// Only load active effects
+		if (_performance_mode && _current_preset >= 0)
 		{
-			const auto matching_files = filesystem::list_files(search_path, "*.fx");
-			_effect_files.insert(_effect_files.end(), matching_files.begin(), matching_files.end());
+			ini_file preset(_preset_files[_current_preset]);
+			const auto techniques = preset.get("", "Techniques").data();
+			LOG_INFO() << "Loading " << techniques.size() << " active techniques";
+			for (const auto &technique : techniques)
+			{
+				LOG_INFO() << "Searching for technique file: " << technique;
+				for (const auto &search_path : _effect_search_paths)
+				{
+					auto effect_file = search_path / technique;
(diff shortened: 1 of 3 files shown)


source/runtime.cpp
```diff
@@ -873,10 +904,10 @@ namespace reshade
 		{
 			if (technique.enabled)
 			{
-				technique_list.push_back(technique.name);
+				technique_list.push_back(technique.effect_filename);
```
crosire: Unfortunately, that's not going to work, since there are effects with multiple techniques (the reason they are grouped by filename in the first place). These two lines should be kept as they are, but in the loading routine one might check if a effect file with the same filename as the technique exists and only load those that are in the list. This still falls apart for techniques which are named different from the file, but should such a technique be encountered (which is probably not the common case), ReShade could just fall back to load everything.

crosire:
The cleanup commit looks fine, I like it.

source/runtime.cpp
```diff
@@ -766,18 +763,16 @@ namespace reshade
...
-			search_path = filesystem::absolute(search_path, parent_path);
-		}
-		for (auto &preset_file : _preset_files)
+		auto to_absolute = [&parent_path](decltype(_preset_files) &paths)
 		{
-			preset_file = filesystem::absolute(preset_file, parent_path);
-		}
+			for (auto &path : paths)
+			{
+				path = std::move(filesystem::absolute(path, parent_path));
```
CookiePLMonster: std::move does nothing here, since you're assigning a temporary.

source/runtime.cpp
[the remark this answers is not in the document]
crosire: Seems I had forgotten to click the submit button on the review *facepalm*

crosire:
Nice!

PR #22, opened 2017-08-14
Change default `active` color to contrast more

RobotCaleb:
I'm color blind and was having a lot of trouble telling whether a checkbox was checked.
This is a contrast tweak.


crosire:
Thanks!

PR #38, opened 2017-12-28
Better depth buffer detection / drawcall tracking for D3D11

FransBouma:
This PR contains a better depth buffer detection system and drawcall tracking. It is designed to work with normal immediate-context focused engines and also for multi-threaded / deferred engines and engines using compute shader stream rendering. The code is implemented for Direct3D 11 as the main problem with the current implementation is around multi-threaded usage of deferred contexts, something that's not implemented in Dx10 or earlier.

It uses device context based tracking to avoid locks / bottlenecks in the runtime and uses just 1 lock around commandlist - tracker pairs which is minor and doesn't impact any drawcalls in the highly multithreaded engine of e.g. Assassin's Creed Origins. It now also tracks calls made to DrawAuto, DrawIndexedInstancedIndirect and DrawIndexedIndirect, something the current build lacks (and adding them makes engines using these methods run into a serious bottleneck in the runtime). 

The depth buffer detection uses heuristics for screensize/depth buffer size matching so it also matches a depth buffer when resolution scaling/supersampling is used where the depth buffer is larger or smaller than the window/screen. 

Performance seems to be a little higher than in the current 3.1 release too (but not that much, a couple of fps). I also fixed the drawcall / vertices display in statistics, which always ended up being 0. 

As I'm not a seasoned C++ developer (I've spent the past 15 years writing C# ;)) I hope I've done everything right, but feel free to comment on the code so I can correct it. 

One thing that's likely going to bug you is the SDK version which is bumped up as I used VC++2017 15.5. Hopefully that's something you can live with. 

Let me know if you have remarks or change requests!


crosire:
I'll take a thorough look at this next week, but it looks good judging by a first quick glance.
Thanks a lot already.

FransBouma:
Cool! :) Things work OK in 32bit too (just tested), altho it was a minor test as I used tombraider (2013) for that. Saw I did forget to add the CXX17 suppression preprocessor flags so I've added those in an extra commit.

FransBouma:
I think I've a memory leak somewhere when the resolution changes. I changed the resolution in a playsession with AC:Origins a couple of times, then used the internal resolution scaling to up the detail and back and memory didn't get freed. It's not a memory leak that increases over time but only when resolution changes. I think it's because I don't reset things when reshade resets its internal structures. Will look into this to see if I can find the issue.

FransBouma:
Committed a couple of fixes. This should be most of it, but there's still one left. When I don't scale the resolution of the game using its resolution scaling, it's OK, scaling the window up to 5K and back to 1920 works OK, memory stays roughly the same, give or take small changes due to caching inside the engine. However once I use resolution scaling, things go off the cliff and I can't get it back to what it was, and it keeps leaking more memory every time I change the resolution. I think this might be due to the fact the code now accepts depth buffers larger than the screen, and a resolution change might leak that, but have to do more testing. (refusing to accept these buffers is promising but still shows leakage of memory).

FransBouma:
Fixed :) I cached the sizes of the depthstencils, so I didn't need to request those on the fly every time, but it turned out that if resolutionscaling/supersampling is used by the game, a lot of the stencils tend not to be cleaned up for some reason (even though I did do refcount based clean up ... ) so this could cause too much memory being allocated and the game could never recover. This is now fixed, I request the sizes on the fly, which is fast enough and works every time, no memory problem anymore, everything is cleaned up nicely. I did miss a ResizeBuffers1 call too, which I've now adjusted to do last minute cleanups as well.

So all good now. Happy 2018!

FransBouma:
on the forums there's a report (https://reshade.me/forum/releases/3744-3-1?start=100#25387) that the texturedesc retrieved sometimes contains 0 for either width or height. I can't reproduce it, no matter what I do, but if it returns 0 (docs say it never contains 0 for width nor height), then depth buffer detection fails. A workaround is of course to add a check for this, but if it's unnecessary (the cause is elsewhere) the check IMHO shouldn't be added. You ever saw texture desc retrieval return a desc with size 0?

source/runtime.cpp
```diff
@@ -196,6 +195,8 @@ namespace reshade
 				}
 			}
 		}
+
+		_drawcalls = _vertices = 0;
```
crosire: Any particular reason behind moving this line?

source/d3d11/depth_counter_tracker.hpp
```diff
@@ -0,0 +1,33 @@
...
+	void update_vertices(UINT amount) { _vertices += amount; }
+
+private:
+	struct depthstencil_counter_info
+	{
+		UINT drawcall_count = 0;
+		UINT vertices_count = 0;
+	};
+
+	std::unordered_map<ID3D11DepthStencilView*, depthstencil_counter_info> _counters_per_used_depthstencil;
```
crosire: I'd rather use a `com_ptr<ID3D11DeviceStencilView>` instead of a naked pointer to avoid manual reference counting (which is prone to errors). We can overload `std::hash<com_ptr<T>>` for this to work.

source/d3d11/d3d11_device_context.cpp
```diff
@@ -110,6 +110,9 @@ ULONG STDMETHODCALLTYPE D3D11DeviceContext::Release()
 
 		LOG(INFO) << "Destroyed 'ID3D11DeviceContext" << (_interface_version > 0 ? std::to_string(_interface_version) : "") << "' object " << this << ".";
 
+		// clear the trackers
+		clear_drawcall_stats();
```
crosire: This is not necessary if `com_ptr` are used as described above, since the `delete` will take care of cleanup.

source/d3d11/depth_counter_tracker.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <d3d11.h>
+#include <unordered_map>
+
+class depth_counter_tracker
```
crosire: Maybe rename this to `depth_stencil_tracker`? Also, it belongs in the `reshade::d3d11` namespace.

source/dxgi/dxgi_swapchain.cpp
```diff
@@ -464,6 +434,7 @@ HRESULT STDMETHODCALLTYPE DXGISwapChain::ResizeBuffers1(UINT BufferCount, UINT W
 			break;
 		case 11:
 			assert(_runtime != nullptr);
+			perform_callcounter_cleanup();
```
crosire: I don't like that this D3D11 specific code is part of the DXGI hooks. This should be generalized. I'm not sure it's even necessary, since `D3D11Device::Release` repeats this work.

source/d3d11/d3d11_device_context.hpp
```diff
@@ -189,8 +190,24 @@ struct D3D11DeviceContext : ID3D11DeviceContext3
...
+	void set_active_depthstencil(ID3D11DepthStencilView* pDepthStencilView);
+	void log_drawcall(UINT vertices);
+	void clear_drawcall_stats();
+	void log_drawcall_stats_in_tracker();
+	depth_counter_tracker& get_depth_counter_tracker();
+	
+	// fields
 	LONG _ref = 1;
+	UINT _drawcalls = 0;
+	UINT _vertices = 0;
```
crosire: `_drawcalls` and `_vertices` appear to be unused, so can be removed.

source/d3d11/d3d11_device_context.hpp
```diff
@@ -189,8 +190,24 @@ struct D3D11DeviceContext : ID3D11DeviceContext3
...
+	UINT _active_depthstencil_drawcalls = 0;
+	UINT _active_depthstencil_vertices = 0;
+	UINT _total_drawcalls = 0;
+	UINT _total_vertices = 0;
```
crosire: I don't think `_active_depthstencil_drawcalls`, `_active_depthstencil_vertices`, `_total_drawcalls` and `_total_vertices` are needed, since the same information is duplicated in `_depthstencil_counters`, so it can be just stored there directly?

source/d3d11/depth_counter_tracker.cpp
```diff
@@ -0,0 +1,135 @@
...
+			const depthstencil_counter_info counters = {};
+			to_track->AddRef();
+			_counters_per_used_depthstencil.emplace(to_track, counters);
+		}
+	}
+}
+
+void depth_counter_tracker::log_drawcalls(ID3D11DepthStencilView* depthstencil, UINT drawcalls, UINT vertices)
+{
+	if (nullptr == depthstencil || drawcalls==0)
```
crosire: Maybe make this an `assert` instead, since the same branch already exists in `D3D11DeviceContext::log_drawcall_stats_in_tracker`?

source/d3d11/d3d11_device.cpp
```diff
@@ -510,3 +513,47 @@ void STDMETHODCALLTYPE D3D11Device::ReadFromSubresource(void *pDstData, UINT Dst
...
+void D3D11Device::merge_counters_per_commandlist(ID3D11CommandList* command_list, depth_counter_tracker& counters_source)
+{
+	const std::lock_guard<std::mutex> lock(_counters_per_commandlist_mutex);
+
+	auto entry = _counters_per_commandlist.find(command_list);
+	if (entry == _counters_per_commandlist.end())
+	{
+		depth_counter_tracker counters_per_depthstencil;
+		counters_per_depthstencil.merge(counters_source);
+		command_list->AddRef();
```
crosire: I wouldn't increase the reference count here. It's not an issue if the pointer goes dangling because the game deleted the command list, since it's not used anywhere and is just there for lookup purposes. That way we don't have to deal with cleaning up the table either.

source/runtime.cpp
FransBouma: The original position reset the values they received during the frame / in onpresent in the various hooked code and the statistics tab therefore showed them as '0' in v3.1 (as 'draw_overlay()' is after the original place in `runtime::on_present()`). The line is for resetting the values for the next frame after present, so moving the statement after `draw_overlay()` fixed that :)

source/dxgi/dxgi_swapchain.cpp
FransBouma: I don't see where D3D11Device::Release is called when the buffers are resized? Debugging shows when resizing the screen, only `DXGI::ResizeBuffers` is called, not the Release in D3D11Device. So the cleanup has to take place at different places. The callcounter cleanup here is done to prevent memory leaks when the screen is resized. It's the location to make sure the call counters in at least the immediatecontext are cleared (and their references).

source/d3d11/depth_counter_tracker.hpp
FransBouma: Moved type to reshade::d3d11 namespace. 
`draw_call_tracker` is a better name indeed. Changed

source/d3d11/depth_counter_tracker.cpp
FransBouma: As this function is only called from `log_drawcall-stats_in_tracker` it's indeed redundant.

source/d3d11/d3d11_device_context.hpp
FransBouma: No, that's not how it works :) 
_active_* counters are for the currently set depthstencil on the context. See the following flow (pseudocode):
- context.Draw() -> no active depthstencil, nothing is counted
- context.setdepthstencil(X) -> the active depthstencil is set to X in context.
- context.Draw() -> has active depthstencil, so incs _active_depthstencil_drawcalls and adds vertices to _active_depthstencil_vertices. 
- context.Draw() -> see previous
- context.Draw() -> see previous
- context.setdepthstencil(Y) -> has a current active depthstencil of X. It will add X with the current values for the _active_* counters in the context's drawcall tracker. Then sets Y as the active depthstencil
etc. 
At the end it also makes sure it collects the last drawcalls for the last active depthstencil, when the tracker is obtained. 

`_total_drawcalls` and `_total_vertices` are the absolute totals, so also the drawcalls which were executed when no depthstencil was active (see `D3D11DeviceContext::log_drawcall`). So all counters are needed and have potentially different values. 

When the context C1 is a deferred context, the drawcall tracker is stored with the commandlist created from it. When the commandlist is then executed on another context C2 the drawcall tracker stats of C1 (which are associated with the commandlist) are merged with the one in C2. This then all bubbles down to all stats being merged eventually in the tracker in the immediatecontext. That is, the ones which were actually executed (e.g. commandlist which was tossed away isn't merged. Commandlist which was executed 10 times is merged 10 times)

source/d3d11/d3d11_device_context.hpp
FransBouma: good catch, removed.

source/d3d11/d3d11_device.cpp
FransBouma: I'll remove the addref. I added it in my quest to get rid of the memory leaks I discovered. I indeed found it unnecessary to use refcounting here (and also with the depthstencil as these are also just for lookup) but didn't know what unordered_map would do with the pointer under the hood, so to be safe I added the ref counter. 
Cleanup is still needed, as each frame commandlists are created and the table otherwise will increase each frame and lists which will never be used will stay in it.

source/d3d11/depth_counter_tracker.hpp
FransBouma: As the pointer is here used just for lookup, it might be unnecessary to use com_ptr as well and simply don't do ref counting here? I think a refcount check then is needed in `draw_call_tracker::get_best_depth_stencil` to check if the buffer is still alive and we're not using a dangling pointer but that would be all? (also I don't know how to overload the hashfunction for com_ptr, but that's one google query away of course ;))

FransBouma: Testing shows refcounters here are absolutely needed, memory leak is otherwise unavoidable when internal rendering resolution in e.g. AC:Origins is increased/decreased. Will look into hashfunction overrides for com_ptr

FransBouma: Hash struct implemented. Hopefully I've done it properly (As I relied on a stackoverflow example for this) Tests show it makes sure everything is cleaned up properly, no ref counting needed :)

FransBouma:
Changes implemented and pushed. :)

source/runtime.cpp
crosire: Oh. Good catch!

source/dxgi/dxgi_swapchain.cpp
crosire: Alright. I can think of a solution later. It's fine for the initial merge.

source/d3d11/d3d11_device_context.cpp
crosire: Have you checked if removing this causes memory leaks. Because if I read everything right it shouldn't now and this line is safe to remove.

crosire: Have you checked if removing this causes memory leaks. Because if I read everything right it shouldn't now and this line is safe to remove.

crosire:
Have you checked if removing the `clear_drawcall_stats();` line as described above causes memory leaks? Because if I read everything right it shouldn't now and the line is safe to remove.

FransBouma:
Removing that call indeed works too, no mem leaks whatsoever. Have removed it in the latest push :)

FransBouma:
Thanks!

PR #40, opened 2018-03-01
New buffer detection techniques

Thalixte:
Add some new buffer detection techniques:
- One at Output Merger State
- One at ClearDepthStencilView stage

Add a new tab to tweak the depth buffer detection


ReShade.vcxproj
```diff
@@ -40,7 +40,7 @@
     <TargetName>ReShade32</TargetName>
   </PropertyGroup>
   <PropertyGroup Label="Configuration" Condition="'$(Platform)'=='x64'">
-    <TargetName>ReShade64</TargetName>
+    <TargetName>Reshade64</TargetName>
```
crosire: Whoops.

source/d3d11/d3d11.cpp
```diff
@@ -22,7 +22,7 @@ HOOK_EXPORT HRESULT WINAPI D3D11CreateDeviceAndSwapChain(IDXGIAdapter *pAdapter,
 	LOG(INFO) << "Redirecting '" << "D3D11CreateDeviceAndSwapChain" << "(" << pAdapter << ", " << DriverType << ", " << Software << ", " << std::hex << Flags << std::dec << ", " << pFeatureLevels << ", " << FeatureLevels << ", " << SDKVersion << ", " << pSwapChainDesc << ", " << ppSwapChain << ", " << ppDevice << ", " << pFeatureLevel << ", " << ppImmediateContext << ")' ...";
 
 #ifdef _DEBUG
-	Flags |= D3D11_CREATE_DEVICE_DEBUG;
+	// Flags |= D3D11_CREATE_DEVICE_DEBUG;
```
crosire: This shouldn't be committed.

source/d3d11/d3d11_device.cpp
```diff
@@ -64,7 +64,7 @@ HRESULT STDMETHODCALLTYPE D3D11Device::QueryInterface(REFIID riid, void **ppvObj
 		riid == __uuidof(ID3D11Device2) ||
 		riid == __uuidof(ID3D11Device3))
 	{
-		#pragma region Update to ID3D11Device1 interface
+#pragma region Update to ID3D11Device1 interface
```
crosire: This entire file can be reverted, since it's just (probably unintended) whitespace changes.

source/d3d11/d3d11_device.hpp
```diff
@@ -26,12 +26,12 @@ struct D3D11Device : ID3D11Device3
 	D3D11Device(const D3D11Device &) = delete;
 	D3D11Device &operator=(const D3D11Device &) = delete;
 
-	#pragma region IUnknown
+#pragma region IUnknown
```
crosire: Same thing for this file.

source/d3d11/d3d11_device_context.cpp
```diff
@@ -1,12 +1,16 @@
 /**
- * Copyright (C) 2014 Patrick Mours. All rights reserved.
- * License: https://github.com/crosire/reshade#license
- */
+* Copyright (C) 2014 Patrick Mours. All rights reserved.
+* License: https://github.com/crosire/reshade#license
+*/
```
crosire: Please make sure whitespace matches that of the original code. There are quite a few of those issues, so I'm not going to mark them all. See the pull request diff.

source/d3d11/d3d11_device_context.hpp
```diff
@@ -183,22 +183,32 @@ struct D3D11DeviceContext : ID3D11DeviceContext3
...
+	com_ptr<ID3D11Texture2D> copy_depth_texture(D3D11_TEXTURE2D_DESC texture_desc, ID3D11Texture2D *depthtexture);
+	bool check_depthstencil(ID3D11DepthStencilView* pDepthStencilView, com_ptr<ID3D11Texture2D> texture, D3D11_TEXTURE2D_DESC texture_desc);
 	void set_active_depthstencil(ID3D11DepthStencilView* pDepthStencilView);
+	void set_active_OM_depthstencil(ID3D11DepthStencilView* pDepthStencilView);
+	void set_active_cleared_depthstencil(ID3D11DepthStencilView* pDepthStencilView);
 
 	void clear_drawcall_stats();
 
 	LONG _ref = 1;
+	unsigned int _vertices_treshold = 10000;
```
crosire: I don't see a reason behind this field. It looks to be a constant, so may as well put it in as a constant where it's used in the cpp file.

source/d3d11/d3d11_device_context.hpp
```diff
@@ -183,22 +183,32 @@ struct D3D11DeviceContext : ID3D11DeviceContext3
...
 	ID3D11DeviceContext *_orig;
 	unsigned int _interface_version;
 	D3D11Device *const _device;
 	com_ptr<ID3D11DepthStencilView> _active_depthstencil;
+	com_ptr<ID3D11Texture2D> _depth_texture;
 	reshade::d3d11::draw_call_tracker _draw_call_tracker;
+	static unsigned int _clear_DSV_iter;
```
crosire: `static` variables are really bad in ReShade. Especially here! Many modern games spawn multiple device contexts on differents threads, so that the code using these variables would now potentially access them simultanously and cause all kinds of bad behavior (race conditions, ...).

source/d3d11/d3d11_runtime.cpp
```diff
@@ -513,6 +515,7 @@ namespace reshade::d3d11
 			return;
 		}
 
+		// LOG(INFO) << "Call to detect_depth_source(tracker)";
```
crosire: This line should go =).

source/d3d11/draw_call_tracker.cpp
```diff
@@ -12,28 +13,43 @@ namespace reshade::d3d11
...
+	{
+		assert(depth_texture != nullptr);
+
+		_active_depth_texture = depth_texture;
 	}
 
 	void draw_call_tracker::track_depthstencil(ID3D11DepthStencilView* depthstencil)
 	{
-		assert(depthstencil != nullptr);
+		// assert(depthstencil != nullptr);
```
crosire: Why?

source/d3d11/draw_call_tracker.cpp
```diff
@@ -59,21 +75,35 @@ namespace reshade::d3d11
 		}
 	}
 
-	ID3D11DepthStencilView* draw_call_tracker::get_best_depth_stencil(UINT width, UINT height)
+	ID3D11DepthStencilView* draw_call_tracker::get_best_depth_stencil(std::string host_process_name, std::unordered_map<std::string, runtime::game> game_list, UINT drawcalls, const com_ptr<ID3D11Device> device, const com_ptr<ID3D11DeviceContext> devicecontext, UINT width, UINT height)
 	{
 		depthstencil_counter_info best_info = { 0 };
 		ID3D11DepthStencilView *best_match = nullptr;
 		float aspect_ratio = ((float)width) / ((float)height);
+		static runtime::game current_game =
```
crosire: This isn't used, so might as well remove it.

source/d3d11/draw_call_tracker.cpp
```diff
@@ -59,21 +75,35 @@ namespace reshade::d3d11
...
+			game_list.count(host_process_name) ?
+			game_list[host_process_name] :
+			runtime::game::UNKNOWN;
+		reshade::runtime::OM_iter = 0;
```
crosire: Uh, oh, static variable ahead.

source/runtime.cpp
```diff
@@ -110,6 +118,8 @@ namespace reshade
 
 		_is_initialized = true;
 		_last_reload_time = std::chrono::high_resolution_clock::now();
+		_init_game_list(_game_list);
+		_host_process_name = _get_host_app();
```
crosire: ReShade should be kept generic, so these changes shouldn't go into the official repository.

source/runtime.cpp
```diff
@@ -439,6 +449,29 @@ namespace reshade
...
+	{
+
+		wchar_t host_app[_MAX_PATH * 2] = {};
+		utility::getHostApp(host_app);
+		std::wstring ws(host_app);
+		std::string str(ws.begin(), ws.end());
+		return str;
+	}
+
+	void const runtime::_init_game_list(std::unordered_map<std::string, game> &game_list)
```
crosire: See above.

source/runtime.cpp
```diff
@@ -1265,9 +1306,9 @@ namespace reshade
 		{
 			ImGui::PushStyleVar(ImGuiStyleVar_ItemSpacing, ImGui::GetStyle().ItemSpacing * 2);
 
-			const char *const menu_items[] = { "Home", "Settings", "Statistics", "About" };
+			const char *const menu_items[] = { "Home", "Settings", "Depth Buffer detection settings", "Statistics", "About" };
```
crosire: This really shouldn't be another menu entry, but rather a section under "Settings".

source/runtime.hpp
```diff
@@ -31,14 +31,70 @@ namespace reshade
 	class runtime abstract
 	{
 	public:
+		enum class game
```
crosire: See above.

source/runtime.hpp
```diff
@@ -31,14 +31,70 @@ namespace reshade
...
+		};
+
 		/// <summary>
 		/// File path to the current module.
 		/// </summary>
 		static filesystem::path s_reshade_dll_path;
 		/// <summary>
+		/// screen width
+		/// </summary>
+		static unsigned int screen_width;
```
crosire: Apart from the fact that `static` variables are a bad idea here again, because applications can technically have multiple DX devices / OGL contexts at the same time, which would instantiate the ReShade runtime multiple times, so that this variable would be read/written to by different runtime instances with different size values, there is also no need for these. Instead use `runtime::frame_width()` and `runtime::frame_height()`. You can get access to all the runtime instances in D3D11 via the `_runtimes` field in a `D3D11Device`. Make sure they are all represented, since again, multiple could technically exist (even though this is rare).

source/runtime.hpp
```diff
@@ -31,14 +31,70 @@ namespace reshade
...
+		/// </summary>
+		static unsigned int depth_buffer_retrieval_mode;
+		/// <summary>
+		/// Depth buffer clearing number
+		/// </summary>
+		static unsigned int depth_buffer_clearing_number;
+		/// <summary>
+		/// Depth buffer texture format
+		/// </summary>
+		static unsigned int depth_buffer_texture_format;
```
crosire: Why are all these `static`. Please make them normal fields in `runtime`. They may even be much better suited as fields in `d3d11_runtime` since they are currently specific to that implementation!! This would make it hard to get access to them in ImGUI though.

source/utility.cpp
```diff
@@ -0,0 +1,189 @@
...
+			fclose(file);
+		}
+
+		if (!success)
+		{
+			LOG(ERROR) << "Failed to write screenshot to " << path << "!";
+		}
+	}
+
+}
```
crosire: This entire file is no longer necessary, since with the changes requested above, none of the code in here is used anymore.

source/utility.hpp
```diff
@@ -0,0 +1,13 @@
...
+#include "filesystem.hpp"
+#include "com_ptr.hpp"
+
+namespace reshade::utility
+{
+	bool findStringIC(std::vector<std::string> vec, const std::string toSearch);
+	void getHostApp(wchar_t* wszProcOut);
+	void capture_depthstencil(const com_ptr<ID3D11Device> &device, const com_ptr<ID3D11DeviceContext> &devicecontext, const com_ptr<ID3D11Texture2D> &texture, D3D11_TEXTURE2D_DESC texture_desc, uint8_t *buffer, bool greyscale);
+	void save_depthstencil(filesystem::path parent_path, filesystem::path filename_without_extension, D3D11_TEXTURE2D_DESC texture_desc, uint8_t *buffer);
+}
```
crosire: See above.

source/d3d11/d3d11_device.cpp
Thalixte: Made the change. I commented this line because, in debug mode, it makes Reshade crash... Do you know why it behave like that ?

ReShade.vcxproj
Thalixte: Made the change. Have renamed the target name to dxgi to ease the debug process. When reverted to the original name, did  not noticed the case.

source/d3d11/d3d11_device_context.hpp
Thalixte: Yes, i know. Will add a mutex here (did it before), but in my testings, did not encountered any race conditions conflict. Leaved it like that in order nort to impact perfs.
Are you agree i add a mutex for it ?

source/d3d11/draw_call_tracker.cpp
Thalixte: In some testings, made Reshade crash. Now, should be ok. Will remove the comment.

source/d3d11/draw_call_tracker.cpp
Thalixte: OK. I made it because on my first attempts, i added hacks per game in the post processing stage, in the draw_call_tracker get_best_depthstencil method.

source/runtime.cpp
Thalixte: I found it clearer like that, but OK, will move it to the settings tab ;-)

source/utility.cpp
Thalixte: OK Removed it...

source/runtime.hpp
Thalixte: Yes, you are totally right. Setting these to static is useless...

source/runtime.cpp
Thalixte: OK, will keep it generic.

source/d3d11/d3d11_device.cpp
Thalixte: yes, it is. Whare ate your format settings for the whitespaces in your IDE ?

source/runtime.hpp
Thalixte: Ok, but if there are multiple runtimes, which one can i use to retrieve the screen width and screen height in a specific d3d11_context ?

source/d3d11/d3d11_device.cpp
crosire: I'm assuming you mean the change above. You need the DX debug layers installed for that flag to work. That should usually happen when you install the Windows SDK via Visual Studio though.

source/runtime.cpp
Thalixte: I found it clearer like that, but OK, will move it to the Settings tab ;-)

source/d3d11/d3d11_device_context.hpp
crosire: Why not use atomic variables? No need for a mutex. See for example http://en.cppreference.com/w/cpp/atomic/atomic.

source/runtime.hpp
crosire: I suppose you can just use the first in the array, since at that point you don't know which runtime the game will present on later on.

Thalixte: OK, made the change as suggested

source/d3d11/d3d11_device_context.hpp
```diff
@@ -191,14 +194,27 @@ struct D3D11DeviceContext : ID3D11DeviceContext3
...
 	LONG _ref = 1;
 	ID3D11DeviceContext *_orig;
 	unsigned int _interface_version;
 	D3D11Device *const _device;
 	com_ptr<ID3D11DepthStencilView> _active_depthstencil;
+	com_ptr<ID3D11Texture2D> _depth_texture;
 	reshade::d3d11::draw_call_tracker _draw_call_tracker;
+	FLOAT _screen_width = 0;
+	FLOAT _screen_height = 0;
+	std::atomic<unsigned int> _clear_DSV_iter = 1;
```
crosire: If this still works with these variables being local to a device context (not static), then `std::atomic` isn't actually needed anymore, because all operations local to one device context have to run on the same thread, so no race conditions can occur. Atomic operations or synchronization is only necessary if multiple device contexts could access the same data.

Thalixte: Yes, but i want those variables to be device context independant. It seems that some games use multiple device contexts in parallel, but maybe i'am wrong...

Thalixte: It is in the case of deferred context (as ClearDepthStencilView method can be embedded in a command list, so i don't know the order of execution)...

source/runtime.hpp
Thalixte: For the two other params, i need them in different part of Reshade, so that's why i prefered to set them static.

Thalixte:
Hi, Crosire... Is there anything remaining ?

source/d3d11/d3d11_device_context.hpp
crosire: That is all good and well, I just noted, that the way they are used right now means atomics are not needed. Had they been static, this would be an entirely different story, but in you code right now they are local to a device context and therefore to a thread, so no atomics necessary.

Thalixte:
Thks, Crosire. Merged your pull request

source/runtime.hpp
```diff
@@ -31,6 +31,29 @@ namespace reshade
...
+		};
+
+		enum depth_buffer_texture_type
+		{
+			BOTH = 0,
+			DEPTH_BUFFER = 1,
+			STENCIL_BUFFER = 2
+		};
+
+		enum class allowed_technique
```
crosire: Just noticed that this is not used anywhere and should probably go.

source/d3d11/draw_call_tracker.cpp
```diff
@@ -12,23 +13,38 @@ namespace reshade::d3d11
...
+				{
+					destination_entry->second.vertices += source_entry.second.vertices;
+					destination_entry->second.drawcalls += source_entry.second.drawcalls;
+				}
 			}
 		}
 	}
 
-	void draw_call_tracker::reset()
+	void draw_call_tracker::reset(bool ball)
```
crosire: What does `ball` mean in this context?

source/d3d11/draw_call_tracker.cpp
```diff
@@ -12,23 +13,38 @@ namespace reshade::d3d11
...
 	{
 		_counters.vertices = 0;
 		_counters.drawcalls = 0;
-		_counters_per_used_depthstencil.clear();
+
+		if (ball == true)
```
crosire: `if (ball)` does the same.

source/d3d11/draw_call_tracker.cpp
```diff
@@ -1,4 +1,5 @@
 #include "draw_call_tracker.hpp"
+#include "log.hpp"
```
crosire: Doesn't look like logging is used anywhere in this file, so this line is unnecessary.

source/d3d11/d3d11_device_context.hpp
crosire: Thoughts?

source/d3d11/d3d11_device_context.cpp
```diff
@@ -16,23 +16,254 @@ void D3D11DeviceContext::log_drawcall(UINT vertices)
...
+	_clear_DSV_iter = 1;
+	_OM_iter = 0;
+	_best_vertices = 0;
+	_best_drawcalls = 0;
+	_draw_call_tracker.reset(true);
 	_active_depthstencil.reset();
+	_depth_texture.reset();
+}
+
+void D3D11DeviceContext::screen_dimensions()
```
crosire: I don't see the reason behind this function and the `_screen_width` + `_screen_height` class fields. They are used in a single function only, so may as well store them on the stack instead.

source/d3d11/d3d11_device_context.cpp
```diff
@@ -16,23 +16,254 @@ void D3D11DeviceContext::log_drawcall(UINT vertices)
...
+			return;
+		}
+
+		reshade::d3d11::draw_call_tracker::depthstencil_counter_info counters = _draw_call_tracker.get_counters();
+
+		if (reshade::runtime::depth_buffer_clearing_number == 0)
+		{
+			set_active_depthstencil(pDepthStencilView);
+			// copy the depth stencil texture
+			_depth_texture = copy_depth_texture(texture_desc, texture.get());
```
crosire: This looks like a bad idea. Creating a new texture every frame is a sure way to kill performance. Why not use the `_depthstencil_texture` field in `d3d11_runtime` and keep a persistent texture around that you can copy into? I feel like large parts of this algorithm should be moved to the `d3d11_runtime` class anyway ... I'm still not a fan of the `static` variables used here, which doing so would solve as well.

source/d3d11/d3d11_device_context.cpp
```diff
@@ -16,23 +16,254 @@ void D3D11DeviceContext::log_drawcall(UINT vertices)
...
+		LOG(ERROR) << "Failed to create depth stencil view copy! HRESULT is '" << std::hex << hr << std::dec << "'.";
+		return nullptr;
+	}
+
+	return depthStencilView_copy;
+}
+
+bool D3D11DeviceContext::check_depthstencil(com_ptr<ID3D11Texture2D> texture, D3D11_TEXTURE2D_DESC texture_desc)
+{	
+	screen_dimensions();
```
crosire: See above. Should better be replaced with:
```c++
const auto runtime = _device->_runtimes.front();
const auto screen_width = static_cast<FLOAT>(runtime->frame_width());
const auto screen_height = static_cast<FLOAT>(runtime->frame_height());
```

source/d3d11/d3d11_device_context.cpp
```diff
@@ -16,23 +16,254 @@ void D3D11DeviceContext::log_drawcall(UINT vertices)
...
+		if (!FAILED(resource->QueryInterface(&texture)))
+		{
+			texture->GetDesc(&texture_desc);
+		}
+		else
+		{
+			return;
+		}
+
+		if (!check_depthstencil(texture, texture_desc))
```
crosire: Both calls to `check_depthstencil` have the same boilerplate code right in front of them (`GetResource` and `GetDesc`). Makes sense to move that code inside the function to avoid code duplication.

source/runtime.hpp
```diff
@@ -31,6 +31,29 @@ namespace reshade
 	class runtime abstract
 	{
 	public:
+		enum depth_buffer_retrieval_mode
+		{
+			POST_PROCESS = 0,
```
crosire: Also, naming convention: These should be snake_case.

crosire:
Anyway, I think I'll merge this when the last formatting issues a sorted

crosire:
Wrong button. I'll merge this once the last few formatting issues mentioned above sorted out. It can be further refined when that's done and its in the repo.

source/d3d11/d3d11_device_context.cpp
Thalixte: changed

source/d3d11/draw_call_tracker.cpp
Thalixte: fixed

source/d3d11/draw_call_tracker.cpp
Thalixte: changed

source/d3d11/draw_call_tracker.cpp
Thalixte: changed

source/runtime.hpp
Thalixte: fixed

source/runtime.hpp
Thalixte: fixed

source/d3d11/d3d11_device_context.cpp
Thalixte: This is because i need them (texture and texture_desc) in the rest of the methods set_active_OM_depthstencil and set_active_cleared_depthstencil

source/d3d11/d3d11_device_context.hpp
Thalixte: Thks Crosire for your suggestions. I have to test it in order to see if it will not break the depth buffer detection, especially when the user force the number of the clear pass he wants reshade to retrieve the depth data

source/d3d11/d3d11_device_context.cpp
Thalixte: That is what i thought at first, but the copyResource method seems to be an asynchronous one, so it does not kill the performance a lot, especially is you call it only once per frame. This method also copy from GPU VRAM to GPU VRAM.
 In my tests, i did not encounter big losses in perfs.
Keep a persistent texture implies that you make a copy of the texture before it is cleared by the game engine, so you still need a copyResource method. Another way to do it, would be to unlink the depth texture to the depthstencil before the engine clears this texture then, in the final rendering stage, relink this texture to the depthstencil, but i do not see a good way to do so....

source/d3d11/d3d11_device_context.cpp
Thalixte: changed

source/d3d11/d3d11_device_context.cpp
crosire: I'm not concerned about the `CopyResource`. That one is fast.I'm concerned about the `CreateTexture`. That one should not be called every frame. Allocation is slow.

Thalixte: Ok, i understand what you mean. Thks for having comitted my pull request ;-)

crosire:
General heads-up: I rebased the repository to clean up the history a bit, so it's a good idea to refork or at least update and checkout the new master before continuing, so any future pull requests point to the correct base commit.

PR #41, opened 2018-04-05
D3d9

Thalixte:
New buffer detection techniques for d3d9. i revert back input.hpp and input.cpp, because in some d3d9 games (Dark messiah, for instance), the gui doesn't appear when pressing the shortcut (alt+F2 in my case)

### DIFF
--- a/source/runtime.hpp
+++ b/source/runtime.hpp
@@ -188,8 +188,12 @@ namespace reshade
 		std::vector<uniform> _uniforms;
 		std::vector<technique> _techniques;
 		int _depth_buffer_texture_format = 0; // No depth buffer texture format filter by default
+		int _depth_buffer_clearing_number = 0;
+		int _depth_buffer_clearing_flag_number = 6;
 		bool _depth_buffer_debug = false;
 		bool _depth_buffer_before_clear = false;
+		bool _depth_buffer_settings_changed = false;
+		bool _restrict_depth_buffer_dimensions = true;
 
 	private:
 		static bool check_for_update(unsigned long latest_version[3]);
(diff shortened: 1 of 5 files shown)


source/runtime.hpp
```diff
@@ -155,6 +163,9 @@ namespace reshade
 		/// Callback function called to apply the post-processing effects to the screen.
 		/// </summary>
 		void on_present_effect();
+		bool is_d3d9 = false;
+		bool is_d3d10 = false;
+		bool is_d3d11 = false;
```
crosire: There is `_renderer_id` for that. These are not needed. Use the following in-place where needed instead:
```
const bool is_d3d9 = (_renderer_id & 0x9000) != 0;
const bool is_d3d10 = (_renderer_id & 0xa000) != 0;
const bool is_d3d11 = (_renderer_id & 0xb000) != 0;
```

crosire:
I don't plan on reverting the input changes. The current solution is better than what was done previously. I'd much rather figure out why it's not working in some games, so please post a list of where it isn't working.

source/runtime.hpp
Thalixte: I know, but in my rig, it did not work... Will retest

Thalixte:
I personnaly check this behaviour in Dark messiah. According to some users, games that use the source engine 2006 have this behaviour...

source/runtime.hpp
Thalixte: Have retested it, and still does not work on my rig. On the other hand, the fact of setting those bool vars in the init stage of the runtime work perfectly.

crosire: Sorry, the code I wrote is wrong. It's actually this:
```c++
const bool is_d3d9 = _renderer_id < 0xa000;
const bool is_d3d10 = _renderer_id >= 0xa000 && _renderer_id < 0xb000;
const bool is_d3d11 = _renderer_id >= 0xb000 && _renderer_id < 0xc000;
```

crosire:
The input problem is fixed (https://github.com/crosire/reshade/commit/9cc66e6f2d560eef693cffe8476c2ef97c9207bf).

Thalixte:
@Crosire: This been a long time since i didn't work on this branch, and i plan to work on  it later. At this stage, it is not fully functionnal, do i prefer to close this pull request.

PR #42, opened 2018-04-09
Made mousedelta work using raw input

luluco250:
I created a new short[2] called _mouse_delta that replaces the old _last_mouse_position. This new one gets it's values from raw input, using raw_data.data.mouse.lLastX and raw_data.data.mouse.lLastY, working very similarly to _mouse_wheel_delta.

This makes it so mousedelta works on any game regardless of cursor position, as some games completely center the cursor and others confine it within the screen bounds, which stops the delta. Now it works independently and (shameless plug) can be used for mouse motion blur shaders even.

I deleted _last_mouse_position as I see no other use for it, unless you prefer to have both a position delta and raw delta.

### DIFF
--- a/source/input.cpp
+++ b/source/input.cpp
@@ -159,9 +159,6 @@ namespace reshade
 
 						if (raw_data.data.mouse.usButtonFlags & RI_MOUSE_WHEEL)
 							input._mouse_wheel_delta += static_cast<short>(raw_data.data.mouse.usButtonData) / WHEEL_DELTA;
-
-						input._mouse_raw_delta[0] = static_cast<short>(raw_data.data.mouse.lLastX);
-						input._mouse_raw_delta[1] = static_cast<short>(raw_data.data.mouse.lLastY);
 						break;
 					case RIM_TYPEKEYBOARD:
 						is_keyboard_message = true;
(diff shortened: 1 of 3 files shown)


luluco250:
Oh I see, I forgot that detail. Maybe make a different uniform called "`mouserawdelta`" could be a simple fix in the meantime?

I'll check out `WM_MOUSEMOVE` and see if I can calculate it there. I think it's best if `mousedelta` doesn't depend on anything in-game.

luluco250:
It doesn't seem possible to get relative mouse movement using `WM_MOUSEMOVE`, games that use it center the cursor to calculate the delta properly.

I think the best tradeoff in this case would be to have a `mouseraw` uniform that provides the raw input values, separate from `mousedelta`. The wheel delta could also be included in `mouseraw` as the Z component, as I can imagine it could be useful in some shaders like zooming.

Should I close this or do you have any further thoughts?

crosire:
Why not simply check if `_mouse_raw_delta` is not zero in `mouse_movement_delta_x()` and friend and return it, otherwise return the difference between the last and current position? No need for an additional uniform variable source.

luluco250:
Well I tested the raw input delta on Mass Effect and it seems to use neither WM_MOUSEMOVE nor raw input, but instead DirectInput, which uses raw input underneath a separate thread.

The only way I was able to correctly get the relative mouse movement without any blocks in that case was through SetCursorPos(), which is obviously a bad idea, unless maybe as a togglable option/key shortcut. Using this method made mousedelta work perfectly and the game camera still moved, but the mouse obviously didn't move in menus.

What if ReShade were to register the mouse as a raw input device, either on the game's hwnd or on it's own? Or maybe it could be possible to hook into DirectInput? Yes, I'm aware that would be some massive additional complexity for a niche problem, but I'm just throwing ideas out there.

luluco250:
I created a separate branch to work on this more, I'll continue this discussion here: https://reshade.me/forum/general-discussion/4154-make-mousedelta-ignore-screen-borders

PR #43, opened 2018-04-17
unknown -  replaced reshade filesystem::path with std::experimental::…

sfinktah:
…filesystem::path

basically 

    namespace reshade::filesystem
    {
	enum class special_folder
	{
		app_data,
		system,
		windows,
	};

	using path = std::experimental::filesystem::path;

with changes where required to make a few other calls work.

### DIFF
--- a/source/d3d10/d3d10_effect_compiler.cpp
+++ b/source/d3d10/d3d10_effect_compiler.cpp
@@ -202,7 +202,7 @@ namespace reshade::d3d10
 		if (_ast.techniques.size() == 0)
 			return;
 		_dump_filename = _ast.techniques[0]->location.source;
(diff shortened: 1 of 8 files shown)


R00dNet:
@sfinktah is it possible to get latest launch off of you please? i missed closure, a lot of nonsense online... :( plat looking for help.

crosire:
I'd rather not use features from an experimental TS until they are standardized right now, especially since (from what I could gather) it's does not look like there is a clearly defined way to convert an UTF-8 string to a `std::experimental::filesystem::path`. There is `.u8string()` for the other way, but not the construction. Also, since the filesystem TS does not include all the features needed for ReShade anyway, the code amount savings aren't worth it I think.

PR #44, opened 2018-04-17
Using count is bad for chrono

sfinktah:


source/runtime.cpp
```diff
@@ -1108,7 +1108,7 @@ namespace reshade
 
 	void runtime::draw_overlay()
 	{
-		const bool show_splash = std::chrono::duration_cast<std::chrono::seconds>(_last_present_time - _last_reload_time).count() < 5;
+		const bool show_splash = (_last_present_time - _last_reload_time) < std::chrono::seconds(5);
```
crosire: Good point. Could even do this:
```c++
using namespace std::chrono_literals;
const bool show_splash = (_last_present_time - _last_reload_time) < 5s;
```

PR #46, opened 2018-04-17
string_codecvt c++17 compatibility

sfinktah:
cons:

* rather lazily used the excellent nowide library.
* didn't include it as a submodule
* added path to project rather than solution

pros:
* works


crosire:
Fixed in 01bbafc8f56ef00cfa3ce221bc622a64b240b8a0.

PR #47, opened 2018-05-17
moved log into the menu

el3ment:
display log messages in menu,
added debug log level
added dynamic filter to log message menu display
added word wrap option to log message menu display

### DIFF
--- a/source/log.hpp
+++ b/source/log.hpp
@@ -31,7 +31,6 @@ namespace reshade::log
 
 	extern std::ofstream stream;
 	extern std::ostringstream linestream;
-	extern int log_level;
 	extern std::mutex _mutex;
 	extern std::vector<std::string> lines;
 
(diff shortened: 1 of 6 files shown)


source/log.hpp
```diff
@@ -35,30 +44,32 @@ namespace reshade::log
...
 		{
-			static_assert(sizeof(std::wstring::value_type) == sizeof(uint16_t), "expected 'std::wstring' to use UTF-16 encoding");
-			std::string utf8_message;
-			utf8_message.reserve(message.size());
-			utf8::unchecked::utf16to8(message.begin(), message.end(), std::back_inserter(utf8_message));
-			return operator<<(utf8_message);
-		}
+			std::string utf8message;
+			utf8::unchecked::utf16to8(message.begin(), message.end(), std::back_inserter(utf8message));
+			return operator<<(utf8message);
```
crosire: What's the reasoning behind removing the `reserve` call? `std::back_inserter` uses `push_back` which greatly benefits from a pre-allocated storage space.

el3ment: oops - this was just a mistake --all three of these lines (44-46) shouldn't have been deleted. I was using an older log.hpp file and didn't realize you had already fixed these.

source/log.hpp
```diff
@@ -35,30 +44,32 @@ namespace reshade::log
...
 		}
 		inline message &operator<<(const wchar_t *message)
 		{
-			static_assert(sizeof(wchar_t) == sizeof(uint16_t), "expected 'wchar_t' to use UTF-16 encoding");
-			std::string utf8_message;
-			utf8::unchecked::utf16to8(message, message + wcslen(message), std::back_inserter(utf8_message));
-			return operator<<(utf8_message);
+			std::wstring wmessage = std::wstring(message);
+			std::string utf8message;
+			utf8::unchecked::utf16to8(wmessage.begin(), wmessage.end(), std::back_inserter(utf8message));
```
el3ment: all of these shouldn't have been changed - sorry about that

el3ment:
fixed the log.hpp changes that shouldn't have been made

PR #48, opened 2018-05-17
Dynamic menu

el3ment:
Moves the tabs in the menu into a dynamic list. Subclasses, or other objects with a reference to runtime can call a function to add a label (like "DX11") and function (`dx11_runtime::draw_dx11_menu()`) to draw the menu. This will allows library-specific configuration displays.

In addition, the load_configuration and save_configuration functions can be subscribed to. This allows subclasses (and plugins) to have specific configuration. In the case of DX11, it allows the debug buffer menu to be drawn to a special tab, and the two dx11 specific settings to be housed in the dx11_runtime class.

A possible breaking change (that isn't that important) is I renamed the BUFFER_DETECTION section in the config file to DX11_BUFFER_DETECTION to be more accurate.


crosire:
This is so much better than the workaround hack that was in before, awesome!

PR #49, opened 2018-05-18
draw_call_tracker now tracks intermediate buffers (normals, motion vectors, etc) and constant buffers

el3ment:
The primary goal of this pull request is to set the groundwork for capturing additional buffers useful for post processing shaders. In the future, Reshade could pass surface normals and/or camera transforms to the post processing shaders for things like ambient occlusion and temporal aliasing. A little more work needs to be done to this effect (heuristics and config options) - but this represents the back-end component

- I wasn't able to test this on deferred context pipelines... but I don't think there should be any issues. It would be ideal if we could test that before merging into master
- pointers to intermediate buffers that are drawn along side depth buffers are stored in a "snapshot"
- pointers to constant buffers that are mapped along side depth buffers are stored
- the  `draw_call_tracker::find_best_snapshot()` function now (changed from `draw_call_tracker::find_best_depthstencil()` returns an object which contains rendertargets as well as the depthstencil
- these additional buffers are not used anywhere yet
- rather than keep track of rendertargets and depthstencils as they are drawn (with an `_active_depthstencil` variable) the rendertargets are queried during an `on_draw`. I don't think this causes any performance issues and it GREATLY enhances the ability of the draw_call_tracker to collect data
- `draw_call_tracker::verticies` and `draw_call_tracker::drawcalls` are now `total_verticies` and `total_drawcalls`
- the additional data about render targets and constant buffers should be displayed in the dx11 configuration/menu .... after we merge the new menu changes in and rebase this pull request, it will be easy to add those

### DIFF
--- a/source/d3d11/draw_call_tracker.cpp
+++ b/source/d3d11/draw_call_tracker.cpp
@@ -1,6 +1,7 @@
 #include "draw_call_tracker.hpp"
 #include "runtime.hpp"
 #include <math.h>
+#include "log.hpp"
 
 namespace reshade::d3d11
 {
@@ -67,7 +68,7 @@ namespace reshade::d3d11
 					}
 					else {
 						// This shouldn't happen - it means somehow someone has called "draw" with a rendertarget without calling track_rendertargets first
-						assert(false);
+						LOG(ERROR) << "Draw has been called on an untracked rendertarget.";
 					}
 				}
 			}
@@ -133,21 +134,18 @@ namespace reshade::d3d11
 		}
 	}
 
-	std::pair<ID3D11DepthStencilView *, ID3D11Texture2D *> draw_call_tracker::find_best_depth_stencil(UINT width, UINT height, DXGI_FORMAT format)
+	draw_call_tracker::intermediate_snapshot_info draw_call_tracker::find_best_snapshot(UINT width, UINT height, DXGI_FORMAT format)
 	{
 		const float aspect_ratio = float(width) / float(height);
 
-		intermediate_snapshot_info  best_info = { };
-		std::pair<ID3D11DepthStencilView *, ID3D11Texture2D *> best_match = { };
+		intermediate_snapshot_info best_snapshot;
 
-		for (const auto &[depthstencil, snapshot] : _counters_per_used_depthstencil)
+		for (auto &[depthstencil, snapshot] : _counters_per_used_depthstencil)
 		{
-
 			if (snapshot.counter.drawcalls == 0 || snapshot.counter.vertices == 0)
 			{
 				continue;
 			}
-
 
 			if (snapshot.texture == nullptr)
 			{
@@ -160,11 +158,12 @@ namespace reshade::d3d11
 					continue;
 				}
 
-				best_info.texture = texture;
+				snapshot.texture = texture;
 			}
 
 			D3D11_TEXTURE2D_DESC desc;
-			best_info.texture->GetDesc(&desc);
+			snapshot.texture->GetDesc(&desc);
+
 			assert((desc.BindFlags & D3D11_BIND_DEPTH_STENCIL) != 0);
 
 			// Check aspect ratio
@@ -185,16 +184,12 @@ namespace reshade::d3d11
 				continue;
 			}
 
-			if (snapshot.counter.drawcalls >= best_info.counter.drawcalls)
+			if (snapshot.counter.drawcalls >= best_snapshot.counter.drawcalls)
 			{
-				best_info = snapshot;
+				best_snapshot = snapshot;
 			}
 		}
 
-		// Warning: Reference to this object is lost after leaving the scope.
-		// But that should be fine since either this tracker or the depth stencil should still have a reference on it left so that it stays alive.
-		std::pair<ID3D11DepthStencilView *, ID3D11Texture2D *> response = {best_info.depthstencil.get(), best_info.texture.get()};
-
-		return response;
+		return best_snapshot;
 	}
 }
--- a/source/d3d11/draw_call_tracker.cpp
+++ b/source/d3d11/draw_call_tracker.cpp
@@ -10,18 +10,14 @@ namespace reshade::d3d11
 		_global_counter.vertices += source.total_vertices();
 		_global_counter.drawcalls += source.total_drawcalls();
 
-		for (auto &[depthstencil, snapshot] : source._counters_per_used_depthstencil)
+		for (const auto &[depthstencil, snapshot] : source._counters_per_used_depthstencil)
 		{
-			const auto destination_entry = _counters_per_used_depthstencil.find(depthstencil);
-
-			_counters_per_used_depthstencil[depthstencil].counter.vertices += snapshot.counter.vertices;
-			_counters_per_used_depthstencil[depthstencil].counter.drawcalls += snapshot.counter.drawcalls;
+			_counters_per_used_depthstencil[depthstencil].stats.vertices += snapshot.stats.vertices;
+			_counters_per_used_depthstencil[depthstencil].stats.drawcalls += snapshot.stats.drawcalls;
 		}
 
-		for (auto &[buffer, snapshot] : source._counters_per_constant_buffer)
+		for (const auto &[buffer, snapshot] : source._counters_per_constant_buffer)
 		{
-			const auto destination_entry = _counters_per_constant_buffer.find(buffer);
-
 			_counters_per_constant_buffer[buffer].vertices += snapshot.vertices;
 			_counters_per_constant_buffer[buffer].drawcalls += snapshot.drawcalls;
 			_counters_per_constant_buffer[buffer].pixelshaders += snapshot.pixelshaders;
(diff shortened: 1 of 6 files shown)


Matsilagi:
Crosire now has little time to see those (i think) but i would totally support if you kept going on with your own branch (as in, rebasing on your own and make it usable even on the meantime)

ReShade would benefit a lot from those indeed, allowing for lots of advanced (and currently) impossible effects to be added on most games.

If you really move on with this, i can gladly do testing and try to find shaders that tests those features if they get properly implemented for usage.

Matsilagi:
Oh I see now, hopefully it wouldn't be too hard for him (or you) to add those to the other renderers, this might really spark even more life on ReShade

el3ment:
@FransBouma it might be worth CC-ing into this pull request for a look over. I didn't realize you were responsible for the original draw call tracker.

FransBouma:
@el3ment Not sure, Crosire has refactored the code quite a bit since my initial commit. One thing that comes to mind is (looking at what this PR should do): you track clear actions on the tracked buffers as well? As a cleared buffer isn't really useful for postprocessing so you might want to postpone the clear till buffer flip

source/d3d11/draw_call_tracker.cpp
```diff
@@ -1,105 +1,168 @@
...
+	void draw_call_tracker::on_draw(ID3D11DeviceContext *context, UINT vertices) {
+		_global_counter.vertices += vertices;
+		_global_counter.drawcalls += 1;
 
-		const auto counters = _counters_per_used_depthstencil.find(depthstencil);
+		ID3D11RenderTargetView *rendertargets[D3D11_SIMULTANEOUS_RENDER_TARGET_COUNT] = { nullptr };
+		ID3D11DepthStencilView *depthstencil = nullptr;
 
-		if (counters != _counters_per_used_depthstencil.end())
+		context->OMGetRenderTargetsAndUnorderedAccessViews(D3D11_SIMULTANEOUS_RENDER_TARGET_COUNT, rendertargets, &depthstencil, 0, 0, nullptr);
```
crosire: Seems like `OMGetRenderTargets` would do the job here as well, since the UAVs aren't used.

source/d3d11/draw_call_tracker.cpp
```diff
@@ -1,105 +1,168 @@
...
-				texture = depthstencil_info.texture;
-			}
-			else
+			if (snapshot.texture == nullptr)
 			{
+				com_ptr<ID3D11Texture2D> texture;
 				com_ptr<ID3D11Resource> resource;
 				depthstencil->GetResource(&resource);
 
 				if (FAILED(resource->QueryInterface(&texture)))
```
crosire: Might as well just do `if (FAILED(resource->QueryInterface(&snapshot.texture)))` and instead of using a temporary `texture` variable.

source/d3d11/draw_call_tracker.cpp
```diff
@@ -1,105 +1,168 @@
...
+		for (unsigned int i = 0; i < D3D11_COMMONSHADER_CONSTANT_BUFFER_API_SLOT_COUNT; i++) {
+			// uses the default drawcalls = 0 the first time around.
+			if (pscbuffers[i] != nullptr) {
+				_counters_per_constant_buffer[pscbuffers[i]].pixelshaders += 1;
+			}
+		}
+
+	}
+
+	void draw_call_tracker::on_map(ID3D11DeviceContext *context, ID3D11Resource *pResource) {
```
crosire: I'm not a big fan of that code style for functions with long bodies: =P
```c++
void function() {
}
```
Much rather would like this:
```c++
void function()
{
}
```

source/d3d11/draw_call_tracker.cpp
```diff
@@ -1,105 +1,168 @@
...
+
+		D3D11_RESOURCE_DIMENSION dim;
+		pResource->GetType(&dim);
+
+		if (dim == D3D11_RESOURCE_DIMENSION::D3D11_RESOURCE_DIMENSION_BUFFER) {
+			_counters_per_constant_buffer[reinterpret_cast<ID3D11Buffer*>(pResource)].mapped += 1;
```
crosire: This should be a `static_cast`, not a `reinterpret_cast`. Also, `D3D11_RESOURCE_DIMENSION::D3D11_RESOURCE_DIMENSION_BUFFER` is doubled, just `D3D11_RESOURCE_DIMENSION_BUFFER` is enough and doesn't lose any information.

source/d3d11/draw_call_tracker.cpp
```diff
@@ -1,105 +1,168 @@
...
-		_counters.vertices += source.vertices();
-		_counters.drawcalls += source.drawcalls();
+		_global_counter.vertices += source.total_vertices();
+		_global_counter.drawcalls += source.total_drawcalls();
 
-		for (auto source_entry : source._counters_per_used_depthstencil)
+		for (auto &[depthstencil, snapshot] : source._counters_per_used_depthstencil)
 		{
-			const auto destination_entry = _counters_per_used_depthstencil.find(source_entry.first);
+			const auto destination_entry = _counters_per_used_depthstencil.find(depthstencil);
```
crosire: This isn't used? Looks like the original intention was to write `destination_entry->second.counter.vertices += ...` below instead of `_counters_per_used_depthstencil[depthstencil].counter.vertices += ...`, but that's not the case here. Same thing a couple of lines below.

source/d3d11/draw_call_tracker.cpp
```diff
@@ -1,105 +1,168 @@
...
 {
 	void draw_call_tracker::merge(const draw_call_tracker& source)
 	{
-		_counters.vertices += source.vertices();
-		_counters.drawcalls += source.drawcalls();
+		_global_counter.vertices += source.total_vertices();
+		_global_counter.drawcalls += source.total_drawcalls();
 
-		for (auto source_entry : source._counters_per_used_depthstencil)
+		for (auto &[depthstencil, snapshot] : source._counters_per_used_depthstencil)
```
crosire: This can be `for (const auto &[depthstencil, snapshot] : source._counters_per_used_depthstencil)` I think. Same for a couple of places below. I prefer to always make things `const` if possible.

source/d3d11/draw_call_tracker.hpp
```diff
@@ -9,32 +9,43 @@ namespace reshade::d3d11
 	class draw_call_tracker
 	{
 	public:
-		UINT vertices() const { return _counters.vertices; }
-		UINT drawcalls() const { return _counters.drawcalls; }
+		struct draw_counter {
```
crosire: Maybe rename this to something like `draw_info` or `draw_data` or `draw_stats`, since it's just data and doesn't contain any functionality that would warrant the `counter` name. Also, extra points for style: =)
```c++
struct draw_stats
{
};
```

el3ment:
@crosire these look great -- I'll push them up tomorrow or Tuesday!.

el3ment:
Sorry it took a few days longer than expected. I have made the changes you talked about, then rebased against your current master which should make the pull request easier and also let me add debugging stats in the DX11 menu to better demonstrate that the intermediate/constant buffer stuff is working properly.

As a side note - I have never gotten D3D11_CREATE_DEVICE_DEBUG to work on my machine. I have DX12, and I've installed the Graphics Tool optional stuff but it always crashes when I have that turned on (that's why commits like [f76aad3 ](https://github.com/crosire/reshade/pull/49/commits/f76aad39c169687c17fac14a8d32796f2446eed0) and [c03209 ](https://github.com/crosire/reshade/pull/49/commits/c032c098dcfb0a234caa0eaac438c2a2ab121557)
always bookend my stuff)

crosire:
Something with the render target tracking causes GTA5 to crash on start. I haven't figured out why. It seems to crash in GTA5 code and not ReShade, but commenting out `on_draw` and the bottom of `track_rendertargets` stops it from crashing.

el3ment:
sounds like GTA5 is releasing a pointer and I need to AddRef somewhere

el3ment:
@crosire I was able to load up GTA5 - is there anything special I should do to recreate your crash?

crosire:
I fixed some leaking references (OMGetRenderTargets and friends increment the reference count of the returned objects, but since the default constructor of `com_ptr` does so too, they were never released again).

But the crash is still observable. I'm not doing anything special apart from launching GTA 5 in the background and then focus the window once it is up. It crashes fractions of a second later. The culprint is the additional view tracking. Commenting out the following code fixes it:
```
for (UINT i = 0; i < num_views; i++)
{
	// If the render target isn't being tracked, this will create it
	_counters_per_used_depthstencil[depthstencil].additional_views[views[i]].drawcalls += 1;
}
```
But I don't understand why, since like I said the default constructor of `com_ptr` increments the reference count of each view (in this case we want this, since they were passed in via `OMSetRenderTargets` and have no additional reference to them), so it should not be possible for them to be released prematurely somewhere.

el3ment:
I have seen instances when num_views is larger than it should be - maybe we try an if statement `if (views[i] != nullptr)` 

Another tricky bit is that incrementing drawcalls might not be the actual problem - but rather later in the pipeline (in particular the "find best" function) stuff is done when drawcalls is not 0.

crosire:
I tried to check for `nullptr` before and it didn't change things (nor should it really, considering that `com_ptr` has `nullptr` checks internally).
I also tried commenting out the other if statement that works on the views, but that didn't help either, so it really is those few lines. And it has to be something with reference counting, otherwise I can't explain why it would crash in GTA and not ReShade. I just don't get why.

crosire:
Nope, disabling the whole method still leads to a crash :disappointed:.

crosire:
It's most definitly reference counting. Replacing `std::map<com_ptr<ID3D11RenderTargetView>, draw_stats> additional_views;` with `std::map<ID3D11RenderTargetView *, draw_stats> additional_views;` works.

PR #58, opened 2018-08-20
 Fix for shaders download in windows 7

Thalixte:
A user reported me that the reshade installer was unable to download the shaders from the github repo, using the uri https://github.com/crosire/reshade-shaders/archive/master.zip.

The pb seems to be related to the version of the TLS protocol supported in the .Net framework.

@Crosire, do you use the .Net framework 4.5 ? If it is the case, maybe consider upgrading to the .Net  framework 4.6.2, as it supports both TLS 1.1 and TLS 1.2 protocols.


crosire:
I'm using .NET 4.7 for development actually. The target for the setup tool is set to 4.6. And I'm running Windows 7 and don't have problems, so those users seem to have their security set up incorrectly (missing important updates that add support for newer TLS versions). I'd rather not set the default protocols to something fixed to work around weak security.

Thalixte:
OK. So no need to add those lines...

PR #59, opened 2018-10-13
Add option to save preset and configuration at screenshot

seri14:
ReShade64 2018-10-13 20-55-14
**[x]** Include Preset (\*.ini)
**[ ]** Include Configuration (*.ini)

ReShade64 2018-10-13 20-55-17
**[x]** Include Preset (\*.ini)
**[x]** Include Configuration (*.ini)

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -763,6 +763,8 @@ namespace reshade
 		config.get("GENERAL", "TutorialProgress", _tutorial_index);
 		config.get("GENERAL", "ScreenshotPath", _screenshot_path);
 		config.get("GENERAL", "ScreenshotFormat", _screenshot_format);
+		config.get("GENERAL", "ScreenshotIncludePreset", _screenshot_include_preset);
+		config.get("GENERAL", "ScreenshotIncludeConfiguration", _screenshot_include_configuration);
 		config.get("GENERAL", "ShowClock", _show_clock);
 		config.get("GENERAL", "ShowFPS", _show_framerate);
 		config.get("GENERAL", "FontGlobalScale", _imgui_context->IO.FontGlobalScale);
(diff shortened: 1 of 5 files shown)


source/runtime.cpp
```diff
@@ -861,7 +863,11 @@ namespace reshade
 	}
 	void runtime::save_config() const
 	{
-		ini_file config(_configuration_path);
+		save_config(_configuration_path);
+	}
+	void runtime::save_config(const filesystem::path &save_path) const
+	{
+		ini_file config(_configuration_path,save_path);
```
crosire: I don't think this is necessary. `ini_file config(save_path)` should have the same effect (since the empty `save_config` overload passes in the correct default path) and it means the changes to the `ini_file` class can be removed.

source/runtime.cpp
```diff
@@ -960,7 +968,11 @@ namespace reshade
 	}
 	void runtime::save_preset(const filesystem::path &path) const
 	{
-		ini_file preset(path);
+		save_preset(path, path);
+	}
+	void runtime::save_preset(const filesystem::path &path, const filesystem::path &save_path) const
+	{
+		ini_file preset(path, save_path);
```
crosire: See above. There should not be a need for the original preset path, since all required information are set in this function anyway.

source/runtime.cpp
```diff
@@ -1081,6 +1093,23 @@ namespace reshade
 		if (!success)
 		{
 			LOG(ERROR) << "Failed to write screenshot to " << path << "!";
+			return;
+		}
+
+		if (_screenshot_include_preset)
+		{
+			if (_current_preset >= 0)
```
crosire: This can be combined into a single branch `_screenshot_include_preset && _current_preset >= 0`.

source/runtime.cpp
seri14: If there is no overload, some settings will always be lost and in certain cases some settings will be lost.

Always:
NoFontScaling
(and unknown keys)

In certain cases:
DX*_BUFFER_DETECTION

source/runtime.cpp
seri14: If there is no overload, Completely disabled effect settings will be lost.

PR #62, opened 2018-11-07
Forum registration broken

Bluscream:
Since you disabled issues, i have to tell it to you this way. But the "enforced" way to discuss reshade is not working at the moment, i tried registering several times but it always threw me an error.


crosire:
1) There is a contact formular on the website. GitHub is the last place for these things.
2) I just tested it and can create a new account without issues.

PR #64, opened 2018-11-11
Shift + Right click to reset UI variable

seri14:
https://reshade.me/forum/suggestions/4243-shift-right-click-to-reset-ui-variable


crosire:
Nice and tidy solution. I like it.

PR #67, opened 2018-12-06
Add option to filter active effects, move errors/warnings into CollapsingHeader

Daodan317081:
Modified statistics->effects a bit.
- Added option to show only active effects
- Moved compiler errors/warnings into a collapsing header

### DIFF
--- a/source/gui.cpp
+++ b/source/gui.cpp
@@ -22,9 +22,6 @@ extern std::filesystem::path g_target_executable_path;
 const ImVec4 COLOR_RED = ImColor(240, 100, 100);
 const ImVec4 COLOR_YELLOW = ImColor(204, 204, 0);
 
-const ImVec4 COLOR_RED_HOVER = ImColor(240, 130, 130);
-const ImVec4 COLOR_YELLOW_HOVER = ImColor(204, 204, 100);
-
 void reshade::runtime::init_ui()
 {
 	// Default shortcut: Home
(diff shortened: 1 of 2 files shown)


Daodan317081:
What would be the best way to make the colors for warnings/errors customizable? I just noticed that the hardcoded COLOR_YELLOW is just barely readable with the 'Light' style.

crosire:
Could use the ones from the code editor color list instead. It already has settings for an "error" and a "warning" color.

Daodan317081:
I tried different things out and just having the collapsing header colored while keeping the compiler output at the default text color is the most usable approach (best readability). One addition that would be nice though is to color the background of the compiler text in red or yellow (so that it looks the same as the collapsing header). But I'm not sure how to do that.

crosire:
I like the "Show only active techniques" filter. But I'm not convinced about the collapsing headers yet. To get error/warning information you now have to do one extra click. And you have to do that on every recompile. Feels like that would get annoying quickly. The collapsing headers could be made open by default, but that kind of defeats their purpose then, since it would looks almost exactly like before.

Daodan317081:
Good point about the collapsing headers. I added this because I was annoyed by warnings that get repeated multiple times (mostly from statements in for-loops, "KNearestNeighbours" from Denoise.fx is such a candidate). I guess a better approach would be to filter the compiler output and print only one of the repeated lines with the additional information about how many times that line repeats. 

But if you want the "Show only active techniques" filter merge only that one commit and reject the other ones. I wouldn't mind that.

crosire:
Done. Thanks!

PR #70, opened 2019-01-08
Support to preset search path

seri14:
This request contains these features.

* Add setting "Preset Search Paths"
* Determinate the absolute path from relative path (for preset search path only)


seri14:
I noticed reason to my modification that filename() == DefaultPreset.ini

PR #71, opened 2019-01-09
Fix to change processing of remove button of list type UI

seri14:
If you empty the setting value, the line is also deleted at the same time.


crosire:
I'm sorry, but I prefer the UI the way it looks right now (with the "-" buttons).

PR #72, opened 2019-01-21
New depthbuffer detection techniques for d3d9

Thalixte:
Add new depthbuffer detection techniques for d3d9, allowing some goold old games to get depth buffer access (Mass Effect trilogy, Half Life 2, TES Oblivion, Medal of Honour [2010], Resident Evil 4 Ultimate HD, Mirror's Edge, ...)


crosire:
First thing to notice: There are a couple of places where merging failed (whitespaces different, and there is a new submodule that should not be in there). Would be cool if you could fix those!

Thalixte:
Fixed i guess

source/d3d9/d3d9_device.cpp
```diff
@@ -456,7 +456,24 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice9::EndScene()
 }
 HRESULT STDMETHODCALLTYPE Direct3DDevice9::Clear(DWORD Count, const D3DRECT *pRects, DWORD Flags, D3DCOLOR Color, float Z, DWORD Stencil)
 {
-	return _orig->Clear(Count, pRects, Flags, Color, Z, Stencil);
+	com_ptr<IDirect3DSurface9> depthstencil;
+	_orig->GetDepthStencilSurface(&depthstencil);
```
crosire: This call can be moved inside the first `if` below, so that it's not executed when only clearing the color buffer.

source/d3d9/runtime_d3d9.cpp
```diff
@@ -281,6 +307,18 @@ namespace reshade::d3d9
 		if (FAILED(_device->BeginScene()))
 			return;
 
+		if (_preserve_depth_buffer)
+		{
+			if (_auto_preserve)
```
crosire: `_preserve_depth_buffer && _auto_preserve`

source/d3d9/runtime_d3d9.cpp
```diff
@@ -378,19 +449,157 @@ namespace reshade::d3d9
 
 		if (depthstencil != nullptr)
 		{
-			if (depthstencil == _depthstencil_replacement)
+			if (_preserve_depth_buffer)
```
crosire: This can be moved inside the `if` below.

source/d3d9/runtime_d3d9.cpp
```diff
@@ -378,19 +449,157 @@ namespace reshade::d3d9
...
+	}
+	void runtime_d3d9::before_clear(com_ptr<IDirect3DSurface9> depthstencil)
+	{
+		if (depthstencil == nullptr)
+			return;
+
+		if (!_preserve_depth_buffer)
+			return;
+
+		if (depthstencil != get_depthstencil_replacement())
```
crosire: These can be combined into a single `if`.

source/d3d9/runtime_d3d9.cpp
```diff
@@ -271,6 +296,19 @@ namespace reshade::d3d9
 		}
 
 		_depth_source_table.clear();
+
+		// reset the new tables
+		_depth_buffer_table.clear();
+
+		if (!_depth_clearing_table.empty())
```
crosire: This `if` is not necessary since the `for` loop doesn't do anything if the table is empty anyway.

source/d3d9/runtime_d3d9.cpp
```diff
@@ -378,19 +484,149 @@ namespace reshade::d3d9
...
+					it->second.drawcall_count = _drawcalls;
+					it->second.vertices_count += vertices;
+				}
+			}
+			else
+			{
+				_preserve_drawcalls++;
+				_preserve_vertices += vertices;
+
+				depthstencil = get_depthstencil_replacement();
```
crosire: This line doesn't do anything, since `depthstencil` is not used afterwards.

source/d3d9/runtime_d3d9.cpp
```diff
@@ -378,19 +484,149 @@ namespace reshade::d3d9
...
+			_preserve_starting_index = get_best_preserve_starting_index(true);
+		}
+
+		if (is_preserved())
+		{
+			// if the current depth buffer replacement texture has to be preserved, replace the depthStencilSurface ref with a dummy one, so the depth buffer replacement texture will not be cleared
+			_device->SetDepthStencilSurface(_depthstencil.get());
+		}
+	}
+	void runtime_d3d9::after_clear(com_ptr<IDirect3DSurface9> depthstencil)
```
crosire: I don't think this has to be this complicated. Instead of splitting this code up into two functions `before_clear` and `after_clear`, how about using a single one that returns a bool which specifies whether to execute the clear operation or not. If it is false, return from `Direct3DDevice9::Clear` without calling the original device `Clear`. This way you are saving a lot of duplicated checks and there is no need to replace the currently set depth stencil surface with another one before clear.

source/d3d9/runtime_d3d9.cpp
```diff
@@ -1394,4 +1767,164 @@ namespace reshade::d3d9
...
+			{
+				return false;
+			}
+		}
+
+		return true;
+	}
+
+	// preservation decision
+	bool runtime_d3d9::is_preserved()
```
crosire: I'd inline this and use something like
```
if (_multi_depthstencil ?
    // comment
    _clear_idx <= _preserve_starting_index :
    // comment
    _clear_idx >= _preserve_starting_index)
{
}
```
Together with above suggestion of moving `before_clear` and `after_clear` into a single function, this would only be called once anyway.

source/d3d9/runtime_d3d9.cpp
```diff
@@ -1394,4 +1767,164 @@ namespace reshade::d3d9
...
+			}
+
+			if (best)
+				result = it.first;
+		}
+
+		return result;
+	}
+
+	bool runtime_d3d9::check_depthstencil_size(com_ptr<IDirect3DSurface9> depthstencil)
```
crosire: Everywhere this is called, `GetDesc` is called either before or soon after. So to avoid duplicated calls, just change this function signature to `bool check_depthstencil_size(const D3DSURFACE_DESC &desc);` and pass in the description in the call.

source/d3d9/runtime_d3d9.cpp
```diff
@@ -1268,38 +1633,66 @@ namespace reshade::d3d9
...
+					_preserve_selected_index = it.first;
+					best_match = it.second.depthstencil.get();
+					vertices = it.second.vertices_count;
+					drawcalls = it.second.drawcall_count;
+				}
+			}
 		}
 
-		if (best_match != nullptr && _depthstencil != best_match)
+		if (best_match != nullptr && (_depthstencil != best_match))
```
crosire: These new parantheses mess up Git's diff =P

source/d3d9/runtime_d3d9.cpp
```diff
@@ -1394,4 +1767,164 @@ namespace reshade::d3d9
...
+
+		if (_disable_depth_buffer_size_restriction)
+		{
+			width = _width;
+			height = _height;
+		}
+
+		const HRESULT hr = _device->CreateTexture(width, height, 1, D3DUSAGE_DEPTHSTENCIL, desc.Format, D3DPOOL_DEFAULT, &depthstencil_replacement, nullptr);
+
+		if (!SUCCEEDED(hr))
```
crosire: `if (FAILED(hr))`

source/d3d9/runtime_d3d9.cpp
```diff
@@ -1394,4 +1767,164 @@ namespace reshade::d3d9
...
+
+			return nullptr;
+		}
+
+		UINT width = desc.Width;
+		UINT height = desc.Height;
+
+		if (_disable_depth_buffer_size_restriction)
+		{
+			width = _width;
```
crosire: Why this?

source/d3d11/d3d11_device_context.cpp
```diff
@@ -449,7 +449,8 @@ void STDMETHODCALLTYPE D3D11DeviceContext::ClearUnorderedAccessViewFloat(ID3D11U
 void STDMETHODCALLTYPE D3D11DeviceContext::ClearDepthStencilView(ID3D11DepthStencilView *pDepthStencilView, UINT ClearFlags, FLOAT Depth, UINT8 Stencil)
 {
 #if RESHADE_DX11_CAPTURE_DEPTH_BUFFERS
-	track_cleared_depthstencil(pDepthStencilView);
+	if(ClearFlags & D3D11_CLEAR_DEPTH)
```
crosire: Formatting strikes again! =) `if (`

source/d3d9/d3d9_device.cpp
```diff
@@ -456,7 +456,27 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice9::EndScene()
...
+
+	const HRESULT hr = _orig->Clear(Count, pRects, Flags, Color, Z, Stencil);
+
+	if ((Flags & D3DCLEAR_ZBUFFER) != 0)
+	{
+		// the depth buffer has been cleared, add a post action to eventually bring back the preserved depth buffer
+		_implicit_swapchain->_runtime->after_clear(depthstencil);
+	}
+
+	return hr;
```
crosire: Suggestion as discussed below:
```
if (Flags & D3DCLEAR_ZBUFFER)
{
	com_ptr<IDirect3DSurface9> depthstencil;
	_orig->GetDepthStencilSurface(&depthstencil);

	if (!_implicit_swapchain->_runtime->on_clear(depthstencil))
		return D3D_OK; // Avoid clearing this depth stencil if it should be preserved
}

return _orig->Clear(Count, pRects, Flags, Color, Z, Stencil);
```

source/d3d9/runtime_d3d9.cpp
```diff
@@ -378,19 +484,149 @@ namespace reshade::d3d9
...
+			else
+			{
+				_preserve_drawcalls++;
+				_preserve_vertices += vertices;
+
+				depthstencil = get_depthstencil_replacement();
+			}
+		}
+	}
+	void runtime_d3d9::before_clear(com_ptr<IDirect3DSurface9> depthstencil)
```
crosire: Make the argument a `const com_ptr<IDirect3DSurface9> &depthstencil` to avoid some `AddRef` and `Release` calls.

source/d3d9/runtime_d3d9.cpp
```diff
@@ -1268,38 +1633,66 @@ namespace reshade::d3d9
...
+					continue;
+				}
+				else
+				{
+					++it;
+				}
+
+				if (depthstencil_info.drawcall_count == 0)
+				{
+					if (!_preserve_depth_buffer)
```
crosire: `if (depthstencil_info.drawcall_count == 0 && !_preserve_depth_buffer)`

source/d3d9/runtime_d3d9.hpp
```diff
@@ -94,13 +117,31 @@ namespace reshade::d3d9
...
+		int _db_drawcalls = 0;
+		int _previous_best_vertices = 0;
+		int _previous_best_drawcalls = 0;
 		D3DFORMAT _backbuffer_format = D3DFMT_UNKNOWN;
 		com_ptr<IDirect3DStateBlock9> _app_state;
 		com_ptr<IDirect3DSurface9> _depthstencil;
 		com_ptr<IDirect3DSurface9> _depthstencil_replacement;
 		com_ptr<IDirect3DSurface9> _default_depthstencil;
 		std::unordered_map<IDirect3DSurface9 *, depth_source_info> _depth_source_table;
+		std::map<int, depth_buffer_info> _depth_buffer_table;
```
crosire: Why does it need both of these tables? They seem to hold very similar information.

source/d3d9/runtime_d3d9.cpp
```diff
@@ -1236,19 +1471,149 @@ namespace reshade::d3d9
...
 				_depthstencil = nullptr;
+				// Force depth-stencil clearing table recreation
+				_depth_buffer_table.clear();
+				_depth_clearing_table.clear();
 			}
 
-			for (const auto &it : _depth_source_table)
+			if (_preserve_depth_buffer)
+			{
+				if (ImGui::Checkbox("multiple depthstencil replacement buffer", &_multi_depthstencil))
```
crosire: Does this help in some games and why? It adds a lot of extra complication.

source/d3d9/runtime_d3d9.hpp
```diff
@@ -94,13 +117,31 @@ namespace reshade::d3d9
...
+		std::map<int, depth_clearing_info> _depth_clearing_table;
```
crosire: Do these have to be `map`s? Looks to me like the clear indices are always linear, so a `vector` would do?

crosire:
This is much more combrehensible now, great job!

source/gui.cpp
```diff
@@ -337,8 +337,10 @@ void reshade::runtime::draw_ui()
 	imgui_io.Fonts->TexID = _imgui_font_atlas.impl.get();
 
 	// Scale mouse position in case render resolution does not match the window size
-	imgui_io.MousePos.x = _input->mouse_position_x() * (imgui_io.DisplaySize.x / _window_width);
-	imgui_io.MousePos.y = _input->mouse_position_y() * (imgui_io.DisplaySize.y / _window_height);
+	// imgui_io.MousePos.x = _input->mouse_position_x() * (imgui_io.DisplaySize.x / _window_width);
+	// imgui_io.MousePos.y = _input->mouse_position_y() * (imgui_io.DisplaySize.y / _window_height);
+	imgui_io.MousePos.x = static_cast<float>(_input->mouse_position_x());
+	imgui_io.MousePos.y = static_cast<float>(_input->mouse_position_y());
```
crosire: Just one more thing: What's up with this?

source/d3d9/runtime_d3d9.cpp
Thalixte: In some games, it helped (for instance Oblivion), but we can wait to implement it

source/gui.cpp
Thalixte: It does not work in the games i tested. Maybe it is dx9 related (did not check on dx10/dx11 games).
In my rig, the mouse does not do anything on the ui. Replacing the new code with the old one fix it.
For info, i tested on Mass Effect 2, Outlast, Half life 2...

crosire: This is necessary to fix compatibility with downsampling. Check the values of `_window_width` and `_window_height` to make sure they contain correct values. If not will need to add some code that fixes those values, but not by reverting this.

crosire: I'm guessing that this line here doesn't do the correct thing in those games: https://github.com/crosire/reshade/blob/81ee29a6456de172337c1af2064d92764e6db087/source/d3d9/runtime_d3d9.cpp#L242

crosire: Also, I cannot reproduce it not working in Half-Life 2 ... Works fine for me there.

Thalixte: Yes, in Half-Life 2 it works, but still not in Mass Effect 2, nor in Outlast, nor in Batman Arkham Asylum

crosire: Works in Mass Effect 2 too. I don't have those other games to test.

PR #73, opened 2019-01-23
Fixed a wrong tab focus

seri14:


crosire:
Looks like we found that at the same time. Fixed it exactly like this in https://github.com/crosire/reshade/commit/3d5e60f238f52220aeee9aad65b3ff5b516cda83.

PR #76, opened 2019-01-30
Add warning to home page if the screenshot directory doesn't exist

Daodan317081:
So it is less likely to loose shots when something is not set up correctly (typo in the screenshot path, folder structure changed, etc.)

### DIFF
--- a/source/gui.cpp
+++ b/source/gui.cpp
@@ -571,6 +571,13 @@ void reshade::runtime::draw_overlay_menu_home()
 	if (!_effects_enabled)
 		ImGui::Text("Effects are disabled. Press '%s' to enable them again.", input::key_name(_effects_key_data).c_str());
 
+	if (!std::filesystem::exists(_screenshot_path))
+	{
+		ImGui::PushStyleColor(ImGuiCol_Text, COLOR_RED);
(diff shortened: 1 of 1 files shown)


source/gui.cpp
```diff
@@ -571,6 +571,13 @@ void reshade::runtime::draw_overlay_menu_home()
 	if (!_effects_enabled)
 		ImGui::Text("Effects are disabled. Press '%s' to enable them again.", input::key_name(_effects_key_data).c_str());
 
+	if (!std::filesystem::exists(_screenshot_path))
+	{
+		ImGui::PushStyleColor(ImGuiCol_Text, COLOR_RED);
+		ImGui::Text("Unable to save screenshots: directory '%s' doesn't exist!", _screenshot_path.string().c_str());
```
crosire: Good idea =) Some improvements:
- This can be simplified with `ImGui::TextColored(...);`.
- Also, it'd change the text to `Unable to save screenshots because path doesn't exist: %s` so that it is readable even when the UI is very narrow (on small resolutions).
- If you use `u8string()` instead of `string()` ImGui will be able to render text for paths with unicode characters.

Daodan317081:
I've added the improvements :)
Thanks!

PR #77, opened 2019-02-05
Change error check and display condition of screenshot path

seri14:


seri14:
Why does the user have to open the home tab for error checking when modied path?

Bluscream:
Why not just create the path if it does not exist?

seri14:
This change (and original pull request) is particularly effective because notify error when got ReShade.ini from friends or internet.

seri14:
Description of #76 included "folder structure changed", but lost it.

seri14:
@Bluscream I think good to make pull request or post to forum.

PR #79, opened 2019-02-06
Not display error of screenshot when screenshot feature is disabled

seri14:


source/gui.cpp
```diff
@@ -571,7 +571,7 @@ void reshade::runtime::draw_overlay_menu_home()
 	if (!_effects_enabled)
 		ImGui::Text("Effects are disabled. Press '%s' to enable them again.", input::key_name(_effects_key_data).c_str());
 
-	if (!_screenshot_path_exists)
+	if (!_screenshot_path_exists && (_screenshot_key_data[0] != 0 || _screenshot_path != ""))
```
crosire: I'd use `_screenshot_path.empty()` instead of `_screenshot_path != ""`, since it's faster.

source/gui.cpp
```diff
@@ -571,7 +571,7 @@ void reshade::runtime::draw_overlay_menu_home()
 	if (!_effects_enabled)
 		ImGui::Text("Effects are disabled. Press '%s' to enable them again.", input::key_name(_effects_key_data).c_str());
 
-	if (!_screenshot_path_exists && (_screenshot_key_data[0] != 0 || _screenshot_path != ""))
+	if (!_screenshot_path_exists && (_screenshot_key_data[0] != 0 || !_screenshot_path.empty()))
```
crosire: Now that I think about it, shouldn't this be `if (!_screenshot_path_exists && _screenshot_key_data[0] != 0)`? The path needs to be valid regardless of whether it is empty or not if a screenshot key is bound.

seri14: Should I change title and code on this pull request?
If apply it idea, more require changes to some other functions.

PR #81, opened 2019-02-08
Support unicode character set for the path

seri14:
From case: https://twitter.com/SkylarTnT/status/1093440629701595142

Avoided setting types changes to `std::wstring`.


crosire:
`std::wstring_convert()` is deprecated in C++17 hence why it was removed in previous commits. There is however `std::filesystem::u8path()` which can be used to construct a `std::filesystem::path` from a UTF-8 sequence.

crosire:
There already is a UTF8-UTF16 converter in the ReShade codebase. Check deps/utfcpp.

seri14:
Why not used it

seri14:
thrown in the trash by very easy reason

seri14:
If you have answer to solve a clear way, you should teach it to me. I am not playing quiz game. I am not want waste of time.

crosire:
I told you a way to do it in one line: https://en.cppreference.com/w/cpp/filesystem/path/u8path ... Not hiding anything. deps/utfcpp is not even needed here. I just didn't understand why you would add another unicode parser if there already is one in.

seri14:
I noticed it now. I am sorry I misunderstood your comment.

PR #82, opened 2019-02-09
Suppress screenshot path error when succeed to take screenshot

seri14:
No comment

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -959,7 +959,7 @@ void reshade::runtime::save_current_preset() const
 		save_preset(_preset_files[_current_preset]);
 }
 
-void reshade::runtime::save_screenshot() const
+void reshade::runtime::save_screenshot()
 {
 	std::vector<uint8_t> data(_width * _height * 4);
 	capture_frame(data.data());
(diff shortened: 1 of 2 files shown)


crosire:
Is this necessary? `_screenshot_path_exists` should always be `true` already, or else `save_screenshot` will always fail. Unless I'm missing something?

Daodan317081:
Currently `_screenshot_path_exists` only gets updated when the settings change. So it its possible that since the last check - where `_screenshot_path_exists` is set to `false` - the file system has changed and the screenshot path is now valid.

seri14:
We already explained to you at #76, #77 and #79.
If not applied #76 and original of #77, occurs that @Daodan317081 said.
If not applied #79 (using @crosire 's #77), the `std::filesystem::exists(empty)` always returns `false`.

Daodan317081:
Now that I think about it:
Maybe it is better to just draw a message on the screen (string at the top left corner?) for a couple of seconds when the screenshot key was pressed.

* If saving was successful: "Screenshot saved to "C:\...."
* Otherwise: "Unable to save screenshot to "C:\..."

And if the screenshot path is invalid at startup add that information to the splash screen.

seri14:
I thinked it first, but I will not implement it. I learned the KISS principle.
It is the best, but it is not simple, requires more complex code.

crosire:
Not that much more complicated. I like it. Implemented.

PR #83, opened 2019-02-12
Change the implementation of "Active to top" to it does sorting by used

seri14:
This request contains these changes.
* Changed the implementation of "Active to top" to "Sort by used"
* When press "Sort by used", unused techniques will sorted by name

### DIFF
--- a/source/gui.cpp
+++ b/source/gui.cpp
@@ -735,7 +735,7 @@ void reshade::runtime::draw_overlay_menu_home()
 
 		ImGui::SameLine();
 
-		if (ImGui::Button("Sort by used", ImVec2(130 - _imgui_context->Style.ItemSpacing.x, 0)))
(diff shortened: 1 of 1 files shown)


source/gui.cpp
```diff
@@ -735,22 +735,38 @@ void reshade::runtime::draw_overlay_menu_home()
 
 		ImGui::SameLine();
 
-		if (ImGui::Button("Active to top", ImVec2(130 - _imgui_context->Style.ItemSpacing.x, 0)))
+		if (ImGui::Button("Sort by used", ImVec2(130 - _imgui_context->Style.ItemSpacing.x, 0)))
```
crosire: This is nice. But I'd keep the "Active to top" button description, even though it does more, simply for consistency with previous versions.

PR #87, opened 2019-02-14
Fix behavior to for duplicate variable name on shader code

seri14:


seri14:
and I have a question. Is the word "ouput" correct?
https://github.com/crosire/reshade/blob/fc831ae5e6ba4dc65e42821b386043ebea88285e/source/effect_codegen_glsl.cpp#L223

source/effect_codegen_glsl.cpp
```diff
@@ -205,18 +205,26 @@ class codegen_glsl final : public codegen
...
+			id = it->second;
 		assert(id != 0);
 		if (const auto it = _names.find(id); it != _names.end())
-			return is_function ? it->second : it->second + '_' + std::to_string(id);
+			return it->second;
 		return '_' + std::to_string(id);
 	}
 
+	template <bool is_unique = true>
+	std::string &define_name(const id id, const std::string &name)
```
crosire: The `name` parameter should best be a `std::string` instead of `const std::string &`, so that the move operator can work properly.

source/effect_codegen_glsl.cpp
```diff
@@ -205,18 +205,26 @@ class codegen_glsl final : public codegen
...
+	{
+		if constexpr (!is_unique)
+			for (auto it = _names.begin(); it != _names.end(); it++)
+				if (it->second == name)
+					return _names[id] = name + '_' + std::to_string(id);
+		return _names[id] = name;
```
crosire: With above fixed this can become `_names[id] = std::move(name)`.

source/effect_codegen_glsl.cpp
crosire: Also you only call `define_name` once now. All other sites still use the `_names` map directly. Those occurances should probably replaced with a call to this function now too.

crosire:
And good catch on the "ouput". You are correct. That should be "output".
The list was taken from the reserved words list in the GLSL spec: https://www.khronos.org/registry/OpenGL/specs/gl/GLSLangSpec.4.40.pdf section 3.6.

source/effect_codegen_glsl.cpp
```diff
@@ -233,8 +253,15 @@ class codegen_glsl final : public codegen
 
 		if (name.compare(0, 3, "gl_") == 0 || s_reserverd_names.count(name))
 			name += '_';
+
+		conform_naming_rule(name);
+	}
+
+	static std::string &conform_naming_rule(std::string &name)
```
crosire: Is there a reason this is a separate function now and not inlined anymore?

PR #88, opened 2019-02-14
Add reset operation of preset

seri14:
This working phase is the for proposal


crosire:
I'm not sure I see why this is necessary? One can just delete the preset and create a new one, which has the same effect.

seri14:
At the time of created request, the add button's behavior is only disable all techniques.
It is not erase toggle keys, and not reset uniform values.
If these problem has only about the uniform values, users may push the reset button.

But I noticed @Mortalitas 's #93, It changed to erase toggle keys before load the preset.
I am not testing about it merge with this commit, but I think it should have been resolved.

PR #89, opened 2019-02-17
Generate more readable hlsl/glsl code

seri14:
This request contains these changes
* Add naming type "formula" on effect parser
* Generate more readable hlsl/glsl code

### DIFF
--- a/source/effect_codegen_glsl.cpp
+++ b/source/effect_codegen_glsl.cpp
@@ -30,8 +30,8 @@ private:
 		unique,
 		// This is a special name that is not modified and should be unique
 		reserved,
-		// A formula token for referring as operand
-		formula,
+		// An expression token for referring as operand
+		expression,
 	};
 
 	id _next_id = 1;
@@ -232,13 +232,13 @@ private:
 	template <naming naming = naming::general>
 	void define_name(const id id, std::string name)
 	{
-		if constexpr (naming != naming::reserved && naming != naming::formula)
+		if constexpr (naming != naming::reserved && naming != naming::expression)
 			escape_name(name);
 		if constexpr (naming == naming::general)
 			if (std::find_if(_names.begin(), _names.end(), [&name](const auto &it) { return it.second == name; }) != _names.end())
 				name += '_' + std::to_string(id);
 		// Remove double underscore symbols from name which can occur due to namespaces but are not allowed in GLSL
-		if constexpr (naming != naming::formula)
+		if constexpr (naming != naming::expression)
 			for (size_t pos = 0; (pos = name.find("__", pos)) != std::string::npos; pos += 3)
 				name.replace(pos, 2, "_US");
 		_names[id] = std::move(name);
@@ -687,7 +687,7 @@ private:
 			}
 		}
 
-		define_name<naming::formula>(res, newcode);
+		define_name<naming::expression>(res, newcode);
 
 		return res;
 	}
@@ -762,7 +762,7 @@ private:
 
 		std::string code;
 		write_constant(code, type, data);
-		define_name<naming::formula>(res, code);
+		define_name<naming::expression>(res, code);
 
 		return res;
 	}
@@ -792,7 +792,7 @@ private:
 		}
 
 		code += '(' + id_to_name(val) + ')' + ')';
-		define_name<naming::formula>(res, code);
+		define_name<naming::expression>(res, code);
 
 		return res;
 	}
@@ -906,7 +906,7 @@ private:
 			code += id_to_name(lhs) + ' ' + operator_code + ' ' + id_to_name(rhs);
 
 		code += ')';
-		define_name<naming::formula>(res, code);
+		define_name<naming::expression>(res, code);
 
 		return res;
 	}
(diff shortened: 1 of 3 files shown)


source/effect_parser.cpp
```diff
@@ -873,7 +873,7 @@ bool reshadefx::parser::parse_expression_unary(expression &exp)
 
 				arguments[i].add_cast_operation(param_type);
 
-				if (symbol.op == symbol_type::function || param_type.has(type::q_out))
```
crosire: This is necessary and for compatibility with the SPIR-V code generator.

source/effect_parser.cpp
```diff
@@ -2666,16 +2666,10 @@ bool reshadefx::parser::parse_variable(type type, std::string name, bool global)
...
 
-			symbol.id = _codegen->define_variable(location, type, std::move(unique_name), global);
-
 			if (initializer_value != 0)
-			{
 				assert(!global); // Global variables cannot have a dynamic initializer
 
-				expression variable; variable.reset_to_lvalue(location, symbol.id, type);
-
-				_codegen->emit_store(variable, initializer_value);
```
crosire: This too is necessary for compatibility with SPIR-V. See the comment right above:
```The initializer expression for variables must be a constant```

File: source/effect_codegen_glsl.cpp [hunk 3]
```diff
@@ -30,6 +30,8 @@ class codegen_glsl final : public codegen
 		unique,
 		// This is a special name that is not modified and should be unique
 		reserved,
+		// A formula token for referring as operand
+		formula,
```
crosire: I suggest to rename this to `expression`.

source/effect_codegen_glsl.cpp
```diff
@@ -758,42 +749,29 @@ class codegen_glsl final : public codegen
...
+		std::string code;
 		write_constant(code, type, data);
-		code += ";\n";
+		define_name<naming::formula>(res, code);
 
 		return res;
 	}
 
-	id   emit_unary_op(const location &loc, tokenid op, const type &res_type, id val) override
+	id   emit_unary_op(const location &, tokenid op, const type &res_type, id val) override
```
crosire: I'm a bit on the fence on whether it is worth throwing out line information for better readability ... But I can see the appeal and it might even improve HLSL compile times, so will probably go for it.

source/effect_parser.cpp
seri14: What means "constant"?
Is this code is illegal?
```hlsl
#line 757
	float _1336 = smoothstep(_1334, _1335, discRadius);
```
I means it requies fixes to
```hlsl
#line 757
	float _1336; // avoid set non-constant value
#line 757
	_1336 = smoothstep(_1334, _1335, discRadius);
```

source/effect_parser.cpp
seri14: I am sorry I do not know what is the problem, so I simply reverted it.

source/effect_parser.cpp
crosire: This is not about HLSL or GLSL. This is about the SPIRV codegen backend. It requires input in a certain form. One of these requirements is that variables cannot have dynamic initializers (Google for SSA form if you want more information). So yes, this code is illegal in SPIR-V. It's fine in HLSL obviously, but that is not the point. ReShade needs to support all 3 codegen backends.

seri14: My example is took by current version. (4.1.1.496)
I think that it requires fix for current behavior.

crosire: I'll say it again. This break SPIR-V. Your changes work fine if you revert modifications to effect_parser.cpp. Please do so, I cannot merge this otherwise.

seri14:
This commit is last in this request because I want to give up.

source/effect_parser.cpp
seri14: Finally, the problem of initializer with call function, I can not modify it.

crosire: This is because for https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpFunctionCall the type of the arguments has to equal the type of the function parameters as defined by https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpFunctionParameter. Since the compiler registers those a pointers, the arguments have to be pointers as well. And the only way to obtain a pointer from an expression is to store the value of that expression in a variable first and then get the pointer to the variable. That's why the compiler generates temporary variables for each function argument.
This is not necessary for HLSL/GLSL of course, but the SPIR-V generator shares the same compiler frontend, so it is done for all of them.

seri14: Thank you for your instruction. It is very useful for learning.

crosire:
Finally got some time to check the performance implications of this. I compiled both the generated HLSL before this pull request and the generated HLSL after this pull request. As it turns out the HLSL compiler is able to generate better code without this pull request (less instructions and less register usage). Probably because some calculations are duplicated with this.
Therefore I won't be merging this unfortunately (more readable HLSL output can't be excused if runtime performance is a potentially bit worse, since most users won't ever look at the HLSL).
I incorporated all your other smaller changes to avoid some unnecessary parentheses in the generated code though.

PR #91, opened 2019-02-20
Add support of annotation "ui_label" for technique

seri14:
When you create BackColor and BackImage with added annotation "ui_label":
replace_1

Error case:
replace_2

### DIFF
--- a/source/effect_expression.hpp
+++ b/source/effect_expression.hpp
@@ -359,6 +359,7 @@ namespace reshadefx
 	struct technique_info
 	{
 		std::string name;
+		std::string display_name;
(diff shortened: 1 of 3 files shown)


seri14:
I want to ask why not support sort?
If it label's first character is not equals to name, you should will missing it while long time.

crosire:
The effect parser should not have knowledge about annotations, as such I couldn't use these changes directly (+ I wanted to avoid duplicating the same data in memory). And it was easier to just ignore sorting for now.

seri14:
Thank you for your review.
I will think another way.

PR #92, opened 2019-02-21
Changed not to share texture name among effects

seri14:
I do not know this is breaking change or not.


crosire:
The texture name being shared is intentional. There even are some effects which rely on this (to access data from other effects). So this would break them.

PR #94, opened 2019-02-24
Add button to clear the search bar

Daodan317081:
Adds a button to clear the search bar. Only pops up when text was entered.


crosire:
The feature I never knew I needed ... This has been super helpful the last couple of days, so thanks for proposing this =).

Bluscream:
<3

PR #96, opened 2019-03-03
Fix loading .fx file it have BOM

seri14:


source/effect_preprocessor.cpp
```diff
@@ -41,10 +41,15 @@ bool reshadefx::preprocessor::add_macro_definition(const std::string &name, std:
...
 {
 	std::ifstream file(path);
+	file.imbue(std::locale("en-us.UTF-8"));
 
 	if (!file.is_open())
 		return false;
 
+	// Remove BOM (0xefbbbf means 0xfeff)
+	if (file.get() != 0xef || file.get() != 0xbb || file.get() != 0xbf)
+		file.seekg(0, SEEK_SET);
```
crosire: `SEEK_SET` is C, not C++. This should be `std::ios::beg` :wink:.

PR #100, opened 2019-03-08
Add prev/next buttons for preset selector

seri14:
prevnext
** This is 4.2.1 based (I forgot to update version.h)


source/gui.cpp
```diff
@@ -622,6 +622,40 @@ void reshade::runtime::draw_overlay_menu_home()
...
+
+			save_config();
+
+			_show_splash = true;
+
+			// Need to reload effects in performance mode, so values are applied
+			if (_performance_mode)
+				load_effects();
+			else
+				load_preset(_preset_files[_current_preset]);
```
crosire: This is no longer necessary. A call to `load_current_preset_()` is sufficient, since it will automatically reload effects in performance mode.

crosire: However, regardless of this, I'm not sure if these two buttons would actually be that helpful (since you can't tell which preset gets selected when you press them). And they reduce the visible text in the combo box even further which can become a problem with long path names.

seri14: This request is from an another user, and I implemented it.

The user said :
* Tired of operation when they want to compare each presets with any tuned situations.
* I want to have experience same as the "list" UI.


I think the implementation of this request is still early, and maybe we should consider after implemented following features:
* Sorting by user for preset list.

and
* Float the full path while item hovered.

or
* Label of textbox is filename, list is full path.

or
* Feature of give alias to the preset. (by user input or rules)
* Change behaviour of the minus button to context menu. (has rename, and remove)

or something else.

crosire:
Closing this as it is superseded by #114.

seri14:
I have little more suggestions after #114. in new request?

crosire:
Jup

PR #103, opened 2019-03-19
Add support the ui_type "spinner"

seri14:
spinner


crosire:
Nice! But I'm not convinced by the name `spinner` yet. That shounds more like a radial menu. Maybe `list` instead? Also, I'd suggest moving this into a function in `gui_widgets.cpp`.

PR #107, opened 2019-03-24
Reduce temporary variables for list ui type

seri14:


crosire:
Whoa, why delete all these pull requests? I just didn't have the time to try them out before merging yet.

seri14:
I just understood why these keep open a month.

PR #108, opened 2019-03-26
fix parasite D3D9 depth buffers

Thalixte:
Found that, filtering the viewport, i can get rid of the parasite depth buffer drawcalls that can occur in some games (Outlast, Half-life 2, Mass Effect 2 cinematics...). Feel free to check it, Crosire ;-)

### DIFF
--- a/source/d3d9/d3d9_device.cpp
+++ b/source/d3d9/d3d9_device.cpp
@@ -439,6 +439,10 @@ HRESULT STDMETHODCALLTYPE Direct3DDevice9::MultiplyTransform(D3DTRANSFORMSTATETY
 }
 HRESULT STDMETHODCALLTYPE Direct3DDevice9::SetViewport(const D3DVIEWPORT9 *pViewport)
 {
+	assert(_implicit_swapchain != nullptr);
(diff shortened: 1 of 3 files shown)


crosire:
Why not simply do `_device->GetViewport` in those places instead of explicitly storing it via a `SetViewport` hook? =)

Thalixte:
As it is used on each drawcall, it is to avoid calling check_depthstencil_size method (and the associated tests) too many times =)

crosire:
Ah, I see.

crosire:
But to be honest, I don't think this would have any real affect on performance. And `GetViewport` is one of the few getters that even works when the device was created with `D3DCREATE_PUREDEVICE`: https://docs.microsoft.com/en-us/windows/desktop/api/d3d9/nf-d3d9-idirect3ddevice9-getviewport.

Thalixte:
Ok, as you like. Made the changes ;-)

Thalixte:
Add to revert back, because retrieving the viewport on the on_drawcall method, or on the on_clear one added parasite depth buffer in Outlast. It works better in auto preserve mode this way ...

PR #110, opened 2019-04-04
Support define of variable-like preprocessor definitions by user

seri14:
for proposal

setifdef

### DIFF
--- a/source/gui.cpp
+++ b/source/gui.cpp
@@ -1862,13 +1862,13 @@ void reshade::runtime::draw_overlay_variable_editor()
 
 						ImGui::PushID(name->c_str());
 
-						ImGui::PushItemWidth(ImGui::GetWindowContentRegionWidth() * 0.66666666f - (button_spacing));
-						ImGui::LabelText("##user_definition_name", name->c_str());
+						ImGui::PushItemWidth(ImGui::GetWindowContentRegionWidth() * 0.66666666f - button_spacing);
+						ImGui::LabelText("##user_definition_name", "%s", name->c_str());
 						ImGui::PopItemWidth();
 
 						ImGui::SameLine(0, button_spacing);
 
-						ImGui::PushItemWidth(ImGui::GetWindowContentRegionWidth() * 0.33333333f - (button_spacing + button_size) + 1);
+						ImGui::PushItemWidth(ImGui::GetWindowContentRegionWidth() * 0.33333333f - button_spacing - button_size + 1);
 						if (global_defined == _global_preprocessor_definitions.end() && preset_defined == _preset_preprocessor_definitions.end())
 							ImGui::LabelText("##value_undefined", "(undefined)");
 						else
--- a/source/gui.cpp
+++ b/source/gui.cpp
@@ -1794,101 +1794,101 @@ void reshade::runtime::draw_overlay_variable_editor()
 				{
 					ImGui::PopStyleVar();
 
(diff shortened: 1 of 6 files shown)


crosire:
I like this! Neat idea of using `#ifdef`/`#ifndef` to check for possible defines. My only concern is that yet another button above each effect is starting to make the UI a bit confusing. Maybe it makes sense to show them inline above the uniforms, not sure, will experiment a bit (my time is limited in the next few weeks though, so it might take me a while).

CeeJayDK:
I'm Commander Shepard, and this is my favourite patch on Github.

seri14:
smaa

source/runtime.cpp
```diff
@@ -247,10 +247,16 @@ void reshade::runtime::load_effect(const std::filesystem::path &path, size_t &ou
...
+		if (pp.append_file(path))
+		{
+			effect.user_definitions = std::move(pp.user_definitions());
+		}
+		else
 		{
 			LOG(ERROR) << "Failed to load " << path << ":\n" << pp.errors();
 			effect.compile_sucess = false;
+
+			effect.user_definitions.clear();
```
crosire: This is not necessary. `effect` is a new object, so `user_definitions` will always be empty here anyway. There is nothing to clear. Either way, doesn't matter much, just do this for simplicity:
```c++
if (!pp.append_file(path))
{
...
}

effect.user_definitions = std::move(pp.user_definitions());
```

source/runtime.cpp
```diff
@@ -273,12 +279,15 @@ void reshade::runtime::load_effect(const std::filesystem::path &path, size_t &ou
 			_renderer_id & 0x10000 ? reshadefx::codegen::backend::glsl : reshadefx::codegen::backend::hlsl;
 
 		// Compile the pre-processed source code (try the compile even if the preprocessor step failed to get additional error information)
-		if (!parser.parse(std::move(source_code), language, shader_model, true, _performance_mode, effect.module))
+		if (parser.parse(std::move(source_code), language, shader_model, true, _performance_mode, effect.module))
+		{
+			parser.errors();
```
crosire: This doesn't make sense? =)

source/gui.cpp
```diff
@@ -1786,6 +1786,139 @@ void reshade::runtime::draw_overlay_variable_editor()
...
 			{
+				const float button_size = ImGui::GetFrameHeight();
+				const float button_spacing = _imgui_context->Style.ItemInnerSpacing.x;
+
+				ImGui::PushStyleVar(ImGuiStyleVar_FramePadding, ImVec2(_imgui_context->Style.FramePadding.x, 0));
+				if (imgui_popup_button("Edit effect preprocessor definitions", -1, ImGuiWindowFlags_NoMove))
+				{
+					ImGui::PopStyleVar();
+
+					std::vector<std::string> user_definitions;
```
crosire: This should be `std::set` to avoid duplicates (`set` instead of `unordered_set` to guarantee ordering in the UI).

seri14:
This is draft. (I'm stopped work for proposal)
example, I guess it is required to support scrolling of definitions. (if use this)
and I do not know if it is good to filter definitions.
My idea is only exclude ReShade*.fxh

seri14:
and behaviour when user mistaken set setting (filename is not have quote, etc)

seri14:
back to self repository

PR #112, opened 2019-04-12
Fixed miscleanup of the "list" UI

seri14:
I broke the code at #107

### DIFF
--- a/source/gui_widgets.cpp
+++ b/source/gui_widgets.cpp
@@ -373,7 +373,7 @@ bool imgui_list_with_buttons(const char *label, const std::string_view ui_items,
 	bool modified = false;
 
 	std::vector<std::string_view> items;
-	for (std::string_view item = ui_items; ui_items.data() + ui_items.size() > item.data();
+	for (std::string_view item = ui_items.data(); ui_items.data() + ui_items.size() > item.data();
 		item = item.data() + item.size() + 1)
 		items.push_back(item);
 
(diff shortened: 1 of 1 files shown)


source/gui_widgets.cpp
```diff
@@ -373,9 +373,10 @@ bool imgui_list_with_buttons(const char *label, const std::string_view ui_items,
 	bool modified = false;
 
 	std::vector<std::string_view> items;
-	for (std::string_view item = ui_items; ui_items.data() + ui_items.size() > item.data();
-		item = item.data() + item.size() + 1)
-		items.push_back(item);
+	if (!ui_items.empty())
+		for (std::string_view item = ui_items.data(); ui_items.data() + ui_items.size() > item.data();
```
crosire: I'm confused. How is `std::string_view item = ui_items` different from `std::string_view item = ui_items.data()` considering that `ui_items` already points to the entire string?

seri14: the string_view have string_view(const string_view &) string_view(const char *) and string_view(const char *, size_t).
there have different behavior each other.

the behavior of string_view(const char *)
string_view.data = const char *
string_view.size = strlen(const char *)

the behavior of string_view(const char *, size_t)
string_view.data = const char *
string_view.size = size_t

the behavior of string_view(const string_view &)
string_view.data = const string_view &.data
string_view.size = const string_view &.size

in this case,
the std::vector<std::string_view> items require refer to strings splitted by '\0'.
in the before code, the std::vector<std::string_view> items have only one item so it is no splitted string.
in the after code,
the item initialized by first of splitted strings.
when the update of item, set the next index of the '\0'. (assume begin of next string)

seri14: so What I did at #107 was moved the strlen code to string_view`s constructor.
https://github.com/seri14/reshade/blob/515287d20ce615c19cf3d4c21b49f83896f04ddc/source/gui_widgets.cpp#L380

PR #113, opened 2019-04-24
Fix loading of effect headers with an UTF-8 BOM

seri14:


crosire:
What does this fix? You replaced C++ fstream with the C file API, but I don't see the benefit yet (yes, the current version may be a bit slower because it reads the file through an iterator, but does it really matter that much?).

seri14:
see line of 553-555

seri14:
I do not see it will faster by applying this. (only 150 differences with simple calculate)
This request includes change the how to read of the effect header file, and it has purpose of omit unnecessary steps.
These code is no needed to some process for each characters here. (system locale, parsing etc) these are just create the std::string from a file.
In the after code, it is only calculate file length, read file and exclude BOM.
[FIO19-C. Do not use fseek() and ftell() to compute the size of a regular file](https://wiki.sei.cmu.edu/confluence/display/c/FIO19-C.+Do+not+use+fseek%28%29+and+ftell%28%29+to+compute+the+size+of+a+regular+file)

crosire:
Pushed a commit which only contains the BOM fix until I decide what to do with this. I'm conflicted because I don't like using POSIX in the code base. I'd much rather only rely on C++ STL features.

seri14:
I did not modify if it was function as a stream enough. (like ini_file class)
but I understand.

PR #114, opened 2019-04-24
Select the preset from directory iterator

seri14:
This request changes the behavior of the preset list.
* GENERAL::PresetFiles will not be used.
* Change the value of GENERAL::CurrentPreset from index of GENERAL::PresetFiles to path of current preset.


seri14:
presetexplorer

seri14:
name_window

crosire:
I like it! But for better backwards compatibility I'd just rename the INI entry to `CurrentPresetPath`. That way `CurrentPreset` can just be ignored and there is no need to check whether it is a path or an index.
Also, this no longer uses the preset search paths, so that option can be removed from both the UI and INI too.

seri14:
About "selection popup should close itself after you click on a file".
In the demo, I selected the ini file but it is not a preset. (ReShade.ini is not have TechniqueSorting)
If select correct preset file, ReShade is open it.
Reason is avoid users destroy game files.

seri14:
Or cancel selection?
Maybe we lill break the preset by unintended, because we believe it is new preset.
(more occurs often when the filename is covered.)

crosire:
Ah, I clicked on the wrong INI during testing. Makes sense now.

PR #116, opened 2019-04-28
Add prev/next buttons for preset explorer

seri14:
This proposal is reimplement of #100

This request contains
* Add prev/next button
* Change contant of preset path to stem
* Highlight current preset in selectable items ~and can not select it~

preset_explorer_prevnext

### DIFF
--- a/source/gui.cpp
+++ b/source/gui.cpp
@@ -2179,7 +2179,13 @@ void reshade::runtime::draw_preset_explorer()
 			}
 
 			if (not_found)
-				condition = condition::pass;
+				if (preset_paths.size() > 0)
+					if (condition == condition::backward)
+						_preset_working_path = preset_paths[preset_paths.size() - 1];
+					else
+						_preset_working_path = preset_paths[0];
+				else
+					condition = condition::pass;
 			else
 				_current_preset_path = std::filesystem::absolute(_preset_working_path);
 		}
--- a/source/gui.cpp
+++ b/source/gui.cpp
@@ -2025,45 +2025,45 @@ void reshade::runtime::draw_preset_explorer()
 	condition condition = condition::pass;
 
 	if (ImGui::ButtonEx("<", ImVec2(button_size, 0), ImGuiButtonFlags_NoNavFocus))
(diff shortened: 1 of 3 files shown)


source/gui.cpp
```diff
@@ -2093,40 +2136,39 @@ void reshade::runtime::draw_preset_explorer()
...
 		}
+		for (const auto &entry : preset_files)
+		{
+			const bool is_current_preset_path = entry == _current_preset_path;
+			if (bool selected = is_current_preset_path;  ImGui::Selectable(entry.path().filename().u8string().c_str(), &selected))
+				_file_selection_path = entry, condition = is_current_preset_path ? condition::none : condition::select;
+
+			if (((condition == condition::backward || condition == condition::forward) && entry == _file_selection_path)
+				|| (is_current_preset_path && ImGui::IsWindowAppearing()))
+				ImGui::SetScrollHereY(); // TODO: not work in is_current_preset_path && ImGui::IsWindowAppearing()
```
seri14: If use these functions, It requires a window equals to root window.

seri14:
Added changes
* Path area's default state is button. If it press then closing and cancel selection. (To edit path, press button with control key)
* Scroll to current preset after it switching.
* Fixed control focus issues. 
  * after open the preset selector, you can scroll the explorer.
  * after press plus button, you can input name instantly
* Fixed behavior when current preset is not found or selection source is empty.

preset_explorer_prevnext2

seri14:
Demo of use the prev/next buttons
preset_explorer_prevnext_demo

source/runtime.cpp
```diff
@@ -875,7 +885,7 @@ void reshade::runtime::save_config(const std::filesystem::path &path) const
 
 void reshade::runtime::load_preset(const std::filesystem::path &path)
 {
-	const ini_file preset(path);
+	const ini_file preset(g_reshade_dll_path.parent_path() / path);
```
crosire: I'd rather force `_current_preset_path` to always be absolute rather than doing this here.

seri14:
I finished implementation of this feature.
ready for review.

PR #117, opened 2019-04-30
Change the default value of NoFontScaling to 1

seri14:
because I met many users that happened troubles by this.
and output key/value to support users who need it.

### DIFF
--- a/source/gui.cpp
+++ b/source/gui.cpp
@@ -213,6 +213,7 @@ void reshade::runtime::init_ui()
 		config.set("GENERAL", "ShowFrameTime", _show_frametime);
 		config.set("GENERAL", "ShowScreenshotMessage", _show_screenshot_message);
 		config.set("GENERAL", "ClockFormat", _clock_format);
+		config.set("GENERAL", "NoFontScaling", _no_font_scaling);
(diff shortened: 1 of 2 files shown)


crosire:
I'm confused as to how this can cause trouble. It only scales as long as the UI is open. And it's explicitly mentioned in the tutorial on first startup, so disabling would be contradicting that.

seri14:
I'm confused too.
If all users read it or fully remember it by not required users, I have not wrote this. who press the [Restart Tutorial] button when scaled?

Mortalitas:
I'll second what Seri is pointing out here, many end users, both those who have completed the tutorial and those who have not (by means of alternative distributions or otherwise), find the feature to be more of a nuisance than a benefit.

Having the scale "locked" (in this case, hotkeys off) by default benefits the majority who will never actually use the feature but may be bewildered when accidentally triggering it.

seri14:
Well, about alternative distributions, It have no problems there. but not touch the default value of TutorialProgress.
and distributor should can completely avoid this problem.

crosire:
Well at least have it enabled by default, and disable it once the tutorial was finished. Then it's no longer a nuisance for those users that cannot read, but is still there to scale the windows during the tutorial.

PR #118, opened 2019-05-08
Fix slow sorting of technique list

seri14:
and change the sorting to ignore case

### DIFF
--- a/source/gui.cpp
+++ b/source/gui.cpp
@@ -701,11 +701,10 @@ void reshade::runtime::draw_overlay_menu_home()
 			}
 			if (const auto it = std::find_if_not(_techniques.begin(), _techniques.end(), [](const reshade::technique &a) { return a.enabled || a.toggle_key_data[0] != 0; }); it != _techniques.end())
 			{
-				static const std::string ui_label("ui_label");
 				std::stable_sort(it, _techniques.end(), [](const reshade::technique &h, const reshade::technique &i)
 					{
-						const auto v = h.annotations.find(ui_label);
-						const auto w = i.annotations.find(ui_label);
+						const auto v = h.annotations.find("ui_label");
+						const auto w = i.annotations.find("ui_label");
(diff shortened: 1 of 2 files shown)


source/gui.cpp
```diff
@@ -699,29 +699,19 @@ void reshade::runtime::draw_overlay_menu_home()
...
-								std::iter_swap(i, k);
-								i = _techniques.begin();
-								break;
-							}
-						}
-					}
-				}
+						const auto v = h.annotations.find(ui_label);
+						const auto w = i.annotations.find(ui_label);
+						std::string x(v == h.annotations.end() ? h.name : v->second.second.string_data);
```
crosire: I'd just do `std::string x = h.annotation_as_string("ui_label");` here for less code.

The static `ui_label` allocation should not be necessary because of short string optimization (SSO) in `std::string`: https://stackoverflow.com/questions/10315041/meaning-of-acronym-sso-in-the-context-of-stdstring/10319672#10319672

source/gui.cpp
```diff
@@ -699,29 +699,19 @@ void reshade::runtime::draw_overlay_menu_home()
 					}
 				}
 			}
-			for (auto i = _techniques.begin(); i != _techniques.end(); ++i)
+			if (const auto it = std::find_if_not(_techniques.begin(), _techniques.end(), [](const reshade::technique &a) { return a.enabled || a.toggle_key_data[0] != 0; }); it != _techniques.end())
```
crosire: This seems unnecessary. Just always do the sort without this `if`. That would also pay the price for sorting when no technique is enabled, but there is little reason for the user to click on "Active to top" then anyway, so it doesn't matter.

seri14: Sorry I have no idea how to the 2pass sorting to 1pass.

source/gui.cpp
seri14: No. Here code require using the Label, but to using the Filename if that nothing.

PR #119, opened 2019-05-12
Fix delaying of depth detection

seri14:
This change is make certain that "save_depth_texture()" function does copy the last used depth texture before "Present()" call.
And add support that if game does clearing depth stencil view at inverted timing.


seri14:
Have problem?

seri14:
Waiting @Thalixte 's review?

seri14:
I found to how to avoid this problem
https://mobile.twitter.com/seri_haruna/status/1134607556565848065

PR #123, opened 2019-05-19
add a fix for weapon display in the depth buffer, and a feature to only focus on the original depthstencil source

Thalixte:
With this PR, i think i am done with DX9 depth buffer detection.

- The first feature, named "Fix for source engine games" is dedicated to the games using the source engine (half-Life2, Portal, left4Dead 2): when autopreserve mode is enabled, the detection fails sometimes because the depthbuffer of the background scene have more vertices than the depth buffer of the main scene. This fix corrects it

- The second feature, named "Fix for user weapon or cockpit" help fix user weapon or cockpit display in the depth buffer in some games (UT3, for instance)

- The third feature, named "Focus on the best original depthstencil source" help remove undesired UI elements from the depth buffer in some games (Race Driver GRID or Oblivion, for instance)

Update: @Crosire, you were right for the viewport test: it is better to put it on the on_draw_call method rather than on the on_set_viewport one.


nomad5oul:
@Thalixte @crosire This fix appears to have been removed after Reshade v4.4.2. I can no longer find the option 'Fix for user weapon or cockpit' under the DX9 depth buffer menu. Please can this be added back in as I need it for some older DX9 games?

PR #127, opened 2019-06-01
Add support for saving  before and after screenshots on the same frame

healingbrew:
This patch allows ReShade to save a clean image before rendering effects by moving the `save_screenshot` call to the `update_and_render_effects` subroutine when the option is selected. The benefit of this is having before-and-after screenshots on the same frame, allowing you to see the effects of the shaders with no changes from the game.

This modifies the signature for `save_screenshot()` into `save_screenshot(bool)` to accomodate for the `-before` and `-after` suffix on the screenshot filename.

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -157,6 +157,13 @@ void reshade::runtime::on_present()
 	// Lock input so it cannot be modified by other threads while we are reading it here
 	const auto input_lock = _input->lock();
 
+	if (_should_save_screenshot)
+	{
+		if (_effects_enabled)
+			save_screenshot(false);
+		_should_save_screenshot = false;
+	}
+
 	// Handle keyboard shortcuts
 	if (!_ignore_shortcuts)
 	{
@@ -167,7 +174,12 @@ void reshade::runtime::on_present()
 			_effects_enabled = !_effects_enabled;
 
 		if (_input->is_key_pressed(_screenshot_key_data))
-			save_screenshot();
+		{
+			if (_screenshot_save_before)
+				_should_save_screenshot = true;
+			else
+				save_screenshot(false);
+		}
 	}
 
 #if RESHADE_GUI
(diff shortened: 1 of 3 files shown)


source/runtime.cpp
```diff
@@ -1022,7 +1039,7 @@ void reshade::runtime::save_current_preset() const
 	save_preset(_current_preset_path);
 }
 
-void reshade::runtime::save_screenshot()
+void reshade::runtime::save_screenshot(bool before)
```
crosire: I'd change this to accept a postfix string which is appened to the path name instead of a boolean:
```c++
void reshade::runtime::save_screenshot(const std::string &postfix)
{
   ...
   const std::wstring screenshot_path = least + postfix + (_screenshot_format == 0 ? L".bmp" : L".png");
   ...
}
```

source/runtime.cpp
```diff
@@ -167,7 +174,12 @@ void reshade::runtime::on_present()
 			_effects_enabled = !_effects_enabled;
 
 		if (_input->is_key_pressed(_screenshot_key_data))
-			save_screenshot();
+		{
+			if (_screenshot_save_before)
+				_should_save_screenshot = true;
```
crosire: This will cause the screenshots to mismatch because they are created one frame apart. Better to move the whole logic into `update_and_render_effects` and avoid having a member field altogether.
```c++
void reshade::runtime::update_and_render_effects()
{
    ...
    const auto input_lock = _input->lock();

    const auto do_screenshot = _input->is_key_pressed(_screenshot_key_data));
    if (do_screenshot && (_screenshot_save_before || !_effects_enabled))
        save_screenshot(_effects_enabled ? "before" : std::string());

    if (!_effects_enabled)
        return;
    ...
    if (do_screenshot)
        save_screenshot(_screenshot_save_before ? "after" : std::string());
}
```

seri14:
I suggest change it to `-original`.

healingbrew:
Requested changes are implemented. The "after" image has no suffix, "before" image has "-original" only when before-after is enabled.

PR #129, opened 2019-06-12
Fix dx10 dx11 depth buffer detection

Thalixte:
@crosire 
This is a fix for some games where depth buffer detection fails since Reshade v3.2.2.
For instance, in Crysis DX10, if we use the last Reshade release (v4.3.0), the depth buffer detection fails. It is the same for Stalker Clear Sky.
I have not found yet a dx11 game where this pb occures, but i prefer to apply the same fix for DX11.
Here are some videos illustrating the pb:

- Crysis DX10 + Reshade v3.2.0 (working depth buffer):
https://youtu.be/gteIJFONvzA

- Crysis DX10 + Reshade v4.3.0 (depth buffer not working):
https://youtu.be/RtHSMa-biF8

- Crysis DX10 + Reshade v4.3.0 + fix (working depth buffer):
https://youtu.be/OXUR_PsJ-zI

The pb seems to happen when the BindFlags of the texture created by the create_depthstencil_replacement method is changed.


source/d3d10/runtime_d3d10.cpp
```diff
@@ -1503,7 +1542,9 @@ bool reshade::d3d10::runtime_d3d10::create_depthstencil_replacement(ID3D10DepthS
 
 		HRESULT hr = S_OK;
 
-		if ((tex_desc.BindFlags & D3D10_BIND_SHADER_RESOURCE) == 0)
+		_depthstencil_texture_desc_changed = ((tex_desc.BindFlags & D3D10_BIND_SHADER_RESOURCE) == 0);
+
+		if (_depthstencil_texture_desc_changed)
 		{
 			_depthstencil_texture.reset();
```
Thalixte: In the actual release, the pb occures when entering this portion of code

source/d3d10/runtime_d3d10.cpp
```diff
@@ -313,6 +313,10 @@ void reshade::d3d10::runtime_d3d10::on_present(draw_call_tracker &tracker)
 	if (_backbuffer_resolved != _backbuffer)
 		_device->ResolveSubresource(_backbuffer_resolved.get(), 0, _backbuffer.get(), 0, _backbuffer_format);
 
+	// Setup real back buffer
+	auto rtv = _backbuffer_rtv[0].get();
+	 _device->OMSetRenderTargets(1, &rtv, nullptr);
+
```
Thalixte: I don't know why, but those lines seems to be necessary for the fix to work

seri14:
I join to this conversation because a title on my library make no working.

crosire:
Mmh. This hasn't to do with the bind flags. It's just the fact that ReShade creates a replacement texture when the one from the game does not support shader binding.
And ReShade used to replace the game one with that, but that was accidentally removed in commit 6e2561aef30a86da94c0741fe9a53cb2e5ba637c.
For now your changes fix that, but it looks like the whole systems needs a refactor at some point, because things that belong together are now heavily split between `draw_call_tracker.cpp`, `d3d11_device_context.cpp` and `runtime_d3d11.cpp`.
Anyway, one nitpick: Instead of adding a new `_depthstencil_texture_desc_changed` field you can simply check if `_depthstencil_replacement != nullptr`, which will have the same effect. Basically revert what was removed here: https://github.com/crosire/reshade/commit/6e2561aef30a86da94c0741fe9a53cb2e5ba637c#diff-84fd0186510c97938d62c38a8735209bL663

source/d3d10/d3d10_device.cpp
```diff
@@ -48,11 +48,13 @@ bool D3D10Device::save_depth_texture(ID3D10DepthStencilView *pDepthStencilView,
 	texture->GetDesc(&desc);
 
 	// Check if aspect ratio is similar to the back buffer one
-	const float screen_aspect_ratio = float(runtime->frame_width()) / float(runtime->frame_height());
+	const float width_factor = desc.Width != runtime->frame_width() ? float(runtime->frame_width()) / desc.Width : 1.0f;
```
crosire: Why this? If `desc.Width == frame_width`, then result is `1.0f`, so the branch makes no sense?

source/d3d10/d3d10_device.cpp
```diff
@@ -246,6 +248,14 @@ void    STDMETHODCALLTYPE D3D10Device::OMSetRenderTargets(UINT NumViews, ID3D10R
 #if RESHADE_DX10_CAPTURE_DEPTH_BUFFERS
 	track_active_rendertargets(NumViews, ppRenderTargetViews, pDepthStencilView);
 #endif
+
+	if (_runtimes.empty())
+		return;
```
crosire: Shouldn't return here, so the original `OMSetRenderTargets` is called even if no runtime is associated with this device (this can happen depending on the settings the device was created with).

This is better:
```c++
if (!_runtime.empty())
{
    const auto runtime = _runtimes.front();
    runtime->on_set_depthstencil_view(pDepthStencilView);
}
```

source/d3d10/d3d10_device.cpp
```diff
@@ -295,8 +305,15 @@ void    STDMETHODCALLTYPE D3D10Device::ClearRenderTargetView(ID3D10RenderTargetV
 }
 void    STDMETHODCALLTYPE D3D10Device::ClearDepthStencilView(ID3D10DepthStencilView *pDepthStencilView, UINT ClearFlags, FLOAT Depth, UINT8 Stencil)
 {
+	if (_runtimes.empty())
+		return;
```
crosire: See above.

source/d3d10/runtime_d3d10.cpp
```diff
@@ -1487,9 +1532,12 @@ void reshade::d3d10::runtime_d3d10::detect_depth_source(draw_call_tracker &track
 
 bool reshade::d3d10::runtime_d3d10::create_depthstencil_replacement(ID3D10DepthStencilView *depthstencil, ID3D10Texture2D *texture)
 {
+	if (!depth_buffer_before_clear && depthstencil == _depthstencil)
+		return false;
+
 	_depthstencil.reset();
 	_depthstencil_replacement.reset();
-	_depthstencil_texture.reset();
+	_depthstencil_texture.reset();
```
crosire: =)

source/d3d11/d3d11_device_context.cpp
```diff
@@ -335,6 +337,14 @@ void    STDMETHODCALLTYPE D3D11DeviceContext::OMSetRenderTargets(UINT NumViews,
 #if RESHADE_DX11_CAPTURE_DEPTH_BUFFERS
 	track_active_rendertargets(NumViews, ppRenderTargetViews, pDepthStencilView);
 #endif
+
+	if (_device->_runtimes.empty())
+		return;
```
crosire: Same as in DX10.

source/d3d11/d3d11_device_context.cpp
```diff
@@ -421,8 +431,15 @@ void    STDMETHODCALLTYPE D3D11DeviceContext::ClearUnorderedAccessViewFloat(ID3D
...
+
+	const auto runtime = _device->_runtimes.front();
+
+	runtime->on_clear_depthstencil_view(pDepthStencilView);
+
 #if RESHADE_DX11_CAPTURE_DEPTH_BUFFERS
-	if (ClearFlags & D3D11_CLEAR_DEPTH)
+	if (ClearFlags & D3D11_CLEAR_DEPTH || (runtime->depth_buffer_more_copies && ClearFlags & D3D11_CLEAR_STENCIL))
```
crosire: I see little point in tracking stencil clears since we are only interested in depth?

Thalixte: Me too, but it makes Bioshock Infinite depth buffer detection work... As it is in very few games, i prefer to let Reshade filter by default on D3D11_CLEAR_DEPTH, then, if necessary, filter also on D3D11_CLEAR_STENCIL...

source/d3d10/d3d10_device.cpp
Thalixte: Yes, you totally right. I will change this

source/d3d10/d3d10_device.cpp
Thalixte: Yes, it's right. Fixed

source/d3d10/d3d10_device.cpp
Thalixte: Changed

source/d3d10/runtime_d3d10.cpp
Thalixte: =)

source/d3d11/d3d11_device_context.cpp
Thalixte: Changed

PR #130, opened 2019-06-19
Uniform migration

seri14:
Read this topic
https://reshade.me/forum/suggestions/5495-uniform-migration

However, I do not know yet if this feature is useful for anyone who has updated the effect files with break changes.

### DIFF
--- a/source/effect_parser.cpp
+++ b/source/effect_parser.cpp
@@ -2566,7 +2566,7 @@ bool reshadefx::parser::parse_variable(type type, std::string name, bool global)
 				return false;
 		}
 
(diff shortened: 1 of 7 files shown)


crosire:
While this seems like a nice feature, I'm doubting the additional complexity it adds is worth it. Also note that the effect compiler is intentionally separated from the runtime so that it can be used standalone (you can copy all the `effect_*` files out of this repository and still use them without any dependencies and there are actually one or two non-public projects which have done that). So such a feature should best be implemented in the runtime and not the parser stage. But like I said I'm not sure spending time on that is worth the payoff, considering that if shader developers rename uniforms, they usually also change the algorithm around them, so that it's not just a rename, but the new variable is actually interpreted differently and would need to be adjusted.

PR #131, opened 2019-06-24
DX12 db detection

Thalixte:
Hi @Crosire,

I apologize first not to have answered to your comments on my last PR, but i worked on the DX12 depth buffer detection feature.
It is still a WIP version, but i submit it to you in order to have your opinion on it.
Here are two videos of the currently available feature:
- https://www.youtube.com/watch?v=mYJ0MNfmgoo
- https://www.youtube.com/watch?v=iCqKbuJlR5E

What works for now:
- depth buffer textures retrieval in standard mode
- depth buffer number of drawcalls and number of vertices stats
- in standard mode, the automatic depth buffer detection

What does not work yet:
- in "Copy depth before clearing" mode, the depth buffer copy does not work (i will investigate)
- in order for the depth buffer switching to work, i need to call the "load_effects()" method, so it recompiles the technique effects. I guess there should be a better way to switch without having to recompile

Tested successfully on Metro Exodus and Shadow of the Tomb Raider. I will propose a dll for testing purpose on the Reshade forum.

The code needs a lot of refactoring, i guess. For instance, i did not manage to use a wrapped D3D12CommandList class, because when Reshade call the D3D12CommandQueue->ExecuteCommandLists method, i get an access violation error, so i hooked each ID3D12CommandList methods separately.
 
For the moment, critical paths are protected by mutexes (both in d3d12_device.cpp and in draw_call_tracker.cpp), but it reduces performances of the game (1 fps in the SOTR benchmark, after i fixed some calls) because it breaks the multi threading benefits. I guess there is a better way to ensure the stability...


source/d3d12/runtime_d3d12.hpp
```diff
@@ -5,11 +5,15 @@
 
 #pragma once
 
+#include "ini_file.hpp"
```
crosire: This is only needed in `runtime_d3d12.cpp`, so should be moved there.

source/d3d11/runtime_d3d11.cpp
```diff
@@ -1513,31 +1513,6 @@ bool reshade::d3d11::runtime_d3d11::create_depthstencil_replacement(ID3D11DepthS
...
-				hr = _device->CreateDepthStencilView(_depthstencil_texture.get(), &dsv_desc, &_depthstencil_replacement);
-			}
-		}
-
-		if (FAILED(hr))
-		{
-			LOG(ERROR) << "Failed to create depth stencil replacement texture! HRESULT is '" << std::hex << hr << std::dec << "'.";
-			return false;
-		}
-
```
crosire: This looks like it was accidentally deleted.

source/d3d12/d3d12.cpp
```diff
@@ -6,6 +6,9 @@
 #include "log.hpp"
 #include "hook_manager.hpp"
 #include "d3d12_device.hpp"
+#include "runtime_d3d12.hpp"
+
+#undef ID3D12GraphicsCommandList_ClearDepthStencilView
```
crosire: The changes in this file are not used (probably leftovers from earlier).

source/d3d12/d3d12_device.cpp
```diff
@@ -325,7 +581,6 @@ HRESULT STDMETHODCALLTYPE D3D12Device::EnqueueMakeResident(D3D12_RESIDENCY_FLAGS
 
 HRESULT STDMETHODCALLTYPE D3D12Device::CreateCommandList1(UINT NodeMask, D3D12_COMMAND_LIST_TYPE Type, D3D12_COMMAND_LIST_FLAGS Flags, REFIID riid, void **ppCommandList)
 {
-	assert(_interface_version >= 4);
```
crosire: Whoopsie.

source/d3d12/d3d12_device.cpp
```diff
@@ -158,15 +379,47 @@ HRESULT STDMETHODCALLTYPE D3D12Device::CreateComputePipelineState(const D3D12_CO
...
+
+	if (riid == __uuidof(ID3D12GraphicsCommandList) || riid == __uuidof(ID3D12GraphicsCommandList1))
+	{
+		if (_runtimes.empty())
+			return hr;
+
+		// retrieve current runtime
+		const auto runtime = _runtimes.front();
+		d3d12_current_runtime = runtime.get();
+		d3d12_current_device = this;
```
crosire: This is rather dangerous, since the app can technically create a commandlist on a different device before recording into the one just created here, which will set `d3d12_current_device` to something wrong. Would be safer to have a global lookup table (`std::unordered_map<ID3D12CommandList *, D3D12Device *>`) to map from commandlist to the matching device and runtime.

ReShade.vcxproj
```diff
@@ -397,6 +399,7 @@
     <ClCompile Include="source\windows\ws2_32.cpp" />
   </ItemGroup>
   <ItemGroup>
+    <ClInclude Include="res\d3dx12.h" />
```
crosire: That shouldn't be here =)

source/d3d12/runtime_d3d12.cpp
```diff
@@ -1400,4 +1453,293 @@ void reshade::d3d12::runtime_d3d12::render_imgui_draw_data(ImDrawData *draw_data
...
+	if (const auto it = _depth_texture_saves.find(idx); it != _depth_texture_saves.end())
+		return it->second;
+
+	if(!(textureDesc.Flags & D3D12_RESOURCE_FLAG_ALLOW_DEPTH_STENCIL))
+		textureDesc.Flags |= D3D12_RESOURCE_FLAG_ALLOW_DEPTH_STENCIL;
+
+	// Create the saved texture pointed by the index if it does not already exist
+	com_ptr<ID3D12Resource> depth_texture_save;
+
+	// Render targets are always either cleared to zero or not cleared at all (see 'ClearRenderTargets' pass state), so we can set the optimized clear value here to zero
```
crosire: This comment doesn't make sense in this context =P

source/d3d12/d3d12.cpp
Thalixte: Yes, it can be removed =)

ReShade.vcxproj
Thalixte: Did not know where to put it (remember, it is a WIP version ;-) )

source/d3d11/runtime_d3d11.cpp
Thalixte: Yes, it was in order to focus on the part of the db detection that works for the majority of the games. It should be reintroduced

source/d3d12/d3d12_device.cpp
Thalixte: Yes... it should be reintroduced

ReShade.vcxproj
seri14: In the next, I think good to create a pull request with draft.

source/d3d12/runtime_d3d12.cpp
```diff
@@ -1118,6 +1170,7 @@ void reshade::d3d12::runtime_d3d12::render_technique(technique &technique)
 		// Setup states
 		cmd_list->SetPipelineState(pass_data.pipeline.get());
 		cmd_list->OMSetStencilRef(pass_info.stencil_reference_value);
+		cmd_list->GetType();
```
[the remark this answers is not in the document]
Thalixte: Yes =) Can be removed

source/d3d12/d3d12.cpp
Thalixte: Yes. It can be removed

source/d3d12/runtime_d3d12.cpp
Thalixte: Yes. Copy/paste stuff. Can be removed

ReShade.vcxproj
Thalixte: I know, but it is a WIP version. Where do you want me to put it ?

Thalixte:
> Great work! Took an initial quick look at it and got some nitpicks below (mostly refactoring stuff, as you mentioned). Will look into it in more detail at the weekend.

Thks a lot. There are still some works to do ;-)

source/d3d12/d3d12_device.cpp
Thalixte: I understand. From what i have tested now, it seems not to raise pbs, but in the future, it probably will...

ReShade.vcxproj
Thalixte: Yes, i will think about it the next time. Do you know how to switch to draft version on github ?

crosire:
What's the reason for adding d3dx12.h? I'd rather not add additional dependencies.

Thalixte:
> What's the reason for adding d3dx12.h? I'd rather not add additional dependencies and instead do whatever you need from that manually.

Yes, that's right. I used a D3D12 helper that i removed, so the library vcan also be removed. It's done.

source/d3d12/d3d12_device.cpp
Thalixte: In Shadow Of the Tomb Raider, it is more than necessary

source/d3d11/runtime_d3d11.cpp
```diff
@@ -1548,13 +1548,6 @@ bool reshade::d3d11::runtime_d3d11::create_depthstencil_replacement(ID3D11DepthS
 			LOG(ERROR) << "Failed to create depth stencil replacement resource view! HRESULT is '" << std::hex << hr << std::dec << "'.";
 			return false;
 		}
-
-		// Update auto depth stencil
-		com_ptr<ID3D11DepthStencilView> current_depthstencil;
-		com_ptr<ID3D11RenderTargetView> targets[D3D11_SIMULTANEOUS_RENDER_TARGET_COUNT];
-		_immediate_context->OMGetRenderTargets(D3D11_SIMULTANEOUS_RENDER_TARGET_COUNT, reinterpret_cast<ID3D11RenderTargetView **>(targets), &current_depthstencil);
-		if (current_depthstencil != nullptr && current_depthstencil == _depthstencil)
-			_immediate_context->OMSetRenderTargets(D3D11_SIMULTANEOUS_RENDER_TARGET_COUNT, reinterpret_cast<ID3D11RenderTargetView *const *>(targets), _depthstencil_replacement.get());
```
crosire: This shouldn't be deleted.

source/d3d12/runtime_d3d12.cpp
```diff
@@ -337,6 +375,32 @@ void reshade::d3d12::runtime_d3d12::on_present()
...
+	HANDLE fenceEvent = nullptr;
+	com_ptr<ID3D12Fence> fence;
+
+	D3D12_COMMAND_QUEUE_DESC descCopyQueue = {};
+	descCopyQueue.Flags = D3D12_COMMAND_QUEUE_FLAG_NONE;
+	descCopyQueue.Type = D3D12_COMMAND_LIST_TYPE_DIRECT;
+	_device->CreateCommandQueue(&descCopyQueue, IID_PPV_ARGS(&copyQueue));
+	_device->CreateCommandAllocator(D3D12_COMMAND_LIST_TYPE_DIRECT, IID_PPV_ARGS(&copyAllocator));
+
+	if (FAILED(_device->CreateCommandList(0, D3D12_COMMAND_LIST_TYPE_COPY, copyAllocator.get(), nullptr, IID_PPV_ARGS(&copyList))))
```
crosire: Don't create a new command queue and allocator here. The `runtime_d3d12` class already has the helper functions `create_command_list` and `execute_command_list` to do the work for you.

```c++
const com_ptr<ID3D12GraphicsCommandList> cmd_list = create_command_list();
transition_state(cmd_list, dest, D3D12_RESOURCE_STATE_COMMON, D3D12_RESOURCE_STATE_COPY_DEST);
cmd_list->CopyResource(dest.get(), src.get());
transition_state(cmd_list, dest, D3D12_RESOURCE_STATE_COPY_DEST, D3D12_RESOURCE_STATE_COMMON);
execute_command_list(cmd_list);
```

source/d3d12/runtime_d3d12.cpp
```diff
@@ -501,7 +565,8 @@ bool reshade::d3d12::runtime_d3d12::init_texture(texture &info)
 	texture_data->resource->SetName(debug_name.c_str());
 #endif
 
-	{   D3D12_DESCRIPTOR_HEAP_DESC heap_desc = { D3D12_DESCRIPTOR_HEAP_TYPE_CBV_SRV_UAV };
+	{
+		D3D12_DESCRIPTOR_HEAP_DESC heap_desc = { D3D12_DESCRIPTOR_HEAP_TYPE_CBV_SRV_UAV };
```
crosire: =)

source/d3d12/runtime_d3d12.hpp
```diff
@@ -22,10 +26,23 @@ namespace reshade::d3d12
...
+#if RESHADE_DX12_CAPTURE_DEPTH_BUFFERS
+		com_ptr<ID3D12Resource> select_depth_texture_save(D3D12_RESOURCE_DESC &texture_desc, const D3D12_HEAP_PROPERTIES *props, D3D12_HEAP_FLAGS *heap_flags);
+#endif
+
+		bool depth_buffer_before_clear = false;
+		bool depth_buffer_more_copies = false;
+		bool extended_depth_buffer_detection = false;
+		unsigned int cleared_depth_buffer_index = 0;
+		int depth_buffer_texture_format = 0; // No depth buffer texture format filter by default
+		unsigned int clear_DSV_iter = 1;
```
crosire: This should be an atomic. `std::atomic<unsigned int> clear_DSV_iter`

source/d3d12/d3d12_device.hpp
```diff
@@ -95,9 +98,24 @@ struct __declspec(uuid("2523AFF4-978B-4939-BA16-8EE876A4CB2A")) D3D12Device : ID
...
+#endif
+	
 	LONG _ref = 1;
 	ID3D12Device *_orig;
 	unsigned int _interface_version;
+	std::vector<std::shared_ptr<reshade::d3d12::runtime_d3d12>> _runtimes;
+	std::unordered_map<ID3D12CommandList *, reshade::d3d12::draw_call_tracker> _trackers_per_commandlist;
+	std::mutex _trackers_per_commandlist_mutex;
+	reshade::d3d12::draw_call_tracker _draw_call_tracker;
+	std::mutex _clear_DSV_iter_mutex;
```
crosire: This is not necessary if `clear_DSV_iter` is an atomic.

source/d3d12/runtime_d3d12.cpp
```diff
@@ -1093,7 +1145,7 @@ void reshade::d3d12::runtime_d3d12::render_technique(technique &technique)
 	for (size_t i = 0; i < technique.passes.size(); ++i)
 	{
 		const auto &pass_info = technique.passes[i];
-		const auto &pass_data = *technique.passes_data[i]->as<d3d12_pass_data>();
+		const auto &pass_data = *technique.passes_data[i]->as<d3d12_pass_data>();
```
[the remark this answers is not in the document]
Thalixte: =)

source/d3d11/runtime_d3d11.cpp
Thalixte: fixed

source/d3d12/runtime_d3d12.cpp
Thalixte: Unfortunately, it does not fix random CTDs (Devil May Cry 5, SOTR). Maybe there is a need to add some fences

source/d3d12/runtime_d3d12.cpp
Thalixte: =)

source/d3d12/runtime_d3d12.hpp
Thalixte: OK. Changed

source/d3d12/d3d12_device.hpp
Thalixte: Changed

source/d3d12/draw_call_tracker.cpp
```diff
@@ -0,0 +1,256 @@
...
+	void draw_call_tracker::track_rendertargets(int format_index, ID3D12Resource *depthstencil, D3D12_CPU_DESCRIPTOR_HANDLE pDepthStencilView)
+	{
+		assert(depthstencil != nullptr);
+
+		if (!check_depth_texture_format(format_index, depthstencil))
+			return;
+
+		std::lock_guard lock(_counters_per_used_depthstencil_mutex);
+
+		if (&_counters_per_used_depthstencil[depthstencil] != nullptr)
```
crosire: `&_counters_per_used_depthstencil[depthstencil]` can never be `nullptr`, since that is a reference to a struct. So this doesn't make sense?

Thalixte: OK Reverted

Thalixte: @Crisire, can you help me find a way to switch between the depth buffers withou having to reload the effects ?

crosire: That's a bit tricky ...

~Is suppose the correct way would mean adding support for dynamic descriptor allocation though. So instead of having a static descriptor heap as part of the effect, there would have to be a way to request an allocator for X number of descriptors, then copy the current effect descriptor table into that new table and then update the entry that references the depth buffer texture to the new texture with `CreateShaderResourceView`. After all that is done the current effect descriptor table would be updated to point to the new one.~

Maybe it works to just keep track where in the descriptor table the depth buffer texture is stored in each effect (aka keep track of the `D3D12_CPU_DESCRIPTOR_HANDLE` of it) and then overwrite that entry for all effects with the new texture via `CreateShaderResourceView`. Have to be careful with synchronization though (this shouldn't happen while there are still frames in flight that may access the descriptor table currently).

Thalixte: Thks for your answer. What's sound strange to me is that, if i remove the load_effects calls, the switch works correctly when in depth buffer copy mode, but is does not work when in standard detection mode...

Thalixte:
Not a WIP version anymore. I think it works good now...

PR #133, opened 2019-07-18
Changed to not show/hide the UI when editing text

seri14:
For who regularly use the Home key


crosire:
There is no ImGui context set at that point, so a call to `ImGui::IsAnyItemActive` is invalid and may crash (`ImGui::SetCurrentContext` is called a couple of lines below, and also in `build_font_atlas`, but not before). 

Simplest way to fix that is to just access the context object directly there:
```c++
... && _imgui_context->ActiveId == 0)
```

PR #134, opened 2019-07-20
Add a button to show disassembled functions

seri14:
d3ddisassemble


seri14:
generated

crosire:
I moved the "Show HLSL/GLSL" button to the technique context menu (line 2040 in gui.cpp), so this is conflicting right now. Would be awesome if you could resolve the conflict by moving your code there as well.

seri14:
This commit has change to remove "Act as a toggle when already showing the generated code"
refer to SammyEnigma#40

PR #138, opened 2019-08-02
Vulkan db detection

Thalixte:
This PR does the following things:
- add a vulkan tab to the UI in order to give user the same functionalities as in the other APIs
- retrieve the depthstencil images, the depthstencil image views, and their relative creating infos
- retrieve drawcall statistics for those depthstencil images
- allow to switch between the retrieved depthstencil images, now without having to reload the effects 
- depth buffer detection now works !!!

What's wrong with this PR:
- ~~some crashes arise when loading a new game in Rage 2 (does not occure in Doom)~~
- ~~only depthstencil images are visible, and not depth buffer ones~~
- ~~switching between depthstencil images requires to reload the effects~~
- ~~maybe it misses some drawcalls stats (for instance, those from the vkCmdDrawIndexedIndirect API method)~~

__UPDATE:__
Depth buffer seems now to be avalable for all vulkan games i tried 😄 
No Man's Sky Beyond:
No Man's Sky

Doom:
DOOMx64v_deptbufferk

Rage 2:
RAGE2_depth_buffer

Strange Brigade:
Strange Brigade_depth_buffer


Thalixte:
@crosire, do you see any reason which explain that only depthstencil images are not visible, not the depth buffer ones (even if i add a VK_IMAGE_USAGE_SAMPLED_BIT at the depthstencil image creation stage) ?
Do you have an efficient way to track the validation errors that occures when updating the descriptor sets after changing the depth buffer image for the shaders ? 
Also, can you  help me reading back the data from the VkBuffer in the vkCmdDrawIndexedIndirect API method ?

source/runtime.cpp
```diff
@@ -103,6 +103,10 @@ reshade::runtime::runtime() :
 }
 reshade::runtime::~runtime()
 {
+	// ensure to reset the runtime if this is called direcly by the app (using alt+F4 for instance)
+	if(!_worker_threads.empty() || !_techniques.empty())
+		on_reset();
```
crosire: I'd rather not have this here. It is not save to call `on_reset` here, because the API may have already been destroyed (e.g. OpenGL context was destroyed, so this could crash, because OpenGL function pointers would be invalid).

ReShade.vcxproj
```diff
@@ -368,7 +368,6 @@
     <ClCompile Include="source\d3d11\state_block.cpp" />
     <ClCompile Include="source\d3d11\draw_call_tracker.cpp" />
     <ClCompile Include="source\d3d12\d3d12.cpp" />
-    <ClCompile Include="source\d3d12\d3d12_command_list.cpp" />
```
crosire: Please don't remove `d3d12_command_list.hpp` and `d3d12_command_list.cpp` yet. They may become useful in the future, since I plan on moving the function hooks into those (after fixing the crash you were seeing witht that). It's also not really related to Vulkan.

source/vulkan/runtime_vk.cpp
```diff
@@ -160,7 +179,7 @@ VkImageView reshade::vulkan::runtime_vk::create_image_view(VkImage image, VkForm
 {
 	VkImageAspectFlags aspect = VK_IMAGE_ASPECT_COLOR_BIT;
 	if (format >= VK_FORMAT_D16_UNORM_S8_UINT && format <= VK_FORMAT_D32_SFLOAT_S8_UINT)
-		aspect = VK_IMAGE_ASPECT_STENCIL_BIT;
+		aspect = VK_IMAGE_ASPECT_STENCIL_BIT | VK_IMAGE_ASPECT_DEPTH_BIT;
```
crosire: This is missing some formats, because the `format >= VK_FORMAT_D16_UNORM_S8_UINT && format <= VK_FORMAT_D32_SFLOAT_S8_UINT` range only covers formats with stencil, but there are valid depth formats without stencil too. So to fix there should be a separate if statement that checks for depth-only formats:
```
if (format >= VK_FORMAT_D16_UNORM_S8_UINT && format <= VK_FORMAT_D32_SFLOAT_S8_UINT)
	aspect = VK_IMAGE_ASPECT_DEPTH_BIT | VK_IMAGE_ASPECT_STENCIL_BIT;
else if (format == VK_FORMAT_D16_UNORM || format == VK_FORMAT_D32_SFLOAT)
	aspect = VK_IMAGE_ASPECT_DEPTH_BIT;
```

source/vulkan/runtime_vk.cpp
```diff
@@ -1183,7 +1226,10 @@ bool reshade::vulkan::runtime_vk::init_technique(technique &info, VkShaderModule
 				num_depth_attachments = 1;
 				const uint32_t attach_idx = num_color_attachments;
 
-				attachment_views[attach_idx] = _default_depthstencil_view;
+				if (_depthstencil_image_view == VK_NULL_HANDLE)
+					attachment_views[attach_idx] = _default_depthstencil_view;
+				else
+					attachment_views[attach_idx] = _depthstencil_image_view;
```
crosire: This is incorrect. Should always be `_default_depthstencil_view`, like it was before. ReShade effects are only allowed to render in the special depth-stencil created by ReShade (which only the stencil is used of, see also implementation in other APIs).

source/vulkan/runtime_vk.cpp
```diff
@@ -1349,7 +1395,7 @@ void reshade::vulkan::runtime_vk::render_technique(technique &technique)
 			begin_info.framebuffer = _swapchain_frames[_swap_index];
 		vk.CmdBeginRenderPass(cmd_list, &begin_info, VK_SUBPASS_CONTENTS_INLINE);
 
-		// Setup states
+		// Setup states
```
crosire: =)

source/vulkan/vulkan_hooks.cpp
```diff
@@ -454,14 +688,449 @@ VkResult VKAPI_CALL vkQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR *pPr
...
+	assert(trampoline != nullptr);
+
+	trampoline(commandBuffer, indexCount, instanceCount, firstIndex, vertexOffset, firstInstance);
+
+	// retrieve the current depth stencil image view
+	reshade::vulkan::draw_call_tracker *command_buffer_tracker = &s_trackers_per_command_buffer.at(commandBuffer);
+	command_buffer_tracker->on_draw(command_buffer_tracker->_depthstencil, indexCount * instanceCount);
+}
+
+void vkCmdDrawIndexedIndirect(VkCommandBuffer commandBuffer, VkBuffer buffer, VkDeviceSize offset, uint32_t drawCount, uint32_t stride)
```
crosire: No need to hook this. It is impossible to get draw call statistics from an indirect draw call, so I'd just ignore those.

source/vulkan/vulkan_hooks.cpp
```diff
@@ -301,12 +479,28 @@ VkResult VKAPI_CALL vkCreateDevice(VkPhysicalDevice physicalDevice, const VkDevi
...
+
+	// retrieve the associated command queues
+	for (uint32_t i = 0; i < pCreateInfo->queueCreateInfoCount; i++)
+	{
+		VkDeviceQueueCreateInfo queueInfo = pCreateInfo->pQueueCreateInfos[i];
+
+		for (uint32_t j = 0; j < queueInfo.queueCount; j++)
+		{
+			VkQueue queue = VK_NULL_HANDLE;
+			vkGetDeviceQueue(device, queueInfo.queueFamilyIndex, j, &queue);
```
crosire: It is not valid to call Vulkan functions directly. They need to go through the dispatch tables.

source/vulkan/vulkan_hooks.cpp
```diff
@@ -236,7 +414,6 @@ VkResult VKAPI_CALL vkCreateDevice(VkPhysicalDevice physicalDevice, const VkDevi
 	dispatch_table.QueuePresentKHR = (PFN_vkQueuePresentKHR)get_device_proc(device, "vkQueuePresentKHR");
 	dispatch_table.QueueWaitIdle = (PFN_vkQueueWaitIdle)get_device_proc(device, "vkQueueWaitIdle");
 	dispatch_table.DeviceWaitIdle = (PFN_vkDeviceWaitIdle)get_device_proc(device, "vkDeviceWaitIdle");
-	dispatch_table.GetDeviceQueue = (PFN_vkGetDeviceQueue)get_device_proc(device, "vkGetDeviceQueue");
```
crosire: Mmh?

source/vulkan/vulkan_hooks.cpp
```diff
@@ -417,6 +650,7 @@ VkResult VKAPI_CALL vkCreateSwapchainKHR(VkDevice device, const VkSwapchainCreat
 	if (!runtime->on_init(*pSwapchain, *pCreateInfo, s_surface_windows.at(pCreateInfo->surface)))
 		LOG(ERROR) << "Failed to initialize Vulkan runtime environment on runtime " << runtime.get() << '.';
 
+	s_device_runtimes[device] = runtime;
```
crosire: A bit questionable, considering that there may be multiple swapchains associated with a device and as such multiple runtimes. I realize the same issue exists in the other APIs and there we simply only update the first runtime, but that's not so great either.

source/vulkan/runtime_vk.cpp
```diff
@@ -398,20 +419,40 @@ void reshade::vulkan::runtime_vk::on_reset()
...
 	runtime::on_present();
+
+	clear_DSV_iter = 1;
+	_current_tracker->reset();
+}
+
+void reshade::vulkan::runtime_vk::on_create_graphics_pipelines(const VkGraphicsPipelineCreateInfo* pCreateInfos)
+{
+	if (pCreateInfos->sType == VK_STRUCTURE_TYPE_PIPELINE_MULTISAMPLE_STATE_CREATE_INFO)
+		_is_multisampling_enabled = true;
```
crosire: This is not doing what you want. Just because a single pipeline has MSAA active, doesn't mean all depth-stencils are multisampled. Better to just ignore this for now (also avoids having to hook `vkCreateGraphicsPipelines`, which is a bonus).

Thalixte:
Thks for all your usefull comments. Will treat them as soon as i can =)

source/vulkan/runtime_vk.cpp
Thalixte: Ok, will check this ;-)

ReShade.vcxproj
Thalixte: I reintroduced them, bit i was forced to remove the ID3D12GraphicsCommandList5 API methods (because of the d3d12.h lib used).

source/vulkan/runtime_vk.cpp
Thalixte: So, i need to make _default_depthstencil_view point to the depthstencil image replacement in the create_depthstencil_replacement method. What happens if the image format does not match with it ? ?

source/vulkan/runtime_vk.cpp
Thalixte: Fixed =)

source/vulkan/vulkan_hooks.cpp
Thalixte: OK. Removed...

source/vulkan/vulkan_hooks.cpp
Thalixte: I guess it is fixed now

source/vulkan/vulkan_hooks.cpp
Thalixte: There is already this code in line 410.

source/vulkan/runtime_vk.cpp
Thalixte: OK. I removed it.

source/vulkan/runtime_vk.cpp
crosire: This attachment is for ReShade effects that use stenciling, like SMAA. It has nothing to do with depth buffer input. So nothing to modify here.

source/vulkan/vulkan_hooks.cpp
crosire: Ah, ok.

source/d3d12/d3d12_command_list.cpp
```diff
@@ -6,23 +6,66 @@
 #include "log.hpp"
 #include "d3d12_device.hpp"
 #include "d3d12_command_list.hpp"
+#include <assert.h>
+#include "runtime_d3d12.hpp"
 
-D3D12CommandList::D3D12CommandList(D3D12Device *device, ID3D12GraphicsCommandList *original) :
+D3D12CommandList::D3D12CommandList(D3D12Device *device, UINT nodeMask, D3D12_COMMAND_LIST_TYPE type, ID3D12CommandAllocator *pCommandAllocator, ID3D12PipelineState *pInitialState, ID3D12GraphicsCommandList *original) :
```
crosire: Would be great if you could just revert these two files to how they are in master.

source/vulkan/runtime_vk.cpp
Thalixte: Oh. I see... I confused this code with another. Reverted...

source/d3d12/d3d12_command_list.cpp
Thalixte: Yes, but i have to update the dx12.h ref. I guess it has somthing to do with this commit: https://github.com/crosire/reshade/commit/40621845a7bea529e336e1013450de96c883256d (Revert all projects back to Windows SDK 10.0.17763.0)

Thalixte: I guess i have to revert back to Windows SDK 10.0.18362.0.

crosire: I already fixed D3D12CommandList to work with 10.0.17763.0, so you shouldn't need to modify it at all. Just merge master into your branch and keep the files as they are and it should work.

Thalixte: Ok. Will do it as soon as i can ;-)

Thalixte: It's done...

source/vulkan/vulkan_hooks.cpp
```diff
@@ -454,14 +693,399 @@ VkResult VKAPI_CALL vkQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR *pPr
...
+					track_cleared_depthstencil(commandBuffer, device, *command_buffer_tracker, depthstencil, image_view_data);
+			}
+		}
+	}
+
+	// important: the trampoline method must be executed after the depth buffer image saving
+	trampoline(commandBuffer, attachmentCount, pAttachments, rectCount, pRects);
+}
+
+VkResult VKAPI_CALL vkEndCommandBuffer(VkCommandBuffer commandBuffer)
```
crosire: This is not doing anything from what I can see, so probably do not need to hook it.

source/vulkan/vulkan_hooks.cpp
```diff
@@ -454,14 +693,399 @@ VkResult VKAPI_CALL vkQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR *pPr
...
+		s_command_buffer_mapping[commandBuffer] = device;
+
+		reshade::vulkan::draw_call_tracker command_buffer_tracker;
+		add_command_buffer_trackers(commandBuffer, command_buffer_tracker);
+	}
+
+	return result;
+}
+
+VkResult VKAPI_CALL vkCreateRenderPass(VkDevice device, const VkRenderPassCreateInfo* pCreateInfo, const VkAllocationCallbacks* pAllocator, VkRenderPass* pRenderPass)
```
crosire: This too does not seem to do anything, so probably don't need to hook it either.

source/vulkan/vulkan_hooks.cpp
```diff
@@ -454,14 +693,399 @@ VkResult VKAPI_CALL vkQueuePresentKHR(VkQueue queue, const VkPresentInfoKHR *pPr
...
+		return result;
+	}
+
+#if RESHADE_VULKAN_CAPTURE_DEPTH_BUFFERS
+	// create an image view where depth buffer is displayable in the shaders
+	VkImageViewCreateInfo createInfo = *pCreateInfo;
+	createInfo.subresourceRange.aspectMask = VK_IMAGE_ASPECT_DEPTH_BIT;
+
+	VkImageView pViewReplacement = VK_NULL_HANDLE;
+	result = trampoline(device, &createInfo, nullptr, &pViewReplacement);
```
crosire: It's probably better to avoid creating a second image view for every image view the game creates. Maybe check whether the format in `pCreateInfo` is a depth format and then modify the aspect mask to always contain `VK_IMAGE_ASPECT_DEPTH_BIT` (before calling the trampoline to create the original image view). That way there is only one image view, which should support what we want it to do.

source/vulkan/vulkan_hooks.cpp
```diff
@@ -12,19 +12,195 @@
...
 static std::unordered_map<VkPhysicalDevice, VkInstance> s_instance_mapping;
+static std::unordered_map<VkQueue, VkDevice> s_queue_mapping;
+static std::unordered_map<VkCommandBuffer, VkDevice> s_command_buffer_mapping;
+static std::unordered_map<VkDevice, std::shared_ptr<reshade::vulkan::runtime_vk>> s_device_runtimes;
 static std::unordered_map<VkSwapchainKHR, std::shared_ptr<reshade::vulkan::runtime_vk>> s_runtimes;
+static std::unordered_map<VkImage, image_data> s_depth_stencil_buffer_images;
+static std::unordered_map<VkImageView, image_view_data> s_depth_stencil_buffer_imageViews;
+static std::unordered_map<VkFramebuffer, attachment_data> s_depth_stencil_buffer_frameBuffers;
+
+static uint32_t find_memory_type_index(const VkPhysicalDeviceMemoryProperties &props, VkMemoryPropertyFlags flags, uint32_t type_bits)
```
crosire: This is unused.

source/vulkan/runtime_vk.cpp
```diff
@@ -898,8 +942,10 @@ bool reshade::vulkan::runtime_vk::compile_effect(effect_data &effect)
 			image_binding.imageView = _backbuffer_texture_view[info.srgb];
 			break;
 		case texture_reference::depth_buffer:
-			// TODO (set to backbuffer for now to make validation layers happy)
-			image_binding.imageView = _backbuffer_texture_view[0];
+			if(_depthstencil_image_view == VK_NULL_HANDLE)
+				image_binding.imageView = _default_depthstencil_view;
```
crosire: Maybe add a comment here as to why this is set (because it technically is incorrect and is just done to make validation layers happy).

Thalixte: OK, will see what happens when i comment those lines...

source/vulkan/vulkan_hooks.cpp
Thalixte: OK. Done

source/vulkan/vulkan_hooks.cpp
Thalixte: OK. Done

source/vulkan/vulkan_hooks.cpp
Thalixte: OK. Done.

source/vulkan/runtime_vk.cpp
Thalixte: Done.

source/vulkan/vulkan_hooks.cpp
Thalixte: Just tried it in Doom, but it appears that updating an actual image does not give to me the depth buffer, wherea the actuel code i made provide a valid one.

PR #139, opened 2019-08-02
Add filter to preset selector

seri14:
filter_demo

PR #140, opened 2019-08-16
Shortcut keys for switching backward or forward between presets

antagonicus:
Added shortcut keys for switching backward or forward between presets. They can be configured in the settings page jut below "Effect Toggle Key". Also the switching is no longer circular, It does not go directly from the last to the first preset anymore. For example: when using depth of field filter with manual focus and  want to quickly go to the last preset (or the first), it will no longer go circular if pressing the shortcut key once more by accident.
There aren't so many changes in code as the git is showing, because there was a re-indentation of a  big block of c ode (which I think I could avoid it in fact, sorry for that)

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -79,6 +79,8 @@ reshade::runtime::runtime() :
 	_reload_key_data(),
 	_effects_key_data(),
 	_screenshot_key_data(),
+	_previous_preset_key_data(),
+	_next_preset_key_data(),
 	_screenshot_path(g_target_executable_path.parent_path())
 {
 	// Default shortcut PrtScrn
@@ -174,7 +176,21 @@ void reshade::runtime::on_present()
 
 		if (_input->is_key_pressed(_screenshot_key_data))
 			_should_save_screenshot = true;
+
+		bool bPreviousPresetKeyPressed = _input->is_key_pressed(_previous_preset_key_data);
+		bool bNextPresetKeyPressed = _input->is_key_pressed(_next_preset_key_data);
+		if (bPreviousPresetKeyPressed || bNextPresetKeyPressed)
+			if(switch_current_preset(bNextPresetKeyPressed))
+			{
+				_last_preset_switching_time = current_time;
+				_is_in_between_presets_transition = true;
+				set_current_preset();
+				save_config();
+			}
 	}
+
+	if (_is_in_between_presets_transition)
+		load_current_preset();
 
 #if RESHADE_GUI
 	// Draw overlay
@@ -882,6 +898,9 @@ void reshade::runtime::load_config()
 	config.get("INPUT", "KeyReload", _reload_key_data);
 	config.get("INPUT", "KeyEffects", _effects_key_data);
 	config.get("INPUT", "KeyScreenshot", _screenshot_key_data);
+	config.get("INPUT", "KeyPreviousPreset", _previous_preset_key_data);
+	config.get("INPUT", "KeyNextPreset", _next_preset_key_data);
+	config.get("INPUT", "PresetTransitionDelay", _preset_transition_delay);
 
 	config.get("GENERAL", "PerformanceMode", _performance_mode);
 	config.get("GENERAL", "EffectSearchPaths", _effect_search_paths);
@@ -921,6 +940,9 @@ void reshade::runtime::save_config(const std::filesystem::path &path) const
 	config.set("INPUT", "KeyReload", _reload_key_data);
 	config.set("INPUT", "KeyEffects", _effects_key_data);
 	config.set("INPUT", "KeyScreenshot", _screenshot_key_data);
+	config.set("INPUT", "KeyPreviousPreset", _previous_preset_key_data);
+	config.set("INPUT", "KeyNextPreset", _next_preset_key_data);
+	config.set("INPUT", "PresetTransitionDelay", _preset_transition_delay);
 
 	config.set("GENERAL", "PerformanceMode", _performance_mode);
 	config.set("GENERAL", "EffectSearchPaths", _effect_search_paths);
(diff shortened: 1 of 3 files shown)


seri14:
I understand that you just added toggle keys and extract the iteration of filesystem items.
but I like choosing the presets circulatly.
Can you give still/GIF images and preset files?
I guess it is a Effect's problem, so should not fix this function.

antagonicus:
I have updated the git repository with round cycling through presets

antagonicus:
Here are 3 preset files using Depth of field shader (DOF.fx)

[presets.zip](https://github.com/crosire/reshade/files/3521681/presets.zip)

seri14:
Sorry I misunderstood that it is a **problem** about DOF focusing.
You explain was about user use case, so there was no need to upload presets.

antagonicus:
Yes, in the case of DOF filter I prefer  to stop on the last preset file, instead of going around to the first one. So the source code now has this stoping as a comment, and for me i just uncomment that part and comment the one that does the circular cycling of presets.

antagonicus:
I made a new update on this fork repository with code for smooth transition between preset files with configurable delay in settings. This new setting is "preset transition delay" and is a "floating point" slider with value between 0.0 and 10.0 (default 1.0).
So using shortcut keys with DOF filter is working very nice and movie like.

antagonicus:
Sorry, the slider is a "int" with values in the range 0 milliseconds to 10*1000 (ms).,
New update on the repository with transition delay setting caption updated:

antagonicus:
It is strange that my last commits that go to the repository that i have forked from yours do not show on this thread as the first commit i did, on those 3 files. Do I need to open another "pull request"?

seri14:
try git push to your remote/master branch as you pushed when creating this pull request.

antagonicus:
Ok, i made the `git push` on my branch. But now it seems to be in conflict with yours, because of that new update you made with _"Keep showing preset selector if require to reload effects_"

seri14:
Actually, your changes is conflicted with #139 now. (It was almost reimplementing.)
#141 is not merged to crosire/master branch yet.
https://github.com/crosire/reshade/commit/66555ca7811cf3ef93110f9ac7a770dda283c823
so now, you should do reimplement based on #139.
But if you can wait to merging #141, you just add features `shortcut key` and `minimum input waiting time` without `draw_preset_explorer()` changes.

antagonicus:
I'll wait. But i'm confused. It seems that i have to make another fork on your updated repository and take my changes there. And the another pull request

seri14:
Revert to start of your forked branch. But you working on master branch in this case.
so require some difficult operations.
1. `git revert --no-edit --no-commit ...9c9d2d5db74f35e8cc7f8548d761713d122b65fe`
2. `git commit --message="Revert ～"`
3. `git cherry-pick 9c9d2d5db74f35e8cc7f8548d761713d122b65fe...(#141 SHA1)`
4. Reimplement.
5. `git commit --message="Added shortcut keys for ～"`
6. `git push https://github.com/antagonicus/reshade.git master:master`

antagonicus:
i ran all the git commands you told me and made the update, but on this thread it still says that there is a conflict on `gui.cpp`

seri14:
I am sorry my english is bad. **You must do it after merging #141**. 
I guess cherry-pick was failed because `(#141 SHA1)` is not exists in crosire's master branch yet.
so currently, this is same as before runs these commands.
**Please wait until #141 is merged**.
And please search/understand these git commands yourself.

crosire:
I can try and fix the merge conflicts myself after #140 has been merged.

crosire:
Fixed conflicts for you. Just FYI: It might take one or two weeks still until I can merge this, since I've got some other stuff on the plate. It would be great if you could fix formatting to follow the same style as the rest of ReShade's source code in the meantime (there are some variables that are not snake_case in there).

antagonicus:
I've made the update on code formating. 
Regarding your last comment about keys: i don't know or understand what you're suggesting.
A note: i've notticed that it would be nice and good to define shortcut keys not only for the shaders, but also for "boolean" properties exposed by shaders, so i'll do this next

seri14:
In currently, if user keep to pressing the shortcut key then
shortcut
so I suggested to use `is_key_released()`.

And please fix where calling `switch_to_next_preset(bool reversed)` function from
```cpp
// runtime.cpp L183
if(switch_to_next_preset(is_next_preset_key_pressed))
```
to
```cpp
if(switch_to_next_preset(is_previous_preset_key_pressed))
```

And sorry I did not noticed this missing change from
```cpp
	// reshade.cpp L943
	// Select a default preset file if none exists yet or not own
	if (path_state == path_state::invalid)
		_current_preset_path = g_reshade_dll_path.parent_path() / L"DefaultPreset.ini";
	else
		_current_preset_path = g_reshade_dll_path.parent_path() / current_preset_path;

	for (const auto &callback : _load_config_callables)
		callback(config);
```
to
```cpp
	// Select a default preset file if none exists yet or not own
	if (path_state == path_state::invalid)
		_current_preset_path = g_reshade_dll_path.parent_path() / L"DefaultPreset.ini";
	else
		_current_preset_path = g_reshade_dll_path.parent_path() / current_preset_path;

#if RESHADE_GUI
	_current_browse_path = _current_preset_path.parent_path();
#endif

	for (const auto &callback : _load_config_callables)
		callback(config);
```

antagonicus:
Yep, code updated

source/runtime.cpp
```diff
@@ -174,8 +176,33 @@ void reshade::runtime::on_present()
 
 		if (_input->is_key_pressed(_screenshot_key_data))
 			_should_save_screenshot = true;
+
+		if (!_is_previous_preset_key_pressed && !_is_next_preset_key_pressed)
```
seri14: You should block to loading effects from user action.

```cpp
&& !is_loading() && _reload_compile_queue.empty()
```

source/runtime.cpp
```diff
@@ -174,8 +176,33 @@ void reshade::runtime::on_present()
...
+		if ((_is_previous_preset_key_pressed && _input->is_key_released(_previous_preset_key_data[0])) ||
+			(_is_next_preset_key_pressed && _input->is_key_released(_next_preset_key_data[0])))
+		{
+			// Ok, the preset shortcut key was released, so reset everything so to be prepared for a new transition
+			_is_previous_preset_key_pressed = false;
+			_is_next_preset_key_pressed = false;
+		}
 	}
 
+	if (_is_in_between_presets_transition)
```
seri14: Maybe same problem

seri14:
I found some problems.

1. I think it is not good to transitioning the combo value (e.g. LUT selection in MultiLUT)
2. I feel the integer value does not transition smoothly
```
// Traced value during transitioning from 0 to 16 in every frame
0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 1 2 2 3 4 5 6 7 8 9 10 11 12 13 14 16
// Traced value during transitioning from 16 to 0 in every frame
15 14 13 12 11 10 9 8 7 6 5 4 3 2 1 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 0 
```
3. Preprocessor definitions (e.g. Different layer files between presets)

antagonicus:
Updated as requested, but with some notes:
- for int values does not make too much sense to smooth the transition and does not work well with the way i made the transitioning code. I personaly only use float transition in games. So i removed the code for int values
- I dont have a solution for presets with diferent shaders list, so i added a tooltip on the preset transition slider saying that in this case the user should set it 0 ms delay

source/runtime.cpp
```diff
@@ -960,10 +960,6 @@ void reshade::runtime::load_config()
 	else
 		_current_preset_path = g_reshade_dll_path.parent_path() / current_preset_path;
 
-#if RESHADE_GUI
-	_current_browse_path = _current_preset_path.parent_path();
-#endif
```
seri14: This is required 😢
This code will make sense if the current preset is not same as the DLL folder.

crosire: This wasn't in master before though... So must have gotten lost somewhere else previously already.

antagonicus:
How can i help?

crosire:
Not your mistake. Just something that wrong before your pull request, so don't worry.

PR #141, opened 2019-08-25
Keep showing preset selector if require to reload effects

seri14:
And contain these changes
* Fix browse path when user quit input mode
* Extract the process of set current preset path to for ready to accept #140 changes

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -1321,10 +1321,17 @@ void reshade::runtime::reset_uniform_value(uniform &variable)
 
 bool reshade::runtime::switch_to_next_preset(bool reversed)
 {
-	std::filesystem::path preset_container_path = (g_reshade_dll_path.parent_path() / _current_browse_path).lexically_normal();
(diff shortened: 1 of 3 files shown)


source/runtime.cpp
```diff
@@ -1299,43 +1319,45 @@ void reshade::runtime::reset_uniform_value(uniform &variable)
...
-{
-	set_current_preset(_current_preset_path);
-}
-void reshade::runtime::set_current_preset(std::filesystem::path path)
+bool reshade::runtime::switch_to_next_preset(bool reversed)
 {
-	enum class path_state { invalid, valid };
-	static const std::filesystem::path reshade_container_path = g_reshade_dll_path.parent_path();
+	std::filesystem::path preset_container_path = (g_reshade_dll_path.parent_path() / _current_browse_path).lexically_normal();
+	std::filesystem::path presets_filter_text = _current_browse_path.filename();
```
crosire: This uses `_current_browse_path`, so looks like that field needs to be moved outside the `RESHADE_GUI` block in `runtime.hpp`.

PR #142, opened 2019-09-07
Change to displaying size bytes with binary prefix

seri14:
Windows OS uses binary prefix.
mbfile

### DIFF
--- a/source/gui.cpp
+++ b/source/gui.cpp
@@ -1302,8 +1302,8 @@ void reshade::runtime::draw_overlay_menu_statistics()
 
 			post_processing_memory_size += memory_size;
 
-			if (memory_size > 1024 * 1024) {
-				memory_view = std::ldiv(memory_size, 1024 * 1024);
+			if (memory_size > 1048576) {
+				memory_view = std::ldiv(memory_size, 1048576);
 				memory_view.rem /= 1000;
 				memory_size_unit = "MB";
 			}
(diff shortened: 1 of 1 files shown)


crosire:
Windows is wrong. I know it shows it like that, but it's still wrong. Windows shows what is generally agreed to be KiB (kibibyte), MiB (mebibyte),... Those are multiples of 1024. KB (kilobyte), MB (megabyte) however by definition are multiples of 1000. Kilo = 1000, Mega = 100000, etc. and applies to sizes too: https://en.wikipedia.org/wiki/Metric_prefix#List_of_SI_prefixes

seri14:
I learned "Windows uses binary prefix" at Windows XP era.
I can understand if displaying the "1024KB" when user created a 1MB texture. (for example 512 * 512 * 4 bytes)
Actually, it displaying "1048kB".
And if create 500 * 500 * 4 byte texture, displaying "1000kB".

I do not think 1048 and 500 * 500 is commonly.

crosire:
I can understand that. But please use the correct suffix then: `MiB` and `KiB`. I do not want to display data that is technically wrong, only because Windows does it too (that does not make it right or better). Other software is also slowly starting to use the correct terms (e.g. Steam).

crosire:
Also I suggest writing `1048576` as `1024 * 1024` to make it clearer.

PR #143, opened 2019-09-10
!!! BUGFIX !!! Force depth texture access to shader in d3d12

Thalixte:
In some D3D12 games (for instance, FIFA 19), Reshade crashes when trying to create a graphic pipeline state, or a new shader resource view based on the depth image.

This is due to the fact that the depth image resource was created with the D3D12_RESOURCE_FLAG_DENY_SHADER_RESOURCE flag. It not only disallows the shader access to this resource, it also crashes the CreateShaderResourceView method or the CreateGraphicsPipelineState method.

Now, FIFA 19 D3D12 has a proper access to depth buffer without crashing. As it is a Frostbite 3 game, maybe this fix can be usefull for some other D3D12 games ;-)

FIFA_19_depth_map_1
FIFA_19_depth_map_2

### DIFF
--- a/source/d3d12/d3d12_device.cpp
+++ b/source/d3d12/d3d12_device.cpp
@@ -513,6 +513,8 @@ D3D12_HEAP_PROPERTIES STDMETHODCALLTYPE D3D12Device::GetCustomHeapProperties(UIN
 }
 HRESULT STDMETHODCALLTYPE D3D12Device::CreateCommittedResource(const D3D12_HEAP_PROPERTIES *pHeapProperties, D3D12_HEAP_FLAGS HeapFlags, const D3D12_RESOURCE_DESC *pResourceDesc, D3D12_RESOURCE_STATES InitialResourceState, const D3D12_CLEAR_VALUE *pOptimizedClearValue, REFIID riidResource, void **ppvResource)
 {
+	D3D12_RESOURCE_DESC new_desc = *pResourceDesc;
+	new_desc.Flags = new_desc.Flags & ~D3D12_RESOURCE_FLAG_DENY_SHADER_RESOURCE;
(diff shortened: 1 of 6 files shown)


source/d3d12/d3d12_device.cpp
```diff
@@ -513,7 +513,9 @@ D3D12_HEAP_PROPERTIES STDMETHODCALLTYPE D3D12Device::GetCustomHeapProperties(UIN
 }
 HRESULT STDMETHODCALLTYPE D3D12Device::CreateCommittedResource(const D3D12_HEAP_PROPERTIES *pHeapProperties, D3D12_HEAP_FLAGS HeapFlags, const D3D12_RESOURCE_DESC *pResourceDesc, D3D12_RESOURCE_STATES InitialResourceState, const D3D12_CLEAR_VALUE *pOptimizedClearValue, REFIID riidResource, void **ppvResource)
 {
-	return _orig->CreateCommittedResource(pHeapProperties, HeapFlags, pResourceDesc, InitialResourceState, pOptimizedClearValue, riidResource, ppvResource);
+	D3D12_RESOURCE_DESC new_desc = *pResourceDesc;
+	new_desc.Flags = new_desc.Flags & ~D3D12_RESOURCE_FLAG_DENY_SHADER_RESOURCE;
```
crosire: We should probably check wether the resource format is a depth format and only remove the flag for those. There might be a performance implication, so it's better to avoid removing the flag for textures that are not interesting to ReShade anyway.

Thalixte: That's what i thinked about, but it seems that this flag is only allowed for depthstencil resources, so a check may be not necessary...

crosire: Ah, right, just read that in the documentation. Ok then.

PR #144, opened 2019-09-14
Add I/O caching for ini files

seri14:
And change the line outputs to non-random for ini file.

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -184,8 +184,8 @@ void reshade::runtime::on_present()
 	// Reset input status
 	_input->next_frame();
 
-	// To delay saving .ini files
-	ini_file::save_cache();
+	// Save .ini files
+	ini_file::cache_loop();
 
 	static int cooldown = 0, traffic = 0;
 
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -955,10 +955,8 @@ void reshade::runtime::save_config() const
 		callback(config);
 }
 
-void reshade::runtime::load_preset(const std::filesystem::path &path)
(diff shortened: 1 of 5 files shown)


source/runtime.cpp
```diff
@@ -1143,7 +1132,12 @@ void reshade::runtime::save_screenshot(const std::wstring &postfix, const bool s
 	}
 	else if (_screenshot_include_preset && should_save_preset)
 	{
-		save_preset(least + L".ini");
+		std::error_code ec;
+
+		if (ini_file::load_cache(_current_preset_path).flush())
```
crosire: I think I'd prefer there to be an `ini_file::flush_cache(path)` function which does this instead. That way `ini_file::flush` can be made private (like `ini_file::save` was before).

source/runtime.cpp
```diff
@@ -973,7 +970,7 @@ void reshade::runtime::load_preset(const std::filesystem::path &path)
 	if (_reload_remaining_effects != 0 && // ... unless this is the 'load_current_preset' call in 'update_and_render_effects'
 		(_performance_mode || preset_preprocessor_definitions != _preset_preprocessor_definitions))
 	{
-		assert(path == _current_preset_path);
+		assert(preset.path() == _current_preset_path);
```
crosire: This assert is no longer necessary, since the path is guaranteed to always be `_current_preset_path` (since it's opened with that above). So can get rid of it.
Conveniently it's also the only use of `ini_file::path()`, so can get rid of that function too.

crosire:
Sorry for the mess of referencing commits. GitHub messed up the merge (didn't do a squash) and my first attempt didn't retain author info. Last one worked though.

PR #145, opened 2019-09-20
UWP compatibility

Thalixte:
This is an attempt to make Reshade works with UWP games. I need your help to make it work (if it is possible).

I managed to make the GUI appear in Forza Horizon 4, as well as make the basic shaders work (those that do not need depth buffer access)...

here is a log: https://pastebin.com/pbcgh42C

FH4
Forza Horizon 4

Forza Horizon 4 good depth buffer

### DIFF
--- a/source/dxgi/dxgi.cpp
+++ b/source/dxgi/dxgi.cpp
@@ -16,6 +16,8 @@
 #include "d3d12/runtime_d3d12.hpp"
 #include <CoreWindow.h>
 
+extern bool g_isUWP;
+
 static void dump_swapchain_desc(const DXGI_SWAP_CHAIN_DESC &desc)
 {
 	LOG(INFO) << "> Dumping swap chain description:";
@@ -242,6 +244,7 @@ HRESULT STDMETHODCALLTYPE IDXGIFactory2_CreateSwapChainForCoreWindow(IDXGIFactor
 	if (com_ptr<ICoreWindowInterop> window_interop; SUCCEEDED(pWindow->QueryInterface(&window_interop)))
 		window_interop->get_WindowHandle(&hwnd);
 
+	g_isUWP = hwnd != nullptr;
 	init_reshade_runtime_d3d(*ppSwapChain, direct3d_version, device_proxy, hwnd);
 
 	return hr;
--- a/source/dxgi/dxgi_swapchain.cpp
+++ b/source/dxgi/dxgi_swapchain.cpp
@@ -288,7 +288,7 @@ HRESULT STDMETHODCALLTYPE DXGISwapChain::GetDesc(DXGI_SWAP_CHAIN_DESC *pDesc)
 }
 HRESULT STDMETHODCALLTYPE DXGISwapChain::ResizeBuffers(UINT BufferCount, UINT Width, UINT Height, DXGI_FORMAT NewFormat, UINT SwapChainFlags)
 {
-	/*LOG(INFO) << "Redirecting IDXGISwapChain::ResizeBuffers" << '(' << this << ", " << BufferCount << ", " << Width << ", " << Height << ", " << NewFormat << ", " << std::hex << SwapChainFlags << std::dec << ')' << " ...";
+	LOG(INFO) << "Redirecting IDXGISwapChain::ResizeBuffers" << '(' << this << ", " << BufferCount << ", " << Width << ", " << Height << ", " << NewFormat << ", " << std::hex << SwapChainFlags << std::dec << ')' << " ...";
 
 	runtime_reset();
 
@@ -303,7 +303,7 @@ HRESULT STDMETHODCALLTYPE DXGISwapChain::ResizeBuffers(UINT BufferCount, UINT Wi
 		return hr;
 	}
 
-	runtime_resize();*/
+	runtime_resize();
 
 	return S_OK;
 }
--- a/source/dxgi/dxgi_swapchain.cpp
+++ b/source/dxgi/dxgi_swapchain.cpp
@@ -45,6 +45,12 @@ DXGISwapChain::DXGISwapChain(D3D12Device *device, IDXGISwapChain3 *original, con
 	_direct3d_device(device, false),
 	_direct3d_version(12),
 	_runtime(runtime) {}
+DXGISwapChain::DXGISwapChain(D3D12Device *device, IDXGISwapChain4 *original, const std::shared_ptr<reshade::runtime> &runtime) :
+	_orig(original),
+	_interface_version(4),
+	_direct3d_device(device, false),
+	_direct3d_version(12),
+	_runtime(runtime) {}
 
 void DXGISwapChain::runtime_reset()
 {
@@ -85,7 +91,7 @@ void DXGISwapChain::runtime_resize()
 		break;
 	case 12:
 		assert(_runtime != nullptr);
-		initialized = std::static_pointer_cast<reshade::d3d12::runtime_d3d12>(_runtime)->on_init(desc);
+		initialized = std::static_pointer_cast<reshade::d3d12::runtime_d3d12>(_runtime)->on_init(desc, desc.OutputWindow);
 		break;
 	}
 
@@ -282,7 +288,7 @@ HRESULT STDMETHODCALLTYPE DXGISwapChain::GetDesc(DXGI_SWAP_CHAIN_DESC *pDesc)
 }
 HRESULT STDMETHODCALLTYPE DXGISwapChain::ResizeBuffers(UINT BufferCount, UINT Width, UINT Height, DXGI_FORMAT NewFormat, UINT SwapChainFlags)
 {
-	LOG(INFO) << "Redirecting IDXGISwapChain::ResizeBuffers" << '(' << this << ", " << BufferCount << ", " << Width << ", " << Height << ", " << NewFormat << ", " << std::hex << SwapChainFlags << std::dec << ')' << " ...";
+	/*LOG(INFO) << "Redirecting IDXGISwapChain::ResizeBuffers" << '(' << this << ", " << BufferCount << ", " << Width << ", " << Height << ", " << NewFormat << ", " << std::hex << SwapChainFlags << std::dec << ')' << " ...";
 
 	runtime_reset();
 
@@ -297,9 +303,9 @@ HRESULT STDMETHODCALLTYPE DXGISwapChain::ResizeBuffers(UINT BufferCount, UINT Wi
 		return hr;
 	}
 
-	runtime_resize();
+	runtime_resize();*/
 
-	return hr;
+	return S_OK;
 }
 HRESULT STDMETHODCALLTYPE DXGISwapChain::ResizeTarget(const DXGI_MODE_DESC *pNewTargetParameters)
 {
--- a/source/dxgi/dxgi_swapchain.cpp
+++ b/source/dxgi/dxgi_swapchain.cpp
@@ -91,7 +91,7 @@ void DXGISwapChain::runtime_resize()
 		break;
 	case 12:
 		assert(_runtime != nullptr);
-		initialized = std::static_pointer_cast<reshade::d3d12::runtime_d3d12>(_runtime)->on_init(desc, desc.OutputWindow);
+		initialized = std::static_pointer_cast<reshade::d3d12::runtime_d3d12>(_runtime)->on_init(desc);
 		break;
 	}
 
@@ -300,12 +300,12 @@ HRESULT STDMETHODCALLTYPE DXGISwapChain::ResizeBuffers(UINT BufferCount, UINT Wi
 	else if (FAILED(hr))
 	{
 		LOG(ERROR) << "> IDXGISwapChain::ResizeBuffers failed with error code " << std::hex << hr << std::dec << '!';
-		return hr;
+		return S_OK;
 	}
 
 	runtime_resize();
 
-	return S_OK;
+	return hr;
 }
 HRESULT STDMETHODCALLTYPE DXGISwapChain::ResizeTarget(const DXGI_MODE_DESC *pNewTargetParameters)
 {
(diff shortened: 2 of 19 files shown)


source/dxgi/dxgi.cpp
```diff
@@ -126,17 +237,23 @@ static void init_reshade_runtime_d3d(T *&swapchain, unsigned int direct3d_versio
 	}
 	else if (direct3d_version == 12)
 	{
-		if (com_ptr<IDXGISwapChain3> swapchain3; SUCCEEDED(swapchain->QueryInterface(&swapchain3)))
+		if (com_ptr<IDXGISwapChain3> swapchain4; SUCCEEDED(swapchain->QueryInterface(&swapchain4)))
```
crosire: There was no change here, so the variable name change is probably unintentional.

source/dxgi/dxgi.cpp
```diff
@@ -126,17 +237,23 @@ static void init_reshade_runtime_d3d(T *&swapchain, unsigned int direct3d_versio
...
-			const auto runtime = std::make_shared<reshade::d3d12::runtime_d3d12>(command_queue->_device->_orig, command_queue->_orig, swapchain3.get());
-			if (!runtime->on_init(desc))
+			const auto runtime = std::make_shared<reshade::d3d12::runtime_d3d12>(command_queue->_device->_orig, command_queue->_orig, swapchain4.get());
+
+			HWND window = desc.OutputWindow;
+
+			if (window == nullptr)
+				window = UWPWindow;
+
+			if (!runtime->on_init(desc, window))
```
crosire: This can be simplified by writing:
```c++
if (desc.OutputWindow == nullptr)
    desc.OutputWindow = UWPWindow;

const auto runtime = ...
if (!runtime->on_init(desc))
...
```
This is because `runtime_d3d12` will use the `OutputWindow` field and the `desc` structure is writable anyway. Doing it this way avoids the changes necessary to `runtime_d3d12`.

source/dxgi/dxgi_swapchain.cpp
```diff
@@ -45,6 +45,12 @@ DXGISwapChain::DXGISwapChain(D3D12Device *device, IDXGISwapChain3 *original, con
 	_direct3d_device(device, false),
 	_direct3d_version(12),
 	_runtime(runtime) {}
+DXGISwapChain::DXGISwapChain(D3D12Device *device, IDXGISwapChain4 *original, const std::shared_ptr<reshade::runtime> &runtime) :
+	_orig(original),
+	_interface_version(4),
+	_direct3d_device(device, false),
+	_direct3d_version(12),
+	_runtime(runtime) {}
```
crosire: This is unused.

source/dxgi/dxgi_swapchain.cpp
```diff
@@ -298,7 +304,7 @@ HRESULT STDMETHODCALLTYPE DXGISwapChain::ResizeBuffers(UINT BufferCount, UINT Wi
 
 	runtime_resize();
 
-	return hr;
+	return S_OK;
```
crosire: This swallows the `DXGI_ERROR_INVALID_CALL` error and does not report it to the application. I have seen applications ignore that error and still work, hence why ReShade has to ignore it too and resize anyway, but it should probably still be reported to the application. Unless this causes problems with UWP games?

source/hook.cpp
```diff
@@ -46,6 +46,8 @@ reshade::hook::status reshade::hook::uninstall()
 	if (!valid())
 		return hook::status::unsupported_function;
 
+	s_export_hook_path.clear();
```
crosire: I don't understand this change.

crosire:
Did some research and it looks like there is a better solution to retrieving the window handle for an UWP window: https://docs.microsoft.com/en-us/windows/win32/api/corewindow/nn-corewindow-icorewindowinterop.
Haven't tested yet, but something like this could work and avoid the process searching code:
```
#include <CoreWindow.h>
...
HRESULT STDMETHODCALLTYPE IDXGIFactory2_CreateSwapChainForCoreWindow(...)
{
    ...
    HWND hwnd = nullptr;
    if (ICoreWindowInterop *window_interop; SUCCEEDED(pWindow->QueryInterface(&window_interop)))
        window_interop->get_WindowHandle(&hwnd);

    init_reshade_runtime_d3d(*ppSwapChain, direct3d_version, device_proxy, hwnd);

    return hr;
}
```

Thalixte:
> Did some research and it looks like there is a better solution to retrieving the window handle for an UWP window: https://docs.microsoft.com/en-us/windows/win32/api/corewindow/nn-corewindow-icorewindowinterop.
> Haven't tested yet, but something like this could work and avoid the process searching code:
> 
> ```c++
> #include <CoreWindow.h>
> ...
> HRESULT STDMETHODCALLTYPE IDXGIFactory2_CreateSwapChainForCoreWindow(...)
> {
>     ...
>     HWND hwnd = nullptr;
>     if (ICoreWindowInterop *window_interop; SUCCEEDED(pWindow->QueryInterface(&window_interop)))
>         window_interop->get_WindowHandle(&hwnd);
> 
>     init_reshade_runtime_d3d(*ppSwapChain, direct3d_version, device_proxy, hwnd);
> 
>     return hr;
> }
> ```

Works perfectly. Thks ;-)

crosire:
Is the `Sleep(5000)` still necessary with this? It's odd.

source/dxgi/dxgi_swapchain.cpp
Andon13: I suspect the problem being masked here is one caused by the flip presentation model that UWP requires rather than anything UWP specific.

Flip model changes DWM composition behavior to share resource views rather than copy. Views of swapchain buffers are discarded immediately as part of Present (...)'s design.

In other words, any swapchain RenderTargetView that was attached to a device context before presentation is implictly discarded after. ReShade's stateblock code is explicitly re-attaching views that were discarded and only the application itself should be doing that or you will break a lot of resource management.

If this is the problem you can verify it by issuing a non-blocking Present (...) call and trying ResizeBuffers (...) a second time. If that works, you need to re-design runtime::on_present (...)'s stateblock to eliminate these side effects.

Thalixte:
> Is the `Sleep(5000)` still necessary with this? It's odd.

Yes, it is ;-) I found it necessary to avoid Reshade to crash when effects are enabled. This seems to be timing issues, as well as concurrency issues, but as i don't manage to make the step by step debug mode to work with Reshade + Forza Horizon 4, I only debug with the logs, which is a painfull task.
Reshade can crashes when effects are enabled. Here is the error in the logs:
```
21:53:10:179 [12700] | ERROR | Failed to create pipeline for pass 0 in technique 'AdvancedCRT'! HRESULT is '0x887a0005'.
21:53:10:179 [12700] | ERROR | Failed to compile "C:\Loisirs\Sport\Forza Horizon 4\reshade-shaders\Shaders\CRT.fx":

21:53:10:181 [12700] | ERROR | Failed to create texture 'V__AS_Pass0Tex' (Width = 1920, Height = 1080, Format = 34)! HRESULT is '0x887a0005'.
```
In this version, i also don't manage to make the inputs to work, so i must force the display of the Reshade UI.
The depth buffer access do not work, as there is no drawcalls or vertices stats.
Maybe this is because of this log entry:
```
21:52:42:437 [12700] | INFO  | Redirecting ID3D12Device::CreateCommandQueue(000001AEA2462120, 00000020E52FFAD8, {0EC870A6-5D7E-4C22-8CFC-5BAAE07616ED}, 00000020E52FFB80) ...
21:52:42:466 [12700] | INFO  | Redirecting ID3D12Device::CreateCommandQueue(000001AEA2462120, 00000020E52FFAD8, {0EC870A6-5D7E-4C22-8CFC-5BAAE07616ED}, 00000020E52FFB80) ...
21:52:42:493 [12700] | INFO  | Redirecting ID3D12Device::CreateCommandQueue(000001AEA2462120, 00000020E52FFAD8, {0EC870A6-5D7E-4C22-8CFC-5BAAE07616ED}, 00000020E52FFB80) ...
21:52:42:545 [07572] | INFO  | Redirecting RegisterClassW(00000020E16FD7E0 { MediaPlayer SMTC window - {323F1A73-5A65-4009-95EA-5FE25E9738E0} }) ...
21:52:42:547 [07572] | INFO  | > Adding 'CS_OWNDC' window class style flag to 'MediaPlayer SMTC window - {323F1A73-5A65-4009-95EA-5FE25E9738E0}'.
21:52:42:552 [12700] | INFO  | Loading image files for textures ...
21:53:09:759 [12700] | INFO  | Redirecting IDXGISwapChain::ResizeBuffers(000001AECD4F2750, 2, 1920, 1080, 28, 0x840) ...
```
Maybe Reshade is linked to the MediaPlayer SMTC instead of the game itself...

source/dxgi/dxgi.cpp
Thalixte: Yes, i missed it, thks. Will fix it as soon as i can.

crosire:
I suspect draw call stats are not recorded because ReShade has to hook the command list interface for that, but it currently only supports `ID3D12GraphicsCommandList` and `ID3D12GraphicsCommandList1`. There are `ID3D12GraphicsCommandList2`, `...3` and `...4` too however, which this game might use, but which ReShade currently ignores (see `D3D12Device::CreateCommandList`).

source/dxgi/dxgi.cpp
Thalixte: OK. Done

source/hook.cpp
Thalixte: Removed it. It is just that s_export_hook_path.clear(); makes the game to crash with an invalid hook call, because it tries to hook an empty module path

Thalixte:
yeah, that's seem to be this ! But, unfortunately, the drawcall stats make me think that i am not on the right window:

Forza_Horizon_4_stats

source/runtime.cpp
```diff
@@ -30,9 +30,14 @@ static bool find_file(const std::vector<std::filesystem::path> &search_paths, st
 	if (path.is_relative())
 		for (const auto &search_path : search_paths)
 		{
-			const std::filesystem::path canonical_search_path = std::filesystem::canonical(g_reshade_dll_path.parent_path() / search_path, ec);
+			std::filesystem::path canonical_search_path = std::filesystem::canonical(g_reshade_dll_path.parent_path() / search_path, ec);
 
+			// Fix for UWP games (ec returns an access denied error)
 			if (ec || canonical_search_path.empty())
+				if (search_path.is_absolute())
+					canonical_search_path = search_path;
```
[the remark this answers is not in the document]
Thalixte: Yes, really ;-) The reason is simple: when the first line try to retrieve the canonical_search_path, it fails due to an access denied error, so ec returns en error code.
In the reshade.ini file, i use an abolsute path to specify the shaders and textures path.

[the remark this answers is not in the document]
Thalixte: Yes, really ;-) The reason is simple: when the first line tries to retrieve the canonical_search_path, it fails due to an access denied error, so ec returns en error code.
In the reshade.ini file, i use an absolute path to specify the shaders and textures path.
For the moment, i cannot use relative paths for this UWP game (due to access rights).
What you have made in your last PR work for any other type of games, it is just a specificity of those UWP titles...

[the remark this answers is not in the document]
seri14: Is same error with change to use "canonical" to "absolute"?
If it is, I will try with this title.

source/dxgi/dxgi.cpp
```diff
@@ -126,17 +127,23 @@ static void init_reshade_runtime_d3d(T *&swapchain, unsigned int direct3d_versio
 	}
 	else if (direct3d_version == 12)
 	{
-		if (com_ptr<IDXGISwapChain3> swapchain3; SUCCEEDED(swapchain->QueryInterface(&swapchain3)))
+		if (com_ptr<IDXGISwapChain4> swapchain4; SUCCEEDED(swapchain->QueryInterface(&swapchain4)))
```
crosire: But why? =P Shouldn't SwapChain3 be supported too?

crosire:
1 draw call with 3 vertices sound very much like a fullscreen triangle being drawn to copy data from somewhere else onto the screen ... I suggest checking all the created command queues. It may be that the game is doing all the rendering on a different command queue than is used to display the image.

source/dxgi/dxgi.cpp
Thalixte: Yes, you're right =p reverted back to IDXGISwapChain3 (IDXGISwapChain4 and HDR is not the purpose of this PR)...

Thalixte: @crosire do you know why i cannot make the input to work correctly on this UWP title ? I looked at the ICoreInputInterop::SetInputSource method, but i don't know where to link the input to the CoreWindow...

source/runtime.cpp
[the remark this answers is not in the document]
Thalixte: Yes, using std::filesystem::absolute instead of std::filesystem::canonical fixes the pb ;-)

Thalixte:
> 1 draw call with 3 vertices sound very much like a fullscreen triangle being drawn to copy data from somewhere else onto the screen ... I suggest checking all the created command queues. It may be that the game is doing all the rendering on a different command queue than is used to display the image.

I managed to filter the commandQueue, so i now only have one queue ;-). For the drawcalls, the pb was the MSAA indicator that did not display the right info: MSAA was enabled.

Now, i have good drawcall stats:

Forza_Horizon_4_stats

Thalixte:
If i use more than one shader, i have a device lost error.... But if i use only one, it works.

Most of the time, the depth buffer does not display correctly:

Forza Horizon 4 depth buffer

But sometimes, it does !!! It's seem to be linked to the shader initialisation...

Forza Horizon 4 good depth buffer

Forza Horizon 4 good depth buffer 2

source/runtime.cpp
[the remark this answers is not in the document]
crosire: 01540646d4801fe51c5bd1d53c40e92f3939704e

crosire:
Closing since the relevant bits have been addressed/merged in 2cb53f3dfbfa3058bd239aeac23173be4dbd6363 and 991a3c0da27cc2a9a4a3e088e94a1850ed8c427b.

PR #147, opened 2019-10-02
Optimization Mode

seri14:
Add support `const uniform` syntax.
And I will add new ReShadeUI definition `__const_uniform`.

```hlsl
// Source code
__const_uniform int iLUT_Selector ...
```
```hlsl
// Result for supported version
const uniform int iLUT_Selector ...

// Result for unsupported version
uniform int iLUT_Selector ...
```

### DIFF
--- a/source/effect_codegen_hlsl.cpp
+++ b/source/effect_codegen_hlsl.cpp
@@ -394,7 +394,7 @@ private:
 
 		define_name<naming::unique>(res, "_Globals_" + info.name);
 
-		if (_uniforms_to_spec_constants && info.has_initializer_value)
+		if ((_uniforms_to_spec_constants || (info.type.has(type::q_const) && info.type.has(type::q_uniform))) && info.has_initializer_value)
 		{
 			std::string &code = _blocks.at(_current_block);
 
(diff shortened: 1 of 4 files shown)


crosire:
I don't think I understand the purpose of this yet. Could you elaborate?

seri14:
The purpose of working this patch is bring the performance of techniques closing to performance mode.

The following variables are not change frequently:
* Additional effect on/off (e.g. DOF)
* Texture switching (e.g. MultiLUT, Bloom)
* Visual effects for debugging (e.g. DisplayDepth, MXAO)

So I will make possible to editing users-frequently-editing variables include above variables ​and possible to optimize above variables that improve in performance mode.

These are known cases about developers should not do:
* "drag" and "slider" is not suitable to use this feature.

seri14:
To on/off Optimization Mode, set macro the `__RESHADE_OPTIMIZATION_MODE__` to `1` or `0`
```hlsl
#if __RESHADE_OPTIMIZATION_MODE__ == 1
// This means ON
#define __const_uniform const uniform
#else
// This means OFF and unsupported version
#define __const_uniform uniform
#endif
```

crosire:
But why do we need this if performance mode is a thing?
I'm more inclined to change the code so that perfomance mode is activated automatically if the user closes the UI and disabled again if the user opens the UI. The shader compilation has gotten pretty fast, so this may start to become feasable.

seri14:
It is easier than this patch.

PR #152, opened 2019-10-05
Fixed ReShade Setup save paths with ANSI encoding

seri14:


seri14:
Changed codes to remove UTF-16 support.

PR #153, opened 2019-10-05
Support for several emulators (Pcsx2, Dolphin, Ppsspp, gliden64...)

Boulotaur2024:
Clean final commit -ready for merging.
OpenGL backend only.

Reshade forum thread (screenshots, details) :
https://reshade.me/forum/general-discussion/5442-release-misc-emulators-with-depth-buffer-access

List of supported emulators :
- Pcsx2
- Redream
- Dolphin
- Ppsspp
- Gliden64

### DIFF
--- a/source/opengl/opengl_hooks.cpp
+++ b/source/opengl/opengl_hooks.cpp
@@ -714,7 +714,7 @@ extern "C"  void WINAPI glFramebufferTextureLayer(GLenum target, GLenum attachme
 	static const auto trampoline = reshade::hooks::call(glFramebufferTextureLayer);
 	trampoline(target, attachment, texture, level, layer);
 
-	if (g_current_runtime) g_current_runtime->on_fbo_attachment(attachment, GL_TEXTURE_2D, texture, level);
+	if (g_current_runtime) g_current_runtime->on_fbo_attachment(attachment, target, texture, level);
 }
 extern "C"  void WINAPI glFramebufferTextureLayerARB(GLenum target, GLenum attachment, GLuint texture, GLint level, GLint layer)
 {
@@ -1958,7 +1958,20 @@ HOOK_EXPORT void WINAPI glTexParameteriv(GLenum target, GLenum pname, const GLin
 	static const auto trampoline = reshade::hooks::call(glTexParameteriv);
 	trampoline(target, pname, params);
 }
+HOOK_EXPORT void WINAPI glTexStorage2D(GLenum target, GLsizei levels, GLenum internalformat, GLsizei width, GLsizei height)
+{
+	if (g_current_runtime) g_current_runtime->on_dimensions_sizing((GLfloat)0, width, height);
 
+	static const auto trampoline = reshade::hooks::call(glTexStorage2D);
+	trampoline(target, levels, internalformat, width, height);
+}
+HOOK_EXPORT void WINAPI glTexStorage3D(GLenum target, GLsizei levels, GLenum internalformat, GLsizei width, GLsizei height, GLsizei depth)
+{
+	if (g_current_runtime) g_current_runtime->on_dimensions_sizing((GLfloat)0, width, height);
+
+	static const auto trampoline = reshade::hooks::call(glTexStorage3D);
+	trampoline(target, levels, internalformat, width, height, depth);
+}
 HOOK_EXPORT void WINAPI glTexSubImage1D(GLenum target, GLint level, GLint xoffset, GLsizei width, GLenum format, GLenum type, const GLvoid *pixels)
 {
 	static const auto trampoline = reshade::hooks::call(glTexSubImage1D);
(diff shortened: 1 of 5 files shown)


source/opengl/runtime_gl.cpp
```diff
@@ -1214,23 +1557,31 @@ void reshade::opengl::runtime_gl::detect_depth_source()
...
-				continue; // Skip candidates that were not used during rendering
-
-			// Detection heuristic based on dimensions and usage
-			if (((it.second.width > _width * 0.95 && it.second.width < _width * 1.05) && (it.second.height > _height * 0.95 && it.second.height < _height * 1.05)) &&
-				((it.second.num_vertices * (1.2f - float(it.second.num_drawcalls) / _drawcalls)) >= (best_info.num_vertices * (1.2f - float(best_info.num_drawcalls) / _drawcalls))))
+			// Auto-detect Depth Buffer or pick selected one
+			if (_altDetection && it.first == _currentDepthBuffer || _manualDetection && it.first == _selectedDepthBuffer) {
+				foundDepthBuffer = true;
+				best_info = it.second;
+				best_match = it.first;
```
crosire: Just `break`ing here would avoid the additional `!_altDetection` check below.

source/opengl/runtime_gl.cpp
```diff
@@ -109,10 +112,25 @@ reshade::opengl::runtime_gl::runtime_gl()
...
 		config.set("OPENGL", "ForceMainDepthBuffer", _force_main_depth_buffer);
+		config.set("OPENGL", "IsUpscaled", _isUpscaled);
+		config.set("OPENGL", "AltDetection", _altDetection);
+		config.set("OPENGL", "SelectedDepthBuffer", _selectedDepthBuffer);
+		config.set("OPENGL", "AppName", _AppName);
+		config.set("OPENGL", "AppMode", _AppMode);
+		config.set("OPENGL", "AppWidth", _AppWidth);
+		config.set("OPENGL", "AppHeight", _AppHeight);
 	});
+	setAppName(_AppName);
```
crosire: I really don't want any application specific code in ReShade. Anything that is needed for a specific application could become necessary somewhere else too, so it should always be generalized.

Boulotaur2024:
> Thank you for your great work!
> 
> Unfortunately a few of the changes conflict with core ReShade principles (being as generic as possible), so I cannot merge it as is. Plus there are some formatting issues (whitespaces and ReShade codebase uses snake case). Please don't take that negatively. I just tend to have a very specific idea of what the ReShade code should contain, which isn't always easy to uphold. If you don't want to continue working on this, I'll try and selectively merge some things not affected by the above and look for solutions for the things that are.

Hello Crosire, thanks a lot for the review.
Reshade is amazing and really neatly done, I forgot to tell you. Now it's done.

I perfectly agree with your nitpicks here and there and I'm willing to find a solution for them -I know my code is not perfect in some places and I suck at C++- *but* I worry more about the fact that you are against application-specific code (is that even proper english, lol ?).
Even though I completely understand your approach : being as generic as possible.
I remember Durante (GeDoSaTo) handled it by designing plugins for specific games without touching the core (see the "GenericDepth" plugin I/we made years ago). That might be a good idea to implement.

Not meant as an excuse but I think OpenGL is dying, there is room for creativity...
Up to you.

I'll try to find some time for bringing changes according to your excellent review
Have a good evening,

crosire:
Yeah, at some point I want to have plugin support too ..... But I haven't gotten around to convince myself to start working on that (since it evolves so many things, like designing the API, finding places where plugins can hook into, worrying about cheating, etc.). And I get the OpenGL hacks. I just wonder if some things can be solved in a way that is more future proof (even though the API is going away in time).

It would help to understand how the Redream issue manifests. Like, how many render contexts (HGLRC) are created, do they use different device contexts (HDC) or not, do the device contexts point to different windows (HWND) or not etc. I might be able to read some stuff into this if you manage to capture and upload a log file somewhere (since that logs most of that stuff). I don't have an AMD card to try myself.

I also understand the `on_fbo_attachment` fix for Dolphin now. Got me thinking and it should work to just pass in `GL_TEXTURE_2D_ARRAY` in the `glFramebufferTextureLayer` instead of `GL_TEXTURE_2D`. Because `glFramebufferTextureLayer` is made for array textures, so that should use an array binding and it not doing so was an oversight when I wrote that. That way we can avoid the Dolphin specific workaround.

The scaling code maybe could be made a feature where the user has to enter the original resolution in the UI (instead of selecting an app and have the dimensions derived from that). Would get rid of the app-specific code at the cost of a little bit of simplicity for users.

As for the whitespace issues and formatting: I can fix those and push that to your branch (so the pull request is updated) if you don't want to worry about that yourself.

source/opengl/opengl_hooks.cpp
```diff
@@ -714,7 +714,7 @@ extern "C"  void WINAPI glFramebufferTextureLayer(GLenum target, GLenum attachme
 	static const auto trampoline = reshade::hooks::call(glFramebufferTextureLayer);
 	trampoline(target, attachment, texture, level, layer);
 
-	if (g_current_runtime) g_current_runtime->on_fbo_attachment(attachment, GL_TEXTURE_2D, texture, level);
+	if (g_current_runtime) g_current_runtime->on_fbo_attachment(attachment, target, texture, level);
```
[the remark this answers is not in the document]
crosire: Scratch that. As mentioned below, you are right that this was wrong before. But passing in `GL_TEXTURE_2D_ARRAY` here should fix it.

crosire:
I fixed the problem with cropped depth buffers in 0a3d362d35e8dad96ef492e6ea928606911494a9. This was actually a bug in ReShade caused by it reusing FBOs for multiple tasks. But the driver crops FBO dimensions to the smallest attachments, which in these cases always was the backbuffer dimensions (so all depth buffers were cropped to that).
I tested in CEMU and it works fine there now, I suspect the same is true for many of the other emulators now too.

deton24:
If it's relevant information in this topic.
Ishiruka Dolphin worked fine with Reshade installed as OpenGL, and D3D11 set in emulator settings. If it was set to OGL, I think it didn't work causing crash (or started constantly crashing after few tries).

PR #154, opened 2019-10-06
Add cache for effect

seri14:
This pull request has following changes:
* Cache preprocessed effect files. (It is about the "Generated HLSL".)
* Cache compiled effects.

Total time of loading 161 effects in my PC:
(preprocess & parse time) + (compile time) = (Total time until display splash screen)
First load (no cache): 12.3 sec + 20.7 sec = 33.0 sec
~Second load (cached): 6.8 sec + 3.1 sec = 9.9 sec~
Second load (cached): 2.5 sec + 2.7 sec = 5.2 sec (https://github.com/crosire/reshade/pull/154/commits/3b66b11e9a6668a4c1138b86f22ba01f89fd61b8)

### DIFF
--- a/ReShadeFX.vcxproj
+++ b/ReShadeFX.vcxproj
@@ -150,6 +150,7 @@
     <ClInclude Include="source\effect_expression.hpp" />
     <ClInclude Include="source\effect_lexer.hpp" />
     <ClInclude Include="source\effect_parser.hpp" />
+    <ClInclude Include="source\effect_parser_injector.hpp" />
     <ClInclude Include="source\effect_preprocessor.hpp" />
     <ClInclude Include="source\effect_symbol_table.hpp" />
   </ItemGroup>
--- a/ReShadeFX.vcxproj.filters
+++ b/ReShadeFX.vcxproj.filters
@@ -15,6 +15,7 @@
     <ClInclude Include="source\effect_expression.hpp" />
     <ClInclude Include="source\effect_lexer.hpp" />
     <ClInclude Include="source\effect_parser.hpp" />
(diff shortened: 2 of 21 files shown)


seri14:
The commit for the module caching is incremental compilation instead of in-memory cache actually.

crosire:
This is an awesome start!

I don't really like the removal of the constant conversion and uniform injector solution though. Would prefer if that stays the way it was before.
There also is Vulkan still where we need to cache the generated SPIR-V binaries instead of HLSL/GLSL source code.

seri14:
Injection is an alternative solution for performance mode.
The "_performance_mode" is user application code, so I thought "This is improvement to more removing user application code from Reshade FX".

It make well-known parser variables for any applications is only "__PIXEL_SIZE" by removing "SPEC_CONSTANT_".

If we will changes something like as performance mode in the future, I think this revertion will be an important indicator for future consideration.
Do you have an idea of ​​an alternative solution?

crosire:
Makes sense. But this currently makes performance mode useless unfortunately, since no effects in the wild use the `const uniform` syntax and we can't expect all authors to update their code (reshade-shaders repository only hosts a subset of effects).
I don't know of a better solution. `SPEC_CONSTANT_*` (which stands for specialization constant in case you are wondering) are only generated if the parser is called with the relevant option enabled, so it's fair to assume an application calling the parser with that active knows how to handle it.

seri14:
Excuse me. This is not the Optimization Mode (#147).
This PR contains change to just removing "SPEC_CONSTANT_*".
This still have the 3035 error.
Please correct.

seri14:
I give up

PR #155, opened 2019-10-09
Shortcut keys for the shader exposed variables of types: bool, combo, list and radio.

antagonicus:
Pressing the shortcut will cycle through all the possible values of that variable.

reshade_variable_shortcut


crosire:
I suggest you reset your branch to master and then apply those changes again (manually). There have been quite some invasive changes since #140, so the codebase is no longer easily mergeable.

PR #162, opened 2019-11-24
Attempt to link cleared depth buffer instances to draw stats

Thalixte:


Thalixte:
This fixes the depth buffer selection in Bioshock 2 remastered

PR #163, opened 2019-11-27
VS2019 Build Fixes

Joshua-Ashton:
Misc. fixes/changes to get Reshade compiling on VS2019.

### DIFF
--- a/source/d3d10/d3d10.cpp
+++ b/source/d3d10/d3d10.cpp
@@ -93,7 +93,7 @@ HOOK_EXPORT HRESULT WINAPI D3D10CreateDeviceAndSwapChain1(IDXGIAdapter *pAdapter
 
 	// Create device proxy unless this is a software device
 	D3D10Device *device_proxy = nullptr;
-	if (DriverType == D3D_DRIVER_TYPE_WARP || DriverType == D3D_DRIVER_TYPE_REFERENCE)
+	if (D3D_DRIVER_TYPE(DriverType) == D3D_DRIVER_TYPE_WARP || D3D_DRIVER_TYPE(DriverType) == D3D_DRIVER_TYPE_REFERENCE)
 	{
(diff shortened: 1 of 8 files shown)


crosire:
Unfortunately these changes are incorrect and will likely break unicode support. The reason this may not compile on VS2019 is because ReShade code base is C++17, but C++20 [introduced a new u8 char type](http://www.open-std.org/jtc1/sc22/wg21/docs/papers/2018/p0482r6.html) which is incompatible with char as before. So to make it work change the project configuration to target C++17 instead of C++20/c++latest. Eventually the code can be updated to use the new u8 char type. But I want to keep VS2017 compatibility for a while, which does not support C++20.

crosire:
The last two commits should fix this (haven't yet installed VS2019, so cannot confirm myself until I have).

Joshua-Ashton:
> Unfortunately these changes are incorrect and will likely break unicode support.

The encoding in a std::string is undefined, as long as you are consistent with using UTF-8 (which I believe it is, although I am not sure about my changes to `input.cpp`) and the execution charset is UTF-8 then it should be fine?

crosire:
Nope. Because the conversion will not be correct. The string returned by `std::filesystem::path::string()` will use the current thread code page to determine how to convert the path from wide char (how its stored internally in `std::filesystem::path`) to string. By default this is not UTF-8, but dependent on the system locale. Unless explicitly set.

Joshua-Ashton:
Thats what I meant about the execution charset.

crosire:
Since ReShade runs in threads created by the target application it cannot guarantee the execution charset is UTF-8. So it would have to set that each time. But then the application might get messed up. So it would have to reset it back to what the application set before each time too. That seems awfully unnecessary when there are explicit functions for this too.

Joshua-Ashton:
Makes sense then :)

PR #165, opened 2019-11-29
Effect preprocessor definitions edition

Thalixte:
I created this PR to answer a request made by Ceejay.dk.

It adds the ability, in the Reshade UI, to edit the effect preprocessor definitions.

To do so, Reshade parses the preprocessor expressions, that look like:
```
#ifndef FXAA_PS3
    #define FXAA_PS3 0
#endif
```
and add the corresponding preprocessor definition in the UI, in the effect variable editor:

Shadow Warrior 2 effect preprocessor_definitions

Each time a preprocessor definition is updated, the user can press the "Apply" button in front of this definition, in order to recompile the effect with the new preprocessor value.
The user also has the ability to reset the effect preprocessor definitions, even when an effect compilation gives an error.


source/runtime.hpp
```diff
@@ -288,6 +295,7 @@ namespace reshade
 		bool _last_reload_successful = true;
 		bool _should_save_screenshot = false;
 		bool _is_in_between_presets_transition = false;
+		bool _effect_preprocessor_modified = false;
```
crosire: There is no need for a global variable like this. This is only ever set to true in one place, so just saving the preset there directly avoid all this bookkeeping.

source/runtime.hpp
```diff
@@ -348,6 +356,7 @@ namespace reshade
 		bool _show_splash = true;
 		bool _show_code_editor = false;
 		bool _show_screenshot_message = true;
+		bool _show_single_effect_splash = false;
```
crosire: Just use the default splash via `_show_splash` I think. This overcomplicates things unnecessarily.

Thalixte: I use a specific bollean var, in order to display a specific splsh message, because the standard one displays too much information when compiling just one technique. Also, it allows me to display the message for only one second instead of five (which is the case in the standard splash message)

Thalixte:
Thls Crosire for your feedback. So, there was already an attempt for this kind of feature ;-)

In the effect_preprocessor.cpp, i think i made very few modifications. It is just that i rely on the add_macro_definition fonctions to add the effect preprocessor definitions to the preprocessor parser, in a dedicated array. Maybe i should have not change the name of the macro structure, but as the other structures have the _info suffix, i thought it was a good idea to change its name.
As for the uniform vars, i thought it was a good idea to use them, because it allowed me to make less modifications ti the gui.cpp code. Also, i thought it could allow us to use some kind of annotation stuff to display inputs according to the preprocessor var type, so we can rely on the gui_widgets already created.
In the #110 PR, the gui code is a little bit complex. Of course, i can use only a preprocessor_definitions array, and display it in the UI, but i wanted those inputs to be placed in a specific category for each technique, after the existing inputs.

source/runtime.hpp
Thalixte: Maybe we can refactor it by changing the _show_splash var type and associate to it three states (OFF, DISPLAY, PREPROCESSOR_DISPLAY)

source/runtime.hpp
Thalixte: There was a request to also save the preset when the "reload" all the effects button was clicked, or when the user use the shortcut key combination to reload the effects, or when switching to performance mode. That's why i added this global var (in order to use it at several places in the gui.cpp and the runtime.cpp code).

crosire: But why would it be necessary to save the preset when the user reloads effects? The preset is already saved when any value is modified (and should simply be saved when a preprocessor definition is modified too). So saving it again would be superfluous.

Thalixte: OK. Made the simplification, but i kept the _effect_preprocessor_modified  var in order to toggle the apply button...

Thalixte:
Closed this since Crosire integrated it in the 1c0014e commit.

PR #166, opened 2019-12-12
Fix remaining issues in dx12 depth buffer detection + workaround for clearing instance with no drawcalls

Thalixte:
This PR aims at fixing the following points:

**1. Workaround for clearing instances with no drawcalls**:
In both depth buffer detection modes, in some titles, the depth buffer can be cleared at the beginning of the frame pipeline, before any draw calls occure. So we get some clearing instances with 0 drawcalls. To avoid this, a global stats var has been introduced in the buffer_detection classes, in order to keep track of the current stats from the previous frame (in other words, the stats between the last clearing instance of the previous frame, and the first clearing instance of the new one). I choosed to use a global var in order to share those stats with all the command lists (because we don't know a-priori which one is used to clear the depth buffer).

**2. Flickering issue in copy mode:**
In copy mode, it is better to select the best depth buffer texture to be copied directly on the fly, and not at the end of the frame, because in this case, the copy can occure on the wrong depth buffer texture during one frame, if the number of clearing instances raises or decreases, which makes the depth buffer display flickering. The drawback is that Reshade will copy more than one texture during the frame, but with the latest modifications in the depth buffer detection code, it is still restricted to less than four copies (even in a game with a lot of clearing instances, like Agent Of Mayhem).

**3. DX12 stats initialization when switching from copy mode to standard mode, and vice versa:**
In both modes, in DX12, when switching from one mode to another, the stats are not correctly initialized, except if the runtime is recreated. This is due to the fact that the _depthstencil_resources_by_handle map is cleared when this operation occures, and is not repopulated after. I added a specific flag in the reset method of the buffer_detection class, in order to  avoid this clearing when the switching occures.

**4. Use the number of vertices to choose the best clearing instance for the best dsv_texture candidate:**
Now, the auto detection of the best clearing instance for the best dsv_texture candidate relies on the number of vertices, not on the clearing index anymore.

Those fixes has been tested on the following games, and seem to work perfectly:

- Bioshock 2 HD Remastered (DX11)
- Agent Of Mayhem (DX11)
- Assassin's Creed Origin (DX11)
- Assassin's Creed Odyssey (DX11)
- Shadow Warriors 2 (DX11)
- Remedy's Control (DX11 + DX12)
- Devil May Cry 5 (DX12)
- Elex (DX11)
- The Witcher 3 Wild Hunt (DX11)
- Project Cars 2 (DX11)


crosire:
Awesome! I only have one little objection, but that's it:
The `s_auto_copy` and `s_previous_stats` variables should not be globals, but rather part of the `buffer_detection_context` classes (of which only 1 exists per device). You can then access these via the `_context` field. So, e.g. in `on_clear_depthstencil`:
```c++
if (counters.current_stats.drawcalls == 0)
    counters.current_stats = _context->_previous_stats;
```

Thalixte:
Thks crosire ;-) Made the changes you requested.

PR #167, opened 2019-12-21
fix issues with building on gcc

DadSchoorse:
std::fmod with float arguments should be the same as std::fmodf, for some reason gcc has no std::fmodf.


crosire:
Awesome! Also went ahead and incorporated your changes from https://github.com/DadSchoorse/vkBasalt/issues/41#issuecomment-568204327 in 173cb19ecc4e62d78a26dba6fab8660dae3cd9b1 to be able to build the HLSL codegen and FXC project too.

PR #168, opened 2020-03-19
Fix gcc warnings

DadSchoorse:
This fixes `-Wswitch` and `-Wmissing-field-initializers`.

There are still `-Wchar-subscripts` warnings in `effect_lexer.cpp`, for example here: https://github.com/crosire/reshade/blob/1773c0f23ab7bb067f9a9544c6f0a304708997b1/source/effect_lexer.cpp#L669

but I didn't know an easy way to fix those.

Well and `-Wunknown-pragmas` is still left, but I think you want those. :frog:

### DIFF
--- a/source/effect_codegen_spirv.cpp
+++ b/source/effect_codegen_spirv.cpp
@@ -1138,6 +1138,8 @@ private:
 						assert(op.from.is_integral());
 						spv_op = op.from.is_signed() ? spv::OpConvertSToF : spv::OpConvertUToF;
 						break;
+					default:
+						break;
 					}
 
 					result = add_instruction(spv_op, convert_type(op.to))
@@ -1246,6 +1248,8 @@ private:
 					break;
 				}
 				assert(false);
+				break;
+			default:
 				break;
 			}
 		}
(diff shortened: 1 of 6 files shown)


crosire:
I've altered your changes slightly by adding comments to make behavior even clearer and committed that (see 905fca0ea71ddc4452465d5aea37ec1a31cc284f). Judging by Godbolt this still got rid of those warnings.

crosire:
Last commit should now have gotten rid of the `-Wchar-subscripts` warnings as well. Closing.

DadSchoorse:
> Last commit should now have gotten rid of the -Wchar-subscripts warnings as well. Closing.

does it fix the underling problem tho? it would still frog stuff up if char is signed on the platform, wouldn't it?

crosire:
It does fix it. All used characters by the lexer are well defined and below 128 (see `type_lookup` content). Any others will always resolve to zero anyway. Signed to unsigned casting too is well defined (in contrast to unsigned to signed casting, which is undefined behavior for values that don't fit). So this simply ensures that negative values won't cause the pointer to become invalid (because addressing some memory before the actual `type_lookup` array), but look up zero instead.

DadSchoorse:
I always thought that signed to unsigned cast is undefined for negative values. But that seems to not be the case, so thanks for fixing these warnings.

PR #173, opened 2020-05-10
fix Reshade detection in Civ VI dx12

Thalixte:
It seems that Civ VI uses the brand new ID3D12Device6 interface.
When enabling it in reshade, Reshade works in Civ VI DX12.


Thalixte:
I also added a fix for Crysis DX10, because depth buffer detection doesn't work anymore in the latest Reshade releases...

source/d3d10/runtime_d3d10.cpp
```diff
@@ -279,6 +279,10 @@ void reshade::d3d10::runtime_d3d10::on_present()
 	if (_backbuffer_resolved != _backbuffer)
 		_device->ResolveSubresource(_backbuffer_resolved.get(), 0, _backbuffer.get(), 0, _backbuffer_format);
 
+	// Setup real back buffer
```
Thalixte, line 282: Those lines are for fixing Crysis DX10 depth buffer detection

source/d3d10/runtime_d3d10.cpp
```diff
@@ -279,6 +279,10 @@ void reshade::d3d10::runtime_d3d10::on_present()
...
+	auto rtv = _backbuffer_rtv[0].get();
+	_device->OMSetRenderTargets(1, &rtv, nullptr);
+
```
Thalixte, line 285: Those lines are for fixing Crysis DX10 depth buffer detection

crosire: But why? As usual I would prefer knowing why this fixes it, since this shouldn't actually do anything.

Thalixte: I must admit i don't know myself why those lines fix the db detection for Crysis. If MSAA was enabled, i would understand, but this is not the case. maybe this could be related to motion blur implementation in this specific game....

crosire:
Btw. the reason I'm not updating to the 18362 SDK is because last time I did that I got reports of crashes. I suspect there to be issues with that SDK and VC++ 2017 (since it's officially only available for VC++ 2019), so the solution is to use the D3D12 header from 18362, but keep the rest as is (which is what 958c0753fff3c42984b3b3f1ce199eca277466db does).

source/d3d10/runtime_d3d10.cpp
crosire: What differs when this is added? Are more draw calls counted (so the correct buffers are already there, just the heuristic picks the wrong one otherwise)? Or are more buffers added to the list?

Thalixte: No, the number of depth buffers are still the same. The initialisation seems to be different. When those lines are not present, there is no depth info at all (the displayDepth shader produces a black screen in normal mode).

Thalixte:
> o

Yes, i know. There must be an issue in the way the SDK is configured. Im myself had to change the C:\Program Files (x86)\Windows Kits\10\DesignTime\CommonConfiguration\Neutral\UAP\10.0.17763.0\UAP.props file:
```
<WindowsSdkDir Condition="'$(WindowsSdkDir)' == ''">$([MSBUILD]::GetDirectoryNameOfFileAbove('$(MSBUILDTHISFILEDIRECTORY)', 'sdkmanifest.xml'))\</WindowsSdkDir>
```
because the Condition was missing (it was present in the 10.0.7763.0 equivalent file):
https://developercommunity.visualstudio.com/content/problem/761788/msb8036-the-windows-sdk-version-100183620-was-not.html

source/d3d10/runtime_d3d10.cpp
Thalixte: Maybe this pb is related to the dsv creation or the depthstencil creation ?

crosire: What happens when you unset all render targets there instead?:
```c++
_device->OMSetRenderTargets(0, nullptr, nullptr);
```

crosire: If that fixes it too, then I think I know what the problem is.
https://docs.microsoft.com/en-us/windows/win32/api/d3d11/nf-d3d11-id3d11devicecontext-pssetshaderresources: ```If an overlapping resource view is already bound to an output slot, such as a rendertarget, then this API will fill the destination shader resource slot with NULL.```
PSSetShaderResources is called first in `render_technique`, which means that if the depth buffer is still bound to the OM (because the game bound it), then it won't actually bound to the ReShade shader.
This also means that you should be able to workaround this by enabling any shader (doesn't matter which one) before DisplayDepth (because ReShade unbinds all render targets after each pass in `render_technique`).

crosire:
The issue with Crysis should be fixed now. Please let me know if it isn't!

PR #177, opened 2020-06-19
fix d3d9 depth buffer detection

Thalixte:
This fix is for d3d9 games where the depth buffer clearance set the depthstencil surface without using the SetDepthStencilSurface method (like RE4 Remastered HD).
In the DX9 tab, make the depth buffer stats similar to the one displayed in the d3d10/d3d11/d3d12 UI.
It also uses a multiple depthstencil surfaces replacement method, in order to fix flickering issues that can occure in some games, because the number of clearing instance evolves on each frame.


source/d3d9/buffer_detection.cpp
```diff
@@ -121,26 +129,48 @@ void reshade::d3d9::buffer_detection::on_clear_depthstencil(UINT clear_flags)
...
+	if (depthstencil == nullptr || _depthstencil_replacement == nullptr || depthstencil != depthstencil_clear_index.first)
 		return;
 
-	counters.clears.push_back(counters.current_stats);
+	// always bound the stats to the original depthstencil surface
+	auto &counters = _counters_per_used_depth_surface[depthstencil];
 
-	if (depthstencil_clear_index == counters.clears.size())
+	// Update stats with data from previous frame
+	if (counters.current_stats.drawcalls == 0 && _first_empty_stats)
```
crosire, line 141: Do we need this? I know this was necessary in D3D11 in some cases, but I'm no aware of cases in D3D9 and I like to avoid code that is not explicitly needed.

source/d3d9/buffer_detection.cpp
```diff
@@ -121,26 +129,48 @@ void reshade::d3d9::buffer_detection::on_clear_depthstencil(UINT clear_flags)
...
 
-	// Reset draw call stats for clears
-	counters.current_stats = { 0, 0 };
+	if (counters.current_stats.vertices <= 4 || counters.current_stats.drawcalls == 0) // Also triggers when '_preserve_depth_buffers' is false, since no clear stats are recorded then
+		return; // Ignore clears when there was no meaningful workload since the last one
+
+	counters.clears.push_back(counters.current_stats);
+
+	// switch to another depthsurface replacement surface
+	switch_depthsurface(depthstencil, counters);
```
crosire, line 154: Considering that I don't think this has to be called in `buffer_detection::reset`, this would be the only other place to call `switch_depthsurface`, so may as well inline it here.

source/d3d9/buffer_detection.hpp
```diff
@@ -27,6 +29,8 @@ namespace reshade::d3d9
 			std::vector<draw_stats> clears;
 		};
 
+		std::vector<com_ptr<IDirect3DSurface9>> _preserved_depthstencil_surfaces;
```
crosire, line 32: This can be `private` ;)

source/d3d9/buffer_detection.hpp
```diff
@@ -39,12 +43,12 @@ namespace reshade::d3d9
...
 		void on_get_depthstencil(IDirect3DSurface9 *&depthstencil);
 		void on_clear_depthstencil(UINT clear_flags);
+		void on_set_viewport(D3DVIEWPORT9 viewport);
 
 		// Detection Settings
 		bool disable_intz = false;
 		bool preserve_depth_buffers = false;
-		UINT depthstencil_clear_index = std::numeric_limits<UINT>::max();
-		UINT depthstencil_clear_index_override = 0;
+		std::pair<com_ptr<IDirect3DSurface9>, UINT> depthstencil_clear_index = { nullptr, 0 };
```
crosire, line 51: I don't think this needs to be a `com_ptr`. There will always be a reference held by `_depthstencil_replacement` or `_preserved_depthstencil_surfaces`. And not having this a `com_ptr` avoid having to release the reference in `buffer_detection::reset`.

source/d3d9/buffer_detection.hpp
Thalixte: Yes ;). Fixed now.

source/d3d9/buffer_detection.cpp
```diff
@@ -15,24 +15,41 @@ void reshade::d3d9::buffer_detection::reset(bool release_resources)
...
+			return;
 
-		// Keep the depth-stencil surface reset back to the replacement, so starting next frame it is the one used again (instead of the original)
-		if (depthstencil != _depthstencil_original)
-			_device->SetDepthStencilSurface(depthstencil.get());
+		// always bound the stats to the original depthstencil surface
+		auto &counters = _counters_per_used_depth_surface[depthstencil];
+
+		// switch to another depthsurface replacement
+		switch_depthsurface(depthstencil, counters);
```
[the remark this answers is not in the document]
Thalixte: This is necessary to switch to the first instance of depthstencil surface in the _preserved_depthstencil_surfaces_vector, in order for the first drawcalls of the next frame to be bound on it.
So it switches to the _preserved_depthstencil_surfaces[0] to prepare the next frame deth buffer detection.

source/d3d9/buffer_detection.cpp
```diff
@@ -72,45 +89,36 @@ void reshade::d3d9::buffer_detection::on_draw(D3DPRIMITIVETYPE type, UINT vertic
...
-		_depthstencil_replacement->GetDesc(&depthstencil_desc);
-
-		if (!check_aspect_ratio(viewport.Width, viewport.Height, depthstencil_desc.Width, depthstencil_desc.Height))
-			return; // Ignore draw calls that go to a smaller viewport (this helps removing infrequent clears in e.g. Mirror's Edge)
-
-		counters.current_stats.vertices += vertices;
-		counters.current_stats.drawcalls += 1;
-	}
+	counters.current_stats.vertices += vertices;
+	counters.current_stats.drawcalls += 1;
```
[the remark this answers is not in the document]
Thalixte: Yes, that's what i did at first, but i thought you would prefer not to call the GetViewport method on each draw calls.

source/d3d9/buffer_detection.cpp
Thalixte: Yes, it i necessary in some games, for instance Oblivion.

Thalixte: I mean, with a step by step debug, it goes into this condition.

source/d3d9/buffer_detection.hpp
Thalixte: OK. Done.

source/d3d9/buffer_detection.cpp
Thalixte: Yes, but as i call it twice, i have prefered to put this on a specific method. And i found it more clear, because, that's exactly what it does, it switches to another dethsurface each time it is called.

source/d3d9/buffer_detection.cpp
```diff
@@ -15,24 +15,41 @@ void reshade::d3d9::buffer_detection::reset(bool release_resources)
...
 #if RESHADE_DEPTH
+	_first_empty_stats = true;
 	_counters_per_used_depth_surface.clear();
 
 	if (release_resources)
 	{
+		_previous_stats = { 0, 0 };
 		update_depthstencil_replacement(nullptr);
+		_depthstencil_replacement.reset();
+		_depthstencil_original.reset();
```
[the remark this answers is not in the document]
Thalixte: OK. Done.

crosire:
I have taken your changes and have changed it so that instead of separate `_depthstencil_replacement` and `_preserved_depthstencil_surfaces` fields, there is only a single vector that manages both (since IMHO it was confusing to keep track of where live references where held before). Pushed that to the https://github.com/crosire/reshade/tree/d3d9_buffer_detection branch. It should still work the same way your changes did otherwise, but would be great if you could test against your games (I tried in Mirror's Edge and it seems to be working).

source/d3d9/buffer_detection.cpp
crosire: I just realized that this does not actually make sense in D3D9, since the replacement surfaces are always cleared at the end of the frame (in `buffer_detection::reset`), so the only time this can be triggered by the game is if it clears a surface at the start of the frame before drawing anything. And that is not a case we want to capture, so this actually hurts more than it helps.

source/d3d9/buffer_detection.cpp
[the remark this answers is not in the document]
crosire: I would have normally, but in this case it should only happen when `preserve_depth_buffers` is true, so is not always happening.

Thalixte:
> I have taken your changes and have changed it so that instead of separate `_depthstencil_replacement` and `_preserved_depthstencil_surfaces` fields, there is only a single vector that manages both (since IMHO it was confusing to keep track of where live references where held before). Pushed that to the https://github.com/crosire/reshade/tree/d3d9_buffer_detection branch. It should still work the same way your changes did otherwise, but would be great if you could test against your games (I tried in Mirror's Edge and it seems to be working).

Nice, thks ;). I tested on Oblivion, Batman Arkham Asylum and RE4 HD Ultimate, and it works stil the same. Unfortunately, i cannot test on more games for the moment (i like to test on Half Life  2, Outlast, Mass Effect 2, and Race Driver Grid).

Thalixte:
OK. tested On Half Life 2, Outlast, Mass Effect 2, and Race Driver Grid. Works as intended ;-). The flickering issue, as well as the viewport issue are fixed now.

crosire:
Awesome, thank you!

PR #179, opened 2020-07-30
Adds a checkbox in Settings for `Do not compile effects automatically`

kimberly990:


crosire:
That option is not in the UI intentionally actually. It's only for use in exceptional situations.

kimberly990:
Since I've compiled a custom build with this option, I've been using it constantly. It's been helpful in nearly _every_ situation, and saves a lot of time and frustration. Please do consider adding it to the UI.

RedDevilus:
I see the potential in this for lower-end machines but also for general usage for all machines, i like to fiddle with settings alot so it would be a nice QoL addition. If it's already an option why not bring it to the general users?
Can i ask what these exceptional situations are?

PR #180, opened 2020-09-12
Uitweaks

luluco250:
I've made two changes to the UI/UX:

- Added a performance mode toggle key.
  - This was an idea TreyM suggested in the Discord code chat, I thought it was neat and pretty simple to implement.
- Made it so that new effect/texture paths added with the "alt-click" copy the last entry if the list isn't empty.
  - I did this so it'd be faster to add a ton of repositories in the same location. For example, if I have all my repos at `C:\Projects\ReShade\`,  I can add an entry to `C:\Projects\ReShade\reshade-shaders\Shaders`, duplicate it a few times and then change the names, much faster than having to copy and then paste each entry individually.


crosire:
I've merged the path list change already. Not sure about the performance mode button yet, will have to think about that some more.

PR #181, opened 2020-09-12
Fixed hook queuing flag

LukeRoss00:
`hook::apply_queued_actions()` was getting called immediately for `queue_enable == true` (probably not important, but it does cause every thread in the process to be suspended every time a single hook is installed)

### DIFF
--- a/source/hook_manager.cpp
+++ b/source/hook_manager.cpp
@@ -354,7 +354,7 @@ bool reshade::hooks::install(const char *name, hook::address target, hook::addre
 	hook.replacement = replacement;
 
 	return install_internal(name, hook, hook_method::function_hook) &&
-		(!queue_enable || hook::apply_queued_actions()); // Can optionally only queue up the hooks instead of installing them right away
+		(queue_enable || hook::apply_queued_actions()); // Can optionally only queue up the hooks instead of installing them right away
 }
 bool reshade::hooks::install(const char *name, hook::address vtable[], unsigned int offset, hook::address replacement)
 {


crosire:
The behavior was correct: When `queue_enable` is `true`, install requests are only queued up and the caller is expected to call `apply_queued_actions` after that is done. When `queue_enable` is `false` the hook is to be installed immediately (and not just queued up), so `apply_queued_actions` has to be called immediately after enqueuing. This change would flip that around and break the code, causing hooks to never get installed at all.
The current code will suspend threads, yes, which is why it's encouraged to call `install` with `queue_enable` set to `true` in case multiple hooks have to be installed at the same time (as is e.g. the case with the OpenGL ones). That is already happening though.

LukeRoss00:
I may be missing something, but the original code (before my change) is:

```
return install_internal(name, hook, hook_method::function_hook) &&
		(!queue_enable || hook::apply_queued_actions());
```

When `queue_enable` is `true`, the first operand of the OR evaluates to `false`, which makes the code evaluate the second operand, so `hook::apply_queued_actions()` is called immediately. Unless I'm misunderstanding your intention, this should not happen, because when `queue_enable` is `true` the caller is expected instead to call `hook::apply_queued_actions()` explicitly at the end, after all requests are batched.

crosire:
I looked at that piece of code multiple times while checking your changes and it still looked correct to me. Guess third times the charm, since I can finally see the problem now. Sorry about that ...

LukeRoss00:
Haha, no problem, I know the feeling! It's a very minor fix anyway

crosire:
Great catch by the way! Nice to see somebody digging into the gritty details at the heart of ReShade.

LukeRoss00:
:-) Lately I've been interested in Reshade for a couple of projects. The first relates to my VR [mod](https://github.com/LukeRoss00/gta5-real-mod) for GTA V. Several people with insanely powerful systems would like to combine it with graphics overhaul mods that are based around ReShade, but the hitch is that ReShade presets are applied to the final backbuffer that gets presented on the screen, while for various reasons my mod works on an internal, higher resolution buffer that GTA V generates but which does not belong to the swapchain. So I would like to see if I can make ReShade work on this higher res buffer instead of the game's backbuffer.

The second reason (why I was single-stepping through your hooking code) is that my GTA V mod was based on 3Dmigoto which is DX11-only, but newer games like Red Dead Redemption 2 and others use DX12 or Vulkan. So I need a new modding tool :-D

PR #182, opened 2020-09-20
Add option 'Load only enabled effects'

seri14:
loads_only_enabled_effects


source/runtime_gui.cpp
```diff
@@ -1799,10 +1807,12 @@ void reshade::runtime::draw_preset_explorer()
...
 
 	if (ImGui::ButtonEx("<", ImVec2(button_size, 0), button_flags))
-		switch_to_next_preset(_current_browse_path, true);
+		if (switch_to_next_preset(_current_browse_path, true))
+			reload_preset = true;
 	ImGui::SameLine(0, button_spacing);
 	if (ImGui::ButtonEx(">", ImVec2(button_size, 0), button_flags))
-		switch_to_next_preset(_current_browse_path, false);
+		if (switch_to_next_preset(_current_browse_path, false))
+			reload_preset = true;
```
seri14, line 1815: This is just a bug.

source/runtime.cpp
```diff
@@ -243,9 +243,30 @@ void reshade::runtime::on_present()
...
+	else if (std::vector<std::string> techniques; !preset.get({}, "Techniques", techniques))
+		effect.skipped = false;
+	else if (std::find_if(techniques.cbegin(), techniques.cend(), [&effect_name](const std::string &technique) {
+		const size_t at_pos = technique.find('@') + 1;
+		return at_pos == 0 || technique.find(effect_name, at_pos) == at_pos; }) != techniques.cend())
+		effect.skipped = false;
+	else
+		effect.skipped = true;
+
+	if (effect.skipped)
```
crosire, line 265: I think doing this check inside `load_effect` could be a problem when editing an effect in the editor, since this check would trigger if saving there too. But if I save an effect in the editor I would expect it to get loaded, even if the preset has it not active.
My suggestion would be to do this check in the loop in `load_effects` which calls `load_effect` instead. That way disabled effects still are not loaded by default, but will be loaded when any other part of the code (like the editor) calls `load_effect`.

source/runtime.cpp
```diff
@@ -243,9 +243,30 @@ void reshade::runtime::on_present()
...
+
 	effect &effect = _effects[index]; // Safe to access this multi-threaded, since this is the only call working on this effect
+	effect = {};
 	effect.source_file = path;
-	effect.compile_sucess = true;
+	effect.compiled = true;
+
+	if (!_effect_load_skipping || _load_option_disable_skipping)
+		effect.skipped = false;
+	else if (std::vector<std::string> techniques; !preset.get({}, "Techniques", techniques))
```
crosire, line 256: I would consider doing a quick check to see if there is any `@` symbol in the technique list. If not, enable all effects, since the preset is using the old format. If yes, do the skip check.

source/runtime_gui.cpp
```diff
@@ -2528,6 +2538,12 @@ void reshade::runtime::draw_variable_editor()
 }
 void reshade::runtime::draw_technique_editor()
 {
+	bool reload_required = false;
+
+	if (_effect_load_skipping_ui)
+		if (size_t skipped_effects = std::count_if(_effects.begin(), _effects.end(), [](const effect &effect) { return effect.skipped; }); skipped_effects > 0)
+			reload_required = ImGui::ButtonEx(std::format("Force load all effects (%lu remaining)", skipped_effects).c_str(), ImVec2(ImGui::GetWindowContentRegionWidth(), 0));
```
crosire, line 2545: Could the `format` call here be replaced with an inline call to either `sprintf_s` or `ImFormatString`? I would like to avoid adding a new header (std_string_ext.hpp).

crosire (commented):
Overall concept looks great, just some small items to discuss further:

source/runtime_gui.cpp
```diff
@@ -1108,6 +1109,13 @@ void reshade::runtime::draw_ui_settings()
 			_tutorial_index = 0;
 	}
 
+	if (ImGui::CollapsingHeader("Effects & Techniques", ImGuiTreeNodeFlags_DefaultOpen))
```
crosire, line 1112: I'd move this in the `General` section, since the relevant settings are stored in the `GENERAL` section of the INI and I have started to try and map INI sections to sections on the Settings page more (hence why there is now a `OVERLAY` and `SCREENSHOTS` section in the INI).
Even better yet would be to move the `Show \"Force load all effects\" button` into the `Overlay` section while at it.

source/runtime.cpp
seri14:
```cpp
else if (std::find_if(techniques.cbegin(), techniques.cend(), [&effect_name](const std::string &technique) {
		const size_t at_pos = technique.find('@') + 1;
		return at_pos == 0 || technique.find(effect_name, at_pos) == at_pos; }) != techniques.cend())
```

`at_pos` is overflow to 0 when `std::string::npos`. this is for no `@` symbol pattern.
so `skipped` is false.

source/runtime.cpp
seri14: See https://github.com/crosire/reshade/pull/182/commits/1e7647dbb271a12067faa135ca652d1f7c5565c4

source/runtime_gui.cpp
seri14: loads_only_enabled_effects2
I do not like this. :thinking:

source/runtime.cpp
crosire: :+1:

source/runtime_gui.cpp
crosire: Gotta say I do :stuck_out_tongue:. The "Load only enabled effects" checkbox is now closer to the effect search path setting, which is a nice corelation and IMHO it makes sense to have any settings that affect the overlay to be in the overlay section (similar to the "Group effect files ..." one).

seri14: Sorry I can't found good position.
If another changes has no problem, I recommend to change this by yourself. 🙏

PR #183, opened 2020-09-20
Use placeholder for search

seri14:
Use ImGui feature instead of user code


crosire:
Nice! Had not realized ImGui can do this now. I just restored the clear button, since I find it can be useful sometimes.

PR #184, opened 2020-09-20
Add config 'ReShadeContainerPath'

seri14:
Change usage the API-named configuration file. (`d3d9.ini`, `d3d10.ini`, `dxgi.ini`, `d3d11.ini` ...)
There are changed to managing folder-unique settings.

**New settings**
```
[INSTALL]
ReShadeContainerPath=(Root folder to resolving when ReShade handling pathes e.g. presets/shaders/textures/.log)
```


seri14:
I am concidering about end support this case.
https://github.com/crosire/reshade/pull/145#discussion_r327587630

crosire:
Neat! Only issue I have is that I would prefer keeping the code in dll_main.cpp instead of adding a new dll_settings.cpp. If you don't mind I can change this and push to your branch.

seri14:
Sure

crosire:
I renamed the `ReShadeContainerPath` setting to `BasePath`, since that communicates what it does more clearly I think,

PR #186, opened 2020-09-21
Change the type of variable used to manage shared textures

seri14:
The second commit is optional. User can check who read and wrote it.

texture_shared


crosire:
I like the idea. I do think that using `std::unordered_set` is overkill here though, considering that it's unlikely for there to be more than two shared instances (which means iterative loopup will be faster than hashing). So I'd consider replacing that with a `std::vector` and just using plain old `std::find` to check if an effect index is in there.

PR #189, opened 2020-09-22
Modify default values of search paths

seri14:
for reset the setting by manually


crosire:
Closing this for the time being, since it will be part of the caching pull request (and I would actually rather keep the default search paths are they are and only have the setup tool change them).

PR #190, opened 2020-09-22
Add cache for effect

seri14:
This is the revenge of #154 
This PR contains #189


source/d3d11/runtime_d3d11.cpp
```diff
@@ -443,7 +446,7 @@ bool reshade::d3d11::runtime_d3d11::init_effect(size_t index)
 			if (_renderer_id < D3D_FEATURE_LEVEL_11_0)
 			{
 				effect.errors += "Compute shaders are not supported in ";
-				effect.errors += "D3D10";
+				effect.errors += "D3D11";
```
seri14, line 449: Found a bug

seri14:
Create the folder `(DLLfolder)\reshade-shaders\Intermediate` before review.

I will push huge commit first. Code correction is after that.

source/d3d11/runtime_d3d11.cpp
crosire: That is actually intentional, see the feature level check above. Compute shaders are only fully supported in feature level 11.0+, but not in feature level 10, hence why it says they are not supported in D3D10 (it's running D3D10 feature level under the D3D11 API).

seri14:
The `effect.restored` is same feature of definition editor error recovation

source/effect_module.hpp
```diff
@@ -193,8 +193,8 @@ namespace reshadefx
 		texture_address_mode address_u = texture_address_mode::clamp;
 		texture_address_mode address_v = texture_address_mode::clamp;
 		texture_address_mode address_w = texture_address_mode::clamp;
-		float min_lod = -3.402823466e+38f; // FLT_MAX
-		float max_lod = +3.402823466e+38f;
+		float min_lod = std::numeric_limits<float>::lowest();
+		float max_lod = std::numeric_limits<float>::max();
```
crosire, line 197: This cannot be used here, since the `<limit>` header is not included at this point (and it seems unnecessary to include it just for this). This compiles with MSVC, because the STL in MSVC includes `<limit>` in `<vector>`, but it does not compile on Clang/GCC. All the `effect_*` files need to retain Linux compatibility (for VkBasalt and others), so this would break that.

source/effect_preprocessor.hpp
```diff
@@ -144,8 +151,8 @@ namespace reshadefx
 		size_t _current_input_index = 0;
 		unsigned short _recursion_count = 0;
 		location _output_location;
-		std::unordered_set<std::string> _used_macros;
-		std::unordered_map<std::string, macro> _macros;
+		std::map<std::string, location> _used_macros;
+		std::map<std::string, macro> _macros;
```
crosire, line 155: Is there a reason these were turned into a `std::map`, rather than `std::unordered_map`?

source/runtime.cpp
```diff
@@ -601,40 +672,51 @@ void reshade::runtime::load_effects()
 #endif
 	_last_shader_reload_successful = true;
 
+	// Copy is required because load_effect is runs asynchronously.
+	ini_file &preset = ini_file::load_cache(_current_preset_path);
```
crosire, line 676: Comment says a copy is required, but a reference is used?

source/runtime_gui.cpp
```diff
@@ -2444,21 +2446,65 @@ void reshade::runtime::draw_variable_editor()
...
-			for (float x = 0, space_x = ImGui::CalcTextSize(" ").x, width = (ImGui::CalcItemWidth() - ImGui::CalcTextSize(category_label.data()).x - 45) / 2; x < width; x += space_x)
+			for (float x = 0, space_x = ImGui::CalcTextSize(" ").x, width = (ImGui::CalcItemWidth() - ImGui::CalcTextSize(category_label.c_str()).x - 45) / 2; x < width; x += space_x)
 				category_label.insert(0, " ");
 
-		if (!_effects[effect_index].definitions.empty() &&
-			ImGui::TreeNodeEx(category_label.c_str(), ImGuiTreeNodeFlags_DefaultOpen | ImGuiTreeNodeFlags_NoTreePushOnOpen))
+		if (ImGui::TreeNodeEx(category_label.c_str(), ImGuiTreeNodeFlags_NoTreePushOnOpen))
 		{
-			for (const std::pair<std::string, std::string> &definition : _effects[effect_index].definitions)
+			if (!effect.preprocessed)
```
crosire, line 2454: I think this whole block of code can be replaced with this:
```c++
if (!effect.preprocessed)
    reload_effect = true;
```
This would avoid the code duplication here. The list will be empty for a single frame, but I think that is fine.

source/effect_preprocessor.hpp
```diff
@@ -85,7 +92,7 @@ namespace reshadefx
 		/// Get a list of all defines that were used in #ifdef and #ifndef lines
 		/// </summary>
 		/// <returns></returns>
-		std::vector<std::pair<std::string, std::string>> used_macro_definitions() const;
+		std::vector<std::pair<std::string, macro_detection_info>> used_macro_definitions() const;
```
crosire, line 95: It looks like the additional information returned here is not actually used anywhere currently, so this change is probably not necessary (yet).

source/runtime.hpp
```diff
@@ -221,10 +225,14 @@ namespace reshade
...
 		unsigned int _drawcalls = 0;
 
-		std::vector<effect> _effects;
+		std::vector<effect> _effects, _previous_effects;
+		std::unordered_map<size_t, effect> _recent_effects;
 		std::vector<texture> _textures;
 		std::vector<technique> _techniques;
 
+		// === Runtime Description ===
+		std::string _renderer_name;
```
crosire, line 234: I'd just use the `_renderer_id` for the cache file names instead of this, to avoid the new field. The cache file names are already not very humand-readable anyway (because of the hash).

crosire (commented):
As discussed previously I think I'll hold off from merging this before the 4.8 release, but I've left some comments below already.

source/effect_preprocessor.hpp
seri14: I prefer to sort the keys (definition names) in definition editor.
But well, following code might be better.
```cpp
// Keep track of used preprocessor definitions (so they can be displayed in the GUI)
effect.definitions.clear();

for (const auto &definition : pp.used_macro_definitions())
{
	if (definition.first.size() <= 4 || ... )
		continue;

	effect.definitions.push_back( ... );
}

std::sort(effect.definitions.begin(), ... ); // ** Add **

// Keep track of included files
effect.included_files = pp.included_files();
```

source/runtime.cpp
seri14: Sorry, this comment is wrong now.

SkonteJonte-deffpkr1234:
Hi, I'm completely new to github, but I would like to help with this fascinating project. I have been using ReShade for maybe a week now and making preset mods for ReShade on nexus mods. Is there any way I could help, or a more appropriate place to ask these questions? Again, totally new to github haha

source/runtime.hpp
SkonteJonte-deffpkr1234: Don't know if you saw my comment "Hi, I'm completely new to github, but I would like to help with this fascinating project. I have been using ReShade for maybe a week now and making preset mods for ReShade on nexus mods. Is there any way I could help, or a more appropriate place to ask these questions? Again, totally new to github haha"

crosire:
Not sure why GitHub still shows there being merge conflicts, since those should have been solved by https://github.com/seri14/reshade/commit/25de5196c4cf4b8eaabe82a4c27e4bbaafdb27ed ... :confused:

seri14:
Status updates is suppressed until the progress. Bad behavior.

PR #192, opened 2020-09-26
Modify time format in the statistics tab

seri14:
format_time

### DIFF
--- a/source/runtime_gui.cpp
+++ b/source/runtime_gui.cpp
@@ -619,11 +619,7 @@ void reshade::runtime::draw_ui()
 
 		if (_show_clock)
 		{
-			const int hour = _date[3] / 3600;
-			const int minute = (_date[3] - hour * 3600) / 60;
-			const int seconds = _date[3] - hour * 3600 - minute * 60;
-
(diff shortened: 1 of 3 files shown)


source/runtime_gui.cpp
```diff
@@ -1357,7 +1349,7 @@ void reshade::runtime::draw_ui_statistics()
 		else
 			ImGui::TextUnformatted("Unknown");
 		ImGui::TextUnformatted(g_target_executable_path.filename().u8string().c_str());
-		ImGui::Text("%d-%d-%d %d", _date[0], _date[1], _date[2], _date[3]);
+		ImGui::Text("%.4d-%.2d-%.2d %.2d:%.2d:%.2d", _date[0], _date[1], _date[2], _date[4], _date[5], _date[6]);
```
crosire, line 1352: The reason this was shown like this is to make it simple for developers to see the value of the `source = "date"` uniform in real time, since the statistics page is primarily targeted at developers. If one wants to see the time in a human readable format, the clock overlay is the choice (Settings > Show clock). So I don't think changing this would be benefitial.

seri14: How about "%.0f ms" text ?

seri14: Ah, I see

PR #194, opened 2020-09-30
Use full CPU time in each thread when loading effects

seri14:


crosire:
This was partially merged in 7e19e5470cb2679c0838d80745b1e529bf16aadd. The rest (using a queue for reload tracking) I think is not worth the added code complexity, considering that the parser is fast enough that these threads don't generally run longer than a second anyway. Closing this for now.

PR #196, opened 2020-10-04
Automatically disable effects when stall in rendering techniques

seri14:


crosire:
I'm not so sure that this is a good idea. There are screenshot artists which will use a ton of effects or use certain special effects which are known to be extremly slow (e.g. FFT) to make a shot look beautiful, even though this may tank FPS down to being unusable normally. But that's fine, since it is only for the purpose of the screenshot. That would break if ReShade automatically disables effects at such low framerates though.

PR #197, opened 2020-10-09
Update opengl_hooks.cpp

qwezzzz1:


crosire:
But why?

crosire:
Closing, since no explanation was provided.

PR #200, opened 2021-01-04
Application crosstalk feature

megai2:
Hi. I want to add an ability to supply data to reshade from games/mods.

For example supplying depth buffer without heuristics or 3d scene without UI
by reshade independent code i.e. without specific forks.
(real story behind that is various requests from Guild Wars 2 community to make a modern
version of https://github.com/04348/Gw2Hook and also some other requests)

I did make some hacky implementation for dx12, that uses command list SetPrivateData
function to supply needed data using magic guid and data checks.
Supplied data is later used on effect reload to replace texture resources with specific names.

General idea works, but it will be nice to hear
some comments about that, before making more changes.

### DIFF
--- a/source/d3d12/d3d12_command_list.cpp
+++ b/source/d3d12/d3d12_command_list.cpp
@@ -6,6 +6,7 @@
 #include "dll_log.hpp"
 #include "d3d12_device.hpp"
 #include "d3d12_command_list.hpp"
+#include "runtime_d3d12.hpp"
 
 D3D12GraphicsCommandList::D3D12GraphicsCommandList(D3D12Device *device, ID3D12GraphicsCommandList *original) :
 	_orig(original),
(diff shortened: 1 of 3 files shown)


crosire:
I'm actually currently working on a add-on API for ReShade, which would allow other mods to communicate with ReShade and even change game's rendering behavior through ReShade. This has not been pushed to GitHub yet though, but can do so soon for feedback.

The "official" solution to provide custom textures to shaders is using semantics: `texture fooBar : MY_SPECIAL_TEXTURE`, so I'll work on making sure the API has means to set those, so that a GW2 mod could provide ReShade with the right resources and have shaders access them via e.g.
```
texture zprepass : ZPREPASS;
texture gbuf_0 : GBUF_0;
texture gbuf_1 : GBUF_1;
texture gbuf_2 : GBUF_2;
texture gbuf_3 : GBUF_3;
texture gbuf_4 : GBUF_4;
texture overlay_0 : OVERLAY_0;
texture overlay_1 : OVERLAY_1;
...
```

megai2:
Good to hear that. 
So for now i can replace
```
		if (texture.unique_name == "V__crosstalk_color")
			resource = crosstalk::get_crosstalk_resource(crosstalk::ResNames::COLOR);
```
lines with 
```
		if (texture.semantic == "CROSSTALK_COLOR")
			resource = crosstalk::get_crosstalk_resource(crosstalk::ResNames::COLOR);
```

And this will work with following up API, without changes in shaders?

crosire:
Probably, yes.

jiangyi0923:
Can you open your modified reshade source code so that I can do Chinese localization?

PR #201, opened 2021-01-07
DX11/DX12:  Implemented depthstencil sorting

FransBouma:
Implemented depthstencil sorting based on the # of times the depth stencil was shown in the debug menu

This greatly helps with keeping the list steady even if there are temporary buffers created per frame. Sorting is per display count then on resolution and then on texture address.

Implemented for DX11 and DX12 for now. I tried to implement it generically in runtime, but I kept running into linker issues with the templated function, so in the end I gave up and implemented it for dx11 and dx12 in their state trackers. 
I hope I have named things properly and didn't make any big mistakes. Tested on CP2077 (DX12) and ROTTR (dx11)

### DIFF
--- a/source/d3d11/state_tracking.cpp
+++ b/source/d3d11/state_tracking.cpp
@@ -80,17 +80,6 @@ void reshade::d3d11::state_tracking::merge(const state_tracking &source)
 
 		target_snapshot.clears.insert(target_snapshot.clears.end(), snapshot.clears.begin(), snapshot.clears.end());
 	}
-	for (const auto& [depthstencil_address, shown_count] : source._shown_count_per_depthstencil_address)
-	{
-		if (_shown_count_per_depthstencil_address.count(depthstencil_address) > 0)
-		{
-			_shown_count_per_depthstencil_address[depthstencil_address] = shown_count + _shown_count_per_depthstencil_address[depthstencil_address];
-		}
-		else
-		{
-			_shown_count_per_depthstencil_address[depthstencil_address] = shown_count;
-		}
-	}
 #endif
 }
 
@@ -219,7 +208,7 @@ bool reshade::d3d11::state_tracking_context::update_depthstencil_clear_texture(D
 	return true;
 }
 
-std::vector<std::pair<ID3D11Texture2D*, reshade::d3d11::state_tracking::depthstencil_info>> reshade::d3d11::state_tracking::sorted_counters_per_used_depthstencil()
+std::vector<std::pair<ID3D11Texture2D*, reshade::d3d11::state_tracking::depthstencil_info>> reshade::d3d11::state_tracking_context::sorted_counters_per_used_depthstencil()
 {
 	struct pair_wrapper
 	{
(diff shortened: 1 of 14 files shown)


FransBouma:
Sorry for the proj changes / sdk upgrades. If you want me to remove these I'll do that. I didn't notice these crept into the PR...

source/d3d11/state_tracking.hpp
```diff
@@ -37,6 +37,7 @@ namespace reshade::d3d11
 #if RESHADE_DEPTH
 		void on_set_render_targets();
 		void on_clear_depthstencil(UINT clear_flags, ID3D11DepthStencilView *dsv, bool rect_draw_call = false);
+		std::vector<std::pair<ID3D11Texture2D*, depthstencil_info>> sorted_counters_per_used_depthstencil();
```
crosire, line 40: This should be part of `state_tracking_context` instead of `state_tracking`, since we don't want a separate instance in every device context and command list, but rather only one. Same thing applies to the `_shown_count_per_depthstencil_address` field (and to D3D12 too). Don't need to bother with merging then either.

source/d3d11/state_tracking.hpp
```diff
@@ -49,6 +50,11 @@ namespace reshade::d3d11
 		bool _has_indirect_drawcalls = false;
 		std::unordered_map<com_ptr<ID3D11Texture2D>, depthstencil_info> _counters_per_used_depth_texture;
 #endif
+
+	private:
+#if RESHADE_DEPTH
+		std::unordered_map<size_t, int> _shown_count_per_depthstencil_address;
```
crosire, line 56: Would be easier to make this a `std::unordered_map<ID3D11Resource *, unsigned int>`, to avoid the ugly address casts. Without the use of `com_ptr` this won't reference count the object, so is otherwise identical. Same applies to D3D12 too.

source/d3d11/state_tracking.cpp
```diff
@@ -208,6 +219,60 @@ bool reshade::d3d11::state_tracking_context::update_depthstencil_clear_texture(D
...
+
+	for (const auto& [dsv_texture, snapshot] : _counters_per_used_depth_texture)
+	{
+		auto dsv_texture_instance = dsv_texture.get();
+		D3D11_TEXTURE2D_DESC desc;
+		dsv_texture->GetDesc(&desc);
+		
+		size_t texture_address = *((size_t*)(&dsv_texture_instance));
+		// get the display counter, if any of this texture.
+		if (_shown_count_per_depthstencil_address.count(texture_address) > 0)
```
crosire, line 244: Would be better to use `.find() != .end()` here, since `.count()` doesn't cancel the lookup after the first find on `unordered_map`, at least on VC++ STL last I checked (I know that doesn't make much sense, considering that a map can only ever have one or zero entries for a key, but it is what it is).

crosire (commented):
That would be wonderful, yes! Overall concept looks good, just have some requests on improving/simplifying the implementation:

source/d3d11/state_tracking.hpp
FransBouma: Done

source/d3d11/state_tracking.hpp
FransBouma: Good point. I feared it might refer to a texture that might have been removed and it would crash but we don't use the actual pointer in that fashion so that's indeed doable. 
Changed.

source/d3d11/state_tracking.cpp
FransBouma: Changed. :)

FransBouma:
All changes made. :)

PR #202, opened 2021-02-20
PoC: OpenVR support

fholger:
The VR community is in desparate need for a tool like Reshade to improve blurry visuals or poor color utilisation on LCD displays. So I thought I'd give it a shot to see how hard it would be to integrate support for VR titles into Reshade :) This is the minimal set of changes I could come up with to apply a Reshade effects toolchain to D3D11 games using OpenVR.

I do not expect this PR to be merged as-is, there is definitely some work to be done before it could be deemed production-ready. I rather wanted to start a discussion with you early and get your feedback, specifically if you are even interested in VR support to begin with :)

There is an interesting architectural challenge with this, as OpenVR support is not actually a new, separate runtime, but rather an extension to the existing graphics runtimes. In particular, it replaces the on_present event for a window's backbuffer with eye texture submissions. Unfortunately, the runtimes are currently fairly tightly coupled with the concept of a window, backbuffer and input. For this first proof of concept, I hacked into the D3D11 runtime and provided it with an alternative constructor/on_init function that does not require a swapchain and instead creates a simple texture as a backbuffer. This allows me to reuse the effect chain application without too much work. A more sophisticated solution might require to decouple the effect chain from the runtime to make it reusable on its own, but I'm not familiar enough with the code to provide a good proposal for this.

As it stands, my current approach will make games create two runtimes, one managing the window's backbuffer, and one managing the OpenVR pipeline. The first one is useful for configuring Reshade via its UI, but it does waste some frametime during normal gameplay. The VR runtime skips some features like UI and input support and only applies the effect chain to the eye textures. One drawback currently is that the two runtimes are not synchronized; if you use the Reshade UI to edit the effect chain, you'll have to restart the game before it is applied to the VR runtime.

Aside from that, I've successfully tested and used it with a few games, most notably Skyrim VR and Fallout 4 VR. Other D3D11 games may work, but I've also seen some crash or produce garbage output. So there's definitely some need for further debugging :)

Other needed improvements:
* support for other graphics APIs besides D3D11
* depth texture support (if games even submit it to the OpenVR compositor, which not many games appear to do)
* better error handling
* possibly code style adjustments

I'll happily continue to work on this, if you want me to, to get it into a state you'd be comfortable merging. If not, that's also fine; it was a fun experiment either way. Thanks for this amazing tool :)


crosire:
This is great! OpenVR support is definitely something I want to merge at some point.
I had actually played around with that myself a year ago, but then pretty much forgot about it and didn't push that branch either. Dug it up again in my history and pushed it now in case you want to take a look: https://github.com/crosire/reshade/tree/experimental/openvr_runtime. Concept is pretty similar to what you have.

As for extending the runtimes to work without a swapchain: I think this can be done similar to how `runtime_d3d12` handles things for d3d12on7 support (which has no swapchain too). Constructor still accepts a swapchain pointer, but it is allowed to set that to `nullptr` and the runtime implementation then only executes the code getting data from the swapchain if that is not the case. Plus there is a separate `on_present` overload which manages the backbuffer resources (so basically what you are doing in `openvr_runtime_d3d11::on_submit`).
https://github.com/crosire/reshade/blob/b505e1baf4cbf81ca257b8b0faf49102ffecd854/source/d3d12/runtime_d3d12.cpp#L129
https://github.com/crosire/reshade/blob/b505e1baf4cbf81ca257b8b0faf49102ffecd854/source/d3d12/runtime_d3d12.cpp#L428

I have a Vive myself, so can help with debugging etc., although I'm mostly busy with the addon API (https://github.com/crosire/reshade/tree/feature/addons) at the moment

source/runtime.cpp
```diff
@@ -128,6 +128,28 @@ bool reshade::runtime::on_init(input::window_handle window)
 
 	return true;
 }
+
+bool reshade::runtime::on_init_vr()
```
crosire, line 132: Think it would be easier to keep `on_init` and just add some special handling in there for when no window handle exists (handle is `nullptr`). See (which just sets an empty input object when there is no window):
https://github.com/crosire/reshade/blob/c1de0e14b68ba452fd3b3806f77c0eb503fffd90/source/runtime.cpp#L140
Also need to move the `assert` out of the `input` constructor in that case:
https://github.com/crosire/reshade/blob/c1de0e14b68ba452fd3b3806f77c0eb503fffd90/source/input.cpp#L55

source/openvr/openvr.cpp
```diff
@@ -0,0 +1,46 @@
+#include "dll_log.hpp"
+#include "hook_manager.hpp"
+#include "openvr_runtime_d3d11.hpp"
+
+reshade::openvr::openvr_runtime_d3d11 *vr_runtime_d3d11 = nullptr;
+
+reshade::hook::address *as_vtable(void *instance)
```
crosire, line 7: You can use the `vtable_from_instance` helper function which is already defined in the included `hook_manager.hpp` header instead.

source/dll_main.cpp
```diff
@@ -905,6 +905,9 @@ BOOL APIENTRY DllMain(HMODULE hModule, DWORD fdwReason, LPVOID)
 		reshade::hooks::register_module(get_system_path() / L"opengl32.dll");
 		// Do not register Vulkan hooks, since Vulkan layering mechanism is used instead
 
+		reshade::hooks::register_module( L"openvr_api.dll" );
+		reshade::hooks::register_module( L"openvr_api64.dll" );
```
crosire, line 909: This should probably be something like this to ensure only the DLL of the same bitness is hooked:
```cpp
#  ifdef WIN64
		reshade::hooks::register_module(L"openvr_api64.dll");
#  else
		reshade::hooks::register_module(L"openvr_api.dll");
#  endif
```

source/d3d11/runtime_d3d11.cpp
```diff
@@ -128,7 +157,7 @@ bool reshade::d3d11::runtime_d3d11::on_init()
 	_backbuffer_format = swap_desc.BufferDesc.Format;
 
 	// Get back buffer texture
-	if (FAILED(_swapchain->GetBuffer(0, IID_PPV_ARGS(&_backbuffer))))
+	if (FAILED(swapchain->GetBuffer(0, IID_PPV_ARGS(&_backbuffer))))
```
crosire, line 160: You may want to check out the code around https://github.com/crosire/reshade/blob/b505e1baf4cbf81ca257b8b0faf49102ffecd854/source/d3d12/runtime_d3d12.cpp#L213 on how this could be done without the code duplication in `on_init_vr`.

source/openvr/openvr_runtime_d3d11.cpp
```diff
@@ -0,0 +1,87 @@
...
+			_right_flipped_horizontally = pBounds->uMax < pBounds->uMin;
+			_right_flipped_vertically = pBounds->vMax < pBounds->vMin;
+		}
+
+		// TODO: handle resize if necessary
+
+		if (++_num_submitted == 2)
+		{
+			// both eyes submitted, apply effect chain and submit to original compositor
+			_runtime->on_submit_vr();
```
crosire, line 59: This should likely still call `on_present`, even though there is no input etc. right now. Can also avoid the duplicated code in `on_submit_vr` that way.

crosire: I'd consider creating a `on_present` overload similar to https://github.com/crosire/reshade/blob/b505e1baf4cbf81ca257b8b0faf49102ffecd854/source/d3d12/runtime_d3d12.cpp#L428, which does the things you are currently doing here (creating a backbuffer texture and copying the submitted textures into it). This would avoid some complexity and since it's a child of `runtime_d3d11`, can also remove the `get_backbuffer` function then.

fholger:
> This is great! OpenVR support is definitely something I want to merge at some point.
> I had actually played around with that myself a year ago, but then pretty much forgot about it and didn't push that branch either. Dug it up again in my history and pushed it now in case you want to take a look: https://github.com/crosire/reshade/tree/experimental/openvr_runtime. Concept is pretty similar to what you have.

Awesome, thanks for your comments! I'll have a look at your earlier branch and the D3D12 runtime - (which I skipped so far, as I'm not familiar with D3D12 :)

I'll try to address your points and also get some more playtesting done. Let's see where it leads :)

source/dll_main.cpp
fholger: I can definitely put the openvr_api64.dll behind a guard. However, for 64bit games the OpenVR dll is often just called `openvr_api.dll`, too, so we need to potentially hook this one in either case.

source/openvr/openvr.cpp
fholger: I originally didn't use this function because it has this convenient static assert to ensure it's only called on polymorphic types. However, since I have a void pointer that could represent different versions of the compositor interface, that assert fails, and I didn't want to remove it.

But since I'm only passing the pointer around, I guess I can just cast it to the compositor interface version from the OpenVR header. That would make the assert happy and might even make the code a bit clearer.

source/d3d11/runtime_d3d11.cpp
fholger: I extracted the common code to an `init_resources` function and removed vr specific functions. As you suggested, you can now pass a texture to `on_present` instead, to be used as the backbuffer.

source/runtime.cpp
fholger: Agreed, that's much better. I initially didn't fully understand what implications this would have, so didn't want to touch existing code too much :)

fholger:
Alright, I gave the PR a once-over and tried to follow your suggestion to add an `on_present` specialisation that takes a texture to use as a backbuffer. Code duplications should be gone, and I also improved error handling and cleanup a little. Decided to treat the submitted eye textures separately instead of in one pass, as this allows us to more closely follow the original Submit calls and possibly avoid surprises with fringe use cases.

As far as a minimal integration goes, I'm reasonably happy with this. Please let me know what you think There are definitely some needed quality of life improvements, but I'm going to start working on them in a separate branch to keep this PR manageable:

- changes made in the Reshade UI (handled by the window runtime) are not reflected in the VR runtime until the game is restarted. We'd need some kind of signal to the VR runtime to reload its effects chain
- the window runtime should probably be inactive if VR is detected (unless the UI is opened, to see changes made); otherwise it's mostly a waste of GPU time
- I'm not sure if the statistics you can view in the Reshade UI show global stats or only those from the window runtime. I'm guessing the latter? Would be cool to be able to see the total.
- exposing the depth texture. I originally thought this wouldn't be possible since only a few games submit their depth texture to the compositor, but I've since discovered that Reshade already has clever mechanisms to discover matching depth textures :) Just need to get access to that from the VR runtime.
- and eventually, of course, support for other graphics APIs (or even VR APIs, although the Oculus SDK and OpenXR are probably a little trickier to integrate due to their more complex interface)

crosire:
I merged this against latest `main` branch now, since a lot of code changed as a result of the addon stuff landing there. Also moved the eye texture management into the actual runtime class to avoid the extra runtime class for OpenVR (based on my initial comments). This makes the patch pretty lean, which is a nice bonus. I'll probably start porting some of the additional stuff from the `experimental/openvr_runtime` branch over after merging this too (e.g. support for games that were build against older OpenVR versions).

fholger:
Awesome, thanks! :)

FYI, I've experimented a bit with getting depth texture support working. It's not quite straight-forward, though, because there are 4 different ways that games might be using depth during stereo rendering, and each of those poses a unique challenge:

- a single texture with doubled width containing both eyes. This is what Skyrim and FO4 do and probably the easiest to get access to, as the existing state tracker can already find that texture, in principle.
- a single array texture with one layer for each eye. This is the necessary layout you'd need to use NVIDIA's single-pass stereo rendering techniques. Don't know of any title using it, but I'm sure they are out there.
- two separate textures for each eye. In this case, the challenge is deciding which texture belongs to which eye. In all likelihood, the state tracker probably needs to record a timestamp of the last draw call for each depth texture to separate them, and then there's still an open question which eye was rendered first. I assume most games will render the left eye first, but it's not guaranteed. Would probably need a config option to override, because I don't think Reshade could figure this out automatically.
- a single texture that's used for both eyes and cleared before the second eye is drawn. I've encountered at least one Unity game that does this (actually, it appears to render to two separate MSAA depth textures first, then resolves both of them to this single texture). Could in principle be handled by saving the previous texture on clear, as the state tracker can already do. Would need the same config option to choose which eye came first, though.

Anyway, I'll see if I can put something together to cover all these cases. So if all goes well, I hope to have another PR for you soon :)

crosire:
Just be aware that the state tracking code has been replaced with the add-on API and the code that finds the depth texture now is based on that and lives here: https://github.com/crosire/reshade/blob/main/source/addon/generic_depth.cpp (generalized for all render APIs). This is rather different to how it worked before.

Aberro:
Does it work with ETS2?

kmj007:
Aberro -> Unfortunately it does not work with ETS2 atm.

It works with Dirt Rally 2 and I'm absolutely amazed by how it looks. Sharpening is one thing, obviosly crucial for VR, but color tweaks are also great for LCD Panels.

Aberro:
@kmj007 you just twist the knife, aren't you?)
Well, I guess, I'll just have to wait and hope for the best, and keep my fingers crossed for fix for ETS2.

kmj007:
@Aberro Try renaming dxgi (x64) to dinput8.dll in ETS2 :)

PR #203, opened 2021-03-02
Small VR improvements

fholger:
Here are a few quick changes which help the experience in VR mode:

- ensure the VR runtime reflects changes made to the config or the preset in the window runtime's GUI
- do not apply effects to the window view in VR mode unless the GUI is active. This is to not waste GPU performance in VR, but still allow easy editing of the preset. Although perhaps this should be a config option?
- show statistics from the VR runtime, not the window runtime. I also applied a quick hack to show more "accurate" times seeing that the VR runtime's on_present is called twice per frame. It's admittedly not pretty.

I'm not sure if I found the cleanest way to implement these changes. I'm open to suggestions :)

### DIFF
--- a/ReShade.vcxproj
+++ b/ReShade.vcxproj
@@ -520,6 +520,7 @@
     <ClInclude Include="source\opengl\runtime_gl.hpp" />
     <ClInclude Include="source\opengl\runtime_gl_objects.hpp" />
     <ClInclude Include="source\opengl\state_block_gl.hpp" />
+    <ClInclude Include="source\openvr\reshade_vr.hpp" />
     <ClInclude Include="source\runtime.hpp" />
     <ClInclude Include="source\runtime_objects.hpp" />
     <ClInclude Include="source\vulkan\format_utils.hpp" />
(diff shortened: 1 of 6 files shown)


crosire:
ReShade now renders an overlay with the GUI of the effect runtime associated with VR, which makes it possible to control the effects from within VR: https://github.com/crosire/reshade/commit/e39bf41f42740f98b2a39731a39a9dd368607773
unknown

As a result I don't think it's necessary to couple the VR/non-VR effect runtime anymore, so closing this for now.

fholger:
Yes, that's definitely better. :)

I do think that we still need a way to disable the non-VR runtime in VR. Some VR gamers tend to run a tight fit when it comes to performance and frame times, and the non-VR runtime needlessly eating a few hundred ms matters. Unless that's already adressed? I didn't have the time to check out the latest version :)

crosire:
Good point, that's missing. But this can probably just be a config option, that when set disables effect loading in runtimes that are not `_is_vr == true`.

fholger:
That should be perfectly adequate :) Thanks!

PR #207, opened 2021-04-15
Fix dx12 not allows effect without any texture or sampler

kingeric1992:
Changes root signature to not use sampler or srv/uav when not present.


crosire:
This was potentially setting the wrong root parameters when there e.g. where textures, but no samplers, so fixed that quickly.

PR #209, opened 2021-05-19
update some intrinsic

kingeric1992:
* fix missing mul(scaler, mat2x4) intrinsic overload
* add transpose(non-square-matrix) intrinsics
* add new rule for type ranking, remove compatibility for matrix trunc

btw, I meant to ask is `input.cpp` has the correct encoding? 
git clone always ends up with errors with `u8"OEM ß"` `u8"Ü"` etc.

also `runtime_vk.cpp` is missing reference to `api::format_to_typeless(...)` from `reshade_api_format_tuils.hpp`


crosire:
The `input.cpp` encoding is correct (it's UTF-8): `ß` and `Ü` are actual german characters on the keyboard and are supposed to be there ;)

PR #210, opened 2021-05-20
Avoid processing relative path when it is already abs

Kein:


crosire:
Did you encounter a problem with the way it was before? Technically this shouldn't make a difference, since both `operator/` and `std::filesystem::absolute` will do nothing if the path is already absolute. So I'm curious if there is a specific reason for this, or just code style.

Kein:
When did you last time tried to use absolute path for `IntermediateCachePath`? It always tries to resolve it to relative and then fails because relative Reshade folder + absolute will fail and will fallback to %TEMP%
This is pretty broke logic, especially so because paths saved in config as absolute, which indicates for the editor this how they can be used. Turns out - not.

crosire:
If I set `IntermediateCachePath=C:\Test\Cache`, and ReShade is in `C:\ReShade` and I step through the code I get this:

```g_reshade_base_path``` = `C:\ReShade` (is guaranteed always absolute, see also `get_base_path()` implementation in dll_main.cpp)
```_intermediate_cache_path``` = `C:\Test\Cache`
=> `resolve_path(_intermediate_cache_path)`:
```path``` = `C:\Test\Cache`
```g_reshade_base_path / path``` = `C:\Test\Cache` (that's what I mean, `operator/` simply ignores the first argument if the second is already absolute, see also 1 in https://en.cppreference.com/w/cpp/filesystem/path/append)
```std::filesystem::absolute(g_reshade_base_path / path, ec)``` = `C:\Test\Cache`
```std::filesystem::canonical(path, ec)``` = `C:\Test\Cache`
=> `resolve_path` returns `true` since `ec` is zero (and therefore the fall back to temp directory branch is skipped)
```_intermediate_cache_path``` = `C:\Test\Cache`

So I just don't understand why there would be a problem here.

It is however important to note that `resolve_path` will fail if the specified path (relative or absolute) does not exist (because `std::filesystem::canonical` will fail then). So if you set `IntermediateCachePath` to a directory that has not yet been created, it will fall back to %TEMP%. But that was intended.

Kein:
My path exist, it is ram FAR32 disk, absolute path "N:\cache", it is treated as relative and combined into incorrect one, fails and then being overwritten by temporal.

PR #211, opened 2021-06-10
Additional reshade hook using dinput8.dll

retroluxfilm:
In same cases it is not possible to use the usual dll namings to get reshade loaded. Especially when the loading from the root game folder is bypassed when loaded from other dlls like api-ms-win-core-synch-l1-2-0.dll.

This helps to be able to hook American Truck Simulator & Euro Truck Simulator 2 without the need to use additonal librarys like the asi loader. Without it it would load reshade but the original dinput8.dll library would not be loaded causing not having any HID inputs.

It might not be the best integration and I would be happy to adjust it with your assistance to fit into the code style of reshade.


crosire:
Looking good. Will just quickly do some purely stylistic changes and merge.

PR #215, opened 2021-06-20
Fix the special uniform "mousepoint" returning invalid coordinates

Mortalitas:
Currently, if the mouse leaves the display in multi-display configurations while the hooked application window is active, out of bounds/invalid mouse coordinates will be returned. Feel free to ignore this one if it's intended for shader developers to take advantage of that functionality.

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -2585,7 +2585,10 @@ void reshade::runtime::update_and_render_effects()
 				}
 				case special_uniform::mouse_point:
 				{
-					set_uniform_value(variable, _input->mouse_position_x(), _input->mouse_position_y());
+					const uint32_t posx = _input->mouse_position_x() > _width ? _width : std::max(_input->mouse_position_x(), 0U);
+					const uint32_t posy = _input->mouse_position_y() > _height ? _height : std::max(_input->mouse_position_y(), 0U);
+
+					set_uniform_value(variable, posx, posy);
 					break;
 				}
 				case special_uniform::mouse_delta:


crosire:
Hard to say which behavior is better :upside_down_face:. Do you have an effect that relies on these being clamped or something else that triggered the need for this change? The clamping should use `_window_width` and `_window_height` though, instead of `_width` and `_height`, since mouse coordinates are in window client coordinates and not render resolution coordinates.

Mortalitas:
> Do you have an effect that relies on these being clamped or something else that triggered the need for this change?
I was actually helping @Radegast-FFXIV with some shader behavior troubleshooting as he recently added coordinate input for several of his novelty shaders, and turned it up through that: https://github.com/Radegast-FFXIV/reshade-shaders
> The clamping should use `_window_width` and `_window_height` though, instead of `_width` and `_height`, since mouse coordinates are in window client coordinates and not render resolution coordinates.
Fixed.

kingeric1992:
what about clamping the value in shader?

Radegast-FFXIV:
As crosire mentioned, there's two ways to address where the responsibility
falls: On the shader dev or on ReShade. In my opinion from an application
perspective, it should fall on ReShade to prevent weird states. There are
caveats to this mentioned below.

It's a different story if we're dealing with the cost of performance and
potential side effects in checking mouse coordinates on the ReShade
application layer instead of leaving it to the shader dev to clamp their
values.

kingeric1992:
I mean sure, and now ppl that want off window pos will now run a dedicated pass and tex buffer (or double unorm buffer for precision) to get the pos by accumulating deltas. 

No one will ever need that anyways so I'm sure that's fine. Just a mental note to myself.

crosire:
The imgui mouse cursor always stays inside the window as well, so I suppose it makes sense for `mousepoint` to follow suit.

PR #216, opened 2021-06-26
Added Windows Multimedia Module (winmm.dll) hook 

retroluxfilm:
Added Windows Multimedia Module (winmm.dll) hook to allow to launch reshade for games that are difficult to hook early to catch the rendering api. For example HL Alyx in VR etc. I choose to add this hook as the winmm.dll ist often the most early dll loaded by the game/application which helps to get the hook in the game at the right time. Had a few issues for VR titles and this helps to have universal hook. Still it might cause a bit more cpu cycles due to additional hooks that are actually only needed to load reshade and not to really to intercept the methods.

I imported all methods from the winmm.dll interface and partly generated the exports and the basic method cages. The rest was done by hand using the original interfaces. Not sure if that was required to automatically pass all dlls calls to the original dll or if there was an easier way on how to do it. Let me know if that is ok how it is implemented or if you would like to have it differently


crosire:
That's a bit too many exports for my taste. DirectInput was a nice addition, since it requires so little code, but I'm not convinced it's necessary to add more.
I'm pretty sure I remember having ReShade running in Half Life Alyx before, but will need to install again to check (usually with Source Engine games you have to put the ReShade DLL in the `bin` directory, since the engine does a `SetDllDirectory` to that location and therefore loads DLLs from there).

retroluxfilm:
Yes I was on the edge about this as well. Especially as this is not the core about what reshade does. So I can totally understand if you dont want that in.

On the other hand I only mangaged to get HL Alyx working by either renaming the reshade dll to kernel32.dll (out of luck and possible with side effects ) in the bin folder. Or an alternative was dropping the reshade elements into the SteamVR folder and naming it dxgi.dll but this also caused that SteamHome and possible other games got hooked as well. Not an ideal solution. By using the optional winmm hook it at least ensures to call all original methods and maintaining their functionality. 

Adding a preprocessor to disable the exports could be an compromise?

resistancelion:
> Yes I was on the edge about this as well. Especially as this is not the core about what reshade does. So I can totally understand if you dont want that in.
> 
> On the other hand I only mangaged to get HL Alyx working by either renaming the reshade dll to kernel32.dll (out of luck and possible with side effects ) in the bin folder. Or an alternative was dropping the reshade elements into the SteamVR folder and naming it dxgi.dll but this also caused that SteamHome and possible other games got hooked as well. Not an ideal solution. By using the optional winmm hook it at least ensures to call all original methods and maintaining their functionality.
> 
> Adding a preprocessor to disable the exports could be an compromise?

Why not to load in memory original dll and then replace only required method, with further de-hooking him when not needed?

PR #217, opened 2021-06-30
change input.cpp encoding from ISO-8859 (28591) to utf-8 with BOM

kingeric1992:
Fix compiler error when current user codepage not supporting German characters.
bom for VS to auto detect the file as utf-8.

### DIFF
--- a/source/input.cpp
+++ b/source/input.cpp
@@ -1,4 +1,4 @@
-/*
+ï»¿/*
  * Copyright (C) 2014 Patrick Mours. All rights reserved.
  * License: https://github.com/crosire/reshade#license
  */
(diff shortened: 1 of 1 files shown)


kingeric1992:
As shown in this img, viewing raw `input.cpp` from github with utf8 result in broken German chars.
未命名
looks fine with non CJK or Unicode codepages.
未命名2

crosire:
I really don't like files with BOM, so tried to save it as plain UTF-8, which seems to have worked as well: https://github.com/crosire/reshade/blob/utf8/source/input.cpp. Could you verify?

kingeric1992:
it compiles with warning, but runs fine. So I guess that works.

> warning C4819: The file contains a character that cannot be represented in the current code page (950). Save the file in Unicode format to prevent data loss

I think it's because VS only use utf-8 when BOM is present.

kingeric1992:
Actually no, It compiles, but it doesn't display properly with German Chars.

Ü shown as ? in the UI.

and if I save the file with BOM, it display properly.

PR #218, opened 2021-06-30
Add Gather overload

kingeric1992:
template: `tex2Dgather#(sampler2D s, float2 uv, int2 offset0, int2 offset1, int2 offset2, int2 offset3)`

https://docs.microsoft.com/en-us/windows/win32/direct3dhlsl/t2d-gatherred-s-float-int2-int2-int2-int2-

https://www.khronos.org/registry/OpenGL-Refpages/gl4/html/textureGatherOffsets.xhtml

https://www.khronos.org/registry/spir-v/specs/unified1/SPIRV.html#OpImageGather with `ConstOffsets`

the function provides performance gain similar to `tex2Dgather#(sampler2D s, float2 uv)` but with ability to address texels not immediate neighbor.

### DIFF
--- a/source/effect_symbol_table_intrinsics.inl
+++ b/source/effect_symbol_table_intrinsics.inl
@@ -1732,8 +1732,11 @@ IMPLEMENT_INTRINSIC_SPIRV(tex2Dfetch, 1, {
 
 // ret tex2DgatherR(s, coords)
 // ret tex2DgatherR(s, coords, offset)
+// ret tex2DgatherR(s, coords, offset0, offset1, offset2, offset3)
+
 DEFINE_INTRINSIC(tex2DgatherR, 0, float4, sampler, float2)
 DEFINE_INTRINSIC(tex2DgatherR, 1, float4, sampler, float2, int2)
+DEFINE_INTRINSIC(tex2DgatherR, 2, float4, sampler, float2, int2, int2, int2, int2)
 IMPLEMENT_INTRINSIC_GLSL(tex2DgatherR, 0, {
 	code += "textureGather(" + id_to_name(args[0].base) + ", " +
 		id_to_name(args[1].base) + ", 0)";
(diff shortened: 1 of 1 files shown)


source/effect_symbol_table_intrinsics.inl
```diff
@@ -1801,10 +1826,27 @@ IMPLEMENT_INTRINSIC_SPIRV(tex2DgatherR, 1, {
...
+IMPLEMENT_INTRINSIC_SPIRV(tex2DgatherR, 2, {
+	add_capability(spv::CapabilityImageGatherExtended);
+
+	const spv::Id component = emit_constant(0u);
+	return add_instruction(spv::OpImageGather, convert_type(res_type))
+		.add(args[0].base)
+		.add(args[1].base)
+		.add(component)
+		.add(spv::ImageOperandsConstOffsetsMask)
+		.add(args[2].base)
```
crosire, line 1838: Looks like `ImageOperandsConstOffsetsMask` accepts a single argument, which should be an array, rather than multiple separate arguments (`Offsets must be an <id> of a constant instruction making an array of size four of vectors of two integer components.`):
Quick test with glslang indeed generates this:
```
%35 = OpConstantComposite %_arr_v2int_uint_4 %25 %28 %31 %34
%36 = OpImageGather %v4float %14 %17 %int_3 ConstOffsets %35
```
So I believe this should to the job:
```c++
const spv::Id component = emit_constant(0u);
const spv::Id offsets = add_instruction(spv::OpConstantComposite, convert_type({ reshadefx::type::t_int, 2, 1, 0, 4 }), _types_and_constants)
	.add(args[2].base)
	.add(args[3].base)
	.add(args[4].base)
	.add(args[5].base)
	.result;
return add_instruction(spv::OpImageGather, convert_type(res_type))
	.add(args[0].base)
	.add(args[1].base)
	.add(component)
	.add(spv::ImageOperandsConstOffsetsMask)
	.add(offsets)
	.result;
```

1265578519:
他有希望支持AMD FSR吗？

PR #221, opened 2021-12-31
Convert setup csproj to SDK-style and bump to .NET 4.6.2

luluco250:
Currently the setup's project targets .NET Framework 4.5, however Microsoft has deprecated this version and [removed download links to the required developer pack](https://dotnet.microsoft.com/en-us/download/dotnet-framework/net45).

As a consequence, opening the solution on a fresh VS install will present a warning about the missing devpack, which cannot be downloaded:

image

[They're also going to deprecate versions 4.5.1 through 4.6.1 on April 26, 2022.](https://docs.microsoft.com/en-us/lifecycle/products/microsoft-net-framework)

With that in mind, I believe it's a good idea to bump it to 4.6.2, which does not have a listed end of support date. 4.8 would be nice, but that'd require it to be installed on the user's machine, whereas older versions like 4.6.2 are more widespread. [Current releases of Windows 10 and 11 come with 4.8 already installed though.](https://docs.microsoft.com/en-us/dotnet/framework/migration-guide/versions-and-dependencies#net-framework-48)

I've also converted the project to use the newer [SDK-style](https://docs.microsoft.com/en-us/dotnet/core/project-sdk/overview) which reduces boilerplate a lot and are easier to reason with (such as avoiding git conflicts due to implicit file listings and having a more human-readable csproj file). This also makes it compatible with future versions like .NET 5.0 and 6.0.

It was relatively easy to do the conversion, the only big change being how AssemblyInfo.cs is generated due to [CL not working with SDK-style projects](https://docs.microsoft.com/en-us/dotnet/core/porting/cpp-cli#ccli-net-core-limitations), but it was simple to collect the version information using a PowerShell script to generate the file.

### DIFF
--- a/source/opengl/opengl_hooks.cpp
+++ b/source/opengl/opengl_hooks.cpp
@@ -1553,7 +1553,7 @@ void APIENTRY glLinkProgram(GLuint program)
 				glGetShaderSource(shaders[i], size, nullptr, source.data());
 			}
 
-			reshade::api::pipeline_desc desc = { reshade::api::pipeline_stage::all_shader_stages };
+			reshade::api::pipeline_desc desc = {};
 			for (int i = 0; i < 16 && shaders[i] != GL_NONE; ++i)
 			{
 				GLint type = GL_NONE;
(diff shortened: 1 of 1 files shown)


crosire:
Unfortunately this breaks building the setup tool: Including the individual files in MSBuild is a necessity for the custom `BuildAndAppendSetupArchive` MSBuild target to work (since that gathers all `Archive` MSBuild items and generates the setup archive out of those). Without that the Vulkan layer JSON files are not included like they should. This also removes the post-build event that signs the executable, which is required for official builds (only I can execute that, but I don't want to have to deal with a different project file, hence why this was behind a MSBuild condition).

Btw., executing the `CL` task should work with SDK-style project as well. This does not work with the old project style by default either, hence why the necessary MSBuild props were included first, which make that task available (`<Import Project="$(VCTargetsPath)\Microsoft.Cpp.Common.props" />` and `<UsingTask TaskName="CL" AssemblyFile="$(VCTargetsPath)Microsoft.Build.CppTasks.Common.dll"/>`), but those were removed here hence it not finding the `CL` task.
I don't really see the benefits of the SDK-style projects in general though. The C++ projects are formatted similar to how it was before anyway, so a bit weird to have it different for the one C# project.

I can get onboard updating to .NET 4.6.2 though. I'm hesitant towards updating all the way to 4.8. Tried that once before and very quickly had to revert back due to tons of users having problems even on Win 10 with .NET 4.8 not being installed: https://github.com/crosire/reshade/commit/7b77101e0f1a2643a73d3681cca442d75a909770.

luluco250:
Ah okay, I didn't pay enough attention to the custom target requirement.
I'll just revert the commit and just bump the version number to 4.6.2 then, that's good enough.

That aside, I did try to use `CL`, but it threw up something about cl.exe not being a "rooted path", no mention of it online, so I just assumed it wasn't tested by Microsoft and hence unsupported. Personally I still find it a bit weird to run the preprocessor instead of just matching a line using a script, but if it works it works.

Oh and SDK-style projects are generally just nicer to work with, less configuration, more consistency (no duplicate entries from Visual Studio when dealing with configuration types). Specially when working with big teams, using it + `Directory.Build.props` for shared configurations makes thing very simple in .NET development, which to be fair isn't exactly the use case here.

And about them being inconsistent with C++ projects, well, that's just Microsoft to blame. 🤷‍♂️

crosire:
The rooted path error would indicate that the `ToolPath="$(VC_ExecutablePath_x86)"` property set on the `CL` task is not doing what it should (e.g. `VC_ExecutablePath_x86` is empty). `VC_ExecutablePath_x86` is set in `Microsoft.Cpp.Common.props` though, so that's weird. But may be that with the old style project some other MSBuild property is defined that triggers one of the conditions in `Microsoft.Cpp.Common.props` which the new project style does not.
This is most definetly not tested by MS though (you aren't really supposed to run the C++ compiler in a C# project ...). Just hacked that in since I've been playing around with MSBuild a bit and had done something similar for https://github.com/crosire/scripthookvdotnet before.

And fair enough on the SDK-style projects. Guess I'm just too used to the old stuff :upside_down_face: (I'm not at all up-to-date on all the .NET advances in recent years, like .NET 5 etc.).

PR #223, opened 2022-01-03
Add screenshot file post-processing

seri14:
Use cases:
* Recompression
* Upload to cloud storage

Examples:

.bmp to .png (with ImageMagick)
```ini
[SCREENSHOT]
PostProcessApplication=C:\Softwares\ImageMagick-7.1.0-portable-Q8-x64\convert.exe
PostProcessArguments="<TARGET>" -quality 90 "<DIR>\<STEM>.png"
```

Post-processing by script (with PowerShell)
```ini
[SCREENSHOT]
PostProcessApplication=C:\Program Files\PowerShell\7\pwsh.exe
PostProcessArguments=-File MyScript.ps1 -TargetScreenshotPath "<TARGET>"
```

### DIFF
--- a/source/imgui_widgets.cpp
+++ b/source/imgui_widgets.cpp
@@ -300,7 +300,7 @@ bool reshade::imgui::search_input_box(char *filter, int filter_size, float width
 	return res;
 }
 
-bool reshade::imgui::file_input_box(const char* name, std::filesystem::path& path, std::filesystem::path& dialog_path, const std::vector<std::wstring>& exts)
+bool reshade::imgui::file_input_box(const char *name, std::filesystem::path &path, std::filesystem::path &dialog_path, const std::vector<std::wstring> &exts)
 {
 	return file_input_box(name, nullptr, path, dialog_path, exts);
 }
@@ -349,7 +349,7 @@ bool reshade::imgui::file_input_box(const char *name, const char *hint, std::fil
 
 	return res;
 }
-bool reshade::imgui::directory_input_box(const char* name, std::filesystem::path& path, std::filesystem::path& dialog_path)
+bool reshade::imgui::directory_input_box(const char *name, std::filesystem::path &path, std::filesystem::path &dialog_path)
 {
 	return directory_input_box(name, nullptr, path, dialog_path);
 }
--- a/source/imgui_widgets.cpp
+++ b/source/imgui_widgets.cpp
@@ -300,7 +300,11 @@ bool reshade::imgui::search_input_box(char *filter, int filter_size, float width
 	return res;
 }
 
-bool reshade::imgui::file_input_box(const char *name, std::filesystem::path &path, std::filesystem::path &dialog_path, const std::vector<std::wstring> &exts)
+bool reshade::imgui::file_input_box(const char* name, std::filesystem::path& path, std::filesystem::path& dialog_path, const std::vector<std::wstring>& exts)
+{
+	return file_input_box(name, nullptr, path, dialog_path, exts);
+}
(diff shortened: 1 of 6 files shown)


source/imgui_widgets.cpp
```diff
@@ -345,7 +349,11 @@ bool reshade::imgui::file_input_box(const char *name, std::filesystem::path &pat
 
 	return res;
 }
-bool reshade::imgui::directory_input_box(const char *name, std::filesystem::path &path, std::filesystem::path &dialog_path)
+bool reshade::imgui::directory_input_box(const char* name, std::filesystem::path& path, std::filesystem::path& dialog_path)
```
crosire, line 352: It appears I had a wrong setting in the `.editorconfig` file which messed up the pointer/reference symbol alignment here (and elsewhere). Sorry about that, fixed now in https://github.com/crosire/reshade/commit/101c479057a0417960581e38aaa08ee449cb74e4. Mind adjusting these accordingly?

source/runtime.cpp
```diff
@@ -3355,6 +3365,212 @@ void reshade::runtime::save_screenshot(const std::wstring &postfix)
...
+			std::string_view replacing = ((std::string_view)_postprocess_arguments).substr(brace_begin + 1, brace_end - (brace_begin + 1));
+			size_t colon_pos = replacing.find(":");
+
+			std::string name;
+			if (colon_pos == std::string::npos)
+				name = replacing;
+			else
+				name = replacing.substr(0, colon_pos);
+
+			static const auto stringicmp = [](const std::string_view& a, const std::string_view& b) {
```
crosire, line 3418: runtime.cpp is Windows only anyway, so simpler to just use `_stricmp` directly here (the `stringicmp` lambda is necessary in the effect compiler code only, since that has to support Linux as well, for other users like VkBasalt).

source/runtime.cpp
```diff
@@ -561,6 +562,10 @@ void reshade::runtime::load_config()
 	config.get("SCREENSHOT", "SaveOverlayShot", _screenshot_save_gui);
 	config.get("SCREENSHOT", "SavePath", _screenshot_path);
 	config.get("SCREENSHOT", "SavePresetFile", _screenshot_include_preset);
+	config.get("SCREENSHOT", "PostProcessApplication", _postprocess_application);
+	config.get("SCREENSHOT", "PostProcessArguments", _postprocess_arguments);
```
crosire, line 566: I'd rename these to:
```
[SCREENSHOT]
PostSaveCommand=
PostSaveCommandArguments=
PostSaveCommandWorkingDirectory=
PostSaveCommandShowWindow=
```

source/runtime.hpp
```diff
@@ -82,6 +82,7 @@ namespace reshade
 		/// Captures a screenshot of the current back buffer resource and returns its image data in 32 bits-per-pixel RGBA format.
 		/// </summary>
 		bool capture_screenshot(uint8_t *pixels) final { return get_texture_data(get_back_buffer_resolved(get_current_back_buffer_index()), api::resource_usage::present, pixels); }
+		int execute_screenshot_postprocess(const std::filesystem::path &screenshot_path);
```
crosire, line 85: I think it would make sense to move this into the `private:` section (and have it return `bool` instead of `int`, for consistency with other members of `runtime`).

crosire (commented):
Cool idea. I'm a bit on the fence with calling this "post-processing", due what post-processing otherwise refers to in ReShade (post-processing effects). But am not entirely sure about a better name either. Maybe "post-save command" (to indicate that this is a command that runs after saving)?

PR #225, opened 2022-01-07
Detach the process of saving screenshot from main thread

seri14:
Performance improves (in my PC)

* Taking a screenshot
greater than 2 sec -> less than 1 sec

* How long stopped the main thread
greater than 2 sec -> few frames


crosire:
Rewrote this a bit to make use of the existing worker threads that `runtime` was already managing for effect loading. This way one can now spam click the screenshot save button and it will simply enqueue the save into another thread (save may fail if spamming too much since the filename is distinguished by seconds only, but that's ok).

PR #227, opened 2022-01-08
Create screenshot directory if it doesn't exist.

luluco250:
This PR makes ReShade create the directory where screenshots are to be saved instead of just erroring out.

It is done by trying to create the directory (through `std::filesystem::create_directories()`) and then checking if it exists afterwards with `std::filesystem::exists()`. If it doesn't, it sets `_failed_to_create_screenshot_dir` to `true` and the error message will specify that it could not have been created.

PS: I've tried using the `std::error_code` returned by `create_directories()` but it would seem to not raise an error even when trying to save to a protected directory (like `C:\Program Files\<subdir>`), this seemed like a more reliable solution.

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -3305,13 +3305,22 @@ void reshade::runtime::save_screenshot(const std::wstring &postfix)
 	filename += postfix;
 	filename += _screenshot_format == 0 ? L".bmp" : _screenshot_format == 1 ? L".png" : L".jpg";
 
-	std::filesystem::path screenshot_path = g_reshade_base_path / _screenshot_path / filename;
+	std::filesystem::path screenshot_dir = g_reshade_base_path / _screenshot_path;
+	std::filesystem::path screenshot_path = screenshot_dir / filename;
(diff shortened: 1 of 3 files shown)


luluco250:
Actually, let me rebase this to get rid of the pointless commits.

seri14:
I remember this PR https://github.com/crosire/reshade/pull/145#discussion_r327587630
The C++ standard library does not always best behavior.
Probably the best result (wanted in this PR) is that try to create upper-directories by V-shape order.

#### Order of the `create_directory` challenge
1. `X:\a\b\c`
2. `X:\a\b` (`if ec.value() != 0`)
3. `X:\a` (`if ec.value() != 0`)
4. `X:\a\b` (`if ec.value() == 0`)
5. `X:\a\b\c` (`if ec.value() == 0`)

crosire:
Merged. I wouldn't worry about stdlib behavior. If it's not always perfect for edge cases that's fine, as long as it generally works (ReShade doesn't need to handle everything).

PR #229, opened 2022-02-04
Fix D3D11 src and dst layer arguments inverted in copy_texture_region

fholger:
Stumbled upon this by chance; it breaks VR rendering in newer Unity games which use an array texture for submission.


crosire:
Nice catch!!

PR #230, opened 2022-02-20
Add features to texture addons

EliasWatson:
# Texture Dump
- Added options header to configure addon behavior
- Added option to switch between full hashing and TexMod hashing
- Added support for saving textures to a subdirectory
- Added support for saving PNG files
- Added saving texture hashes to a set for optimization purposes.
  If a texture has already been dumped in the current session, then the dump function will not write it to disk again.
  This could be further improved by reading the filenames from the texture subdirectory and loading the hashes into the set, preventing the addon from dumping textures that were dumped in a previous gameplay session.

Default behavior is to save PNG textures to `<reshade-dir>/texdump/<texture-hash>.png`
Optimization is enabled by default

image

# Texture Replace
- Added options header to configure addon behavior
- Added option to switch between full hashing and TexMod hashing
- Added support for loading textures from a subdirectory
- Added support for loading PNG files

Default behavior is to load PNG textures from `<reshade-dir>/texreplace/<texture-hash>.png`

_(By "reshade-dir", I mean the directory that has the game executable, "ReShade.ini", "reshade-shaders/", etc)_

### DIFF
--- a/examples/04-texture_dump/dump_options.hpp
+++ b/examples/04-texture_dump/dump_options.hpp
@@ -1,0 +1,30 @@
+#pragma once
+
+/// CONSTANTS
+// You shouldn't need to modify these if you're just editing options
+#define DUMP_HASH_FULL 0
+#define DUMP_HASH_TEXMOD 1
+
+#define DUMP_FMT_BMP 0
+#define DUMP_FMT_PNG 1
+
+
+/// OPTIONS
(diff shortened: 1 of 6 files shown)


GrantMcLravy:
Typical PNG supremacist. BMP is a relevant and valuable format and should still be the default!

PR #231, opened 2022-03-19
Remove unused build target "Release App"

seri14:


crosire:
I do use the "Release App" configuration actually. Not often, but every now and then I profile the effect compiler performance with it (which is why that configuration has the `Profile` flag set to `true` under `<ItemDefinitionGroup Condition="'$(Configuration)|$(Platform)'=='Release App|x64'">`). It's not exposed as a Solution configuration, which is why it looks unused, but it can still be built via Build > Batch Build (same way I'm building the "Release Signed" configuration, which too is not exposed via a Solution configuration.

PR #233, opened 2022-05-07
Added swapchain_desc

MajorPainTheCactus:
Added swapchain_desc to create_swapchain event in order to have more control over the creation of the swap chain (required for enabling HDR in applications)

### DIFF
--- a/include/reshade_api_device.hpp
+++ b/include/reshade_api_device.hpp
@@ -951,6 +951,27 @@ namespace reshade::api
 	};
 
 	/// <summary>
+	/// Describes a swap chain and its back buffer resources.
+	/// </summary>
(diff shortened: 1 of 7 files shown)


crosire:
Great! Only issue I'm having is that changing the signature of the `create_swapchain` event like this is a breaking change that would cause existing add-ons using it to crash. There is a way around this though: Instead of adding a new parameter, can extend the existing `back_buffer_desc` parameter to not be a `resource_desc`, but a new `swapchain_desc` type that derives from `resource_desc`. This way existing add-ons will continue to work (they just access the `resource_desc` part of the new type), whereas new add-ons can make use of the added fields.

PR #234, opened 2022-05-14
Ternary operator tries upper-casting here

jqly:
VS emits below error message.

```
Error C2446 ':': no conversion from 'reshade::hook' to 'const named_hook'
```

### DIFF
--- a/source/hook_manager.cpp
+++ b/source/hook_manager.cpp
@@ -317,7 +317,9 @@ static reshade::hook find_internal(reshade::hook::address target, reshade::hook:
 				(target == nullptr || hook.target == target);
 		});
 
-	return it != s_hooks.cend() ? *it : reshade::hook {};
+	if (it != s_hooks.cend())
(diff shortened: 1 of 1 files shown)


crosire:
Which MSVC version are you seeing this with? This should be valid code as per the implicit casting rules applied for the conditional operator (https://en.cppreference.com/w/cpp/language/operator_other#Conditional_operator) and it compiles for me in VC++2017, 2019 and 2022. So I'm confused as to where you get this error from?

jqly:
Yeah, I hit this problem in conformance mode. Because I copied reshade's hook manager, for my own code project ;)

crosire:
Aha! =D Had been confused there for a second. Thanks for reporting this (and sorry for not just merging as-is, just figured that when making it work with conformance mode, might as well fix all problems that pop up in the rest of the codebase as well). Happy to hear part of the code is useful for other projects as well!

PR #236, opened 2022-06-10
Add feature "Favorite preset save path"

seri14:


crosire:
I have some reservations with this sadly:
- The removal of the "+" button makes it a bit difficult to create a new preset (there is a "Create" button in the dialog, but it disappears when one clicks on any preset and doesn't come back without closing/reopening the dialog) and removed the option to start with/without a duplicate of the current preset.
- I'm not entirely sure there is an advantage to an explicit favorite preset selection: If the current preset is in a directory with all the other favorite presets, then opening the selection dialog did already show a list of all of them before, so this only seems to add value if the current preset is somewhere else (which is probably unlikely, apart from right after installation, but then the favorite preset path would not yet be set up either).
- I prefer to logically separate code in different translation units. Meaning e.g. ini_file.cpp should ideally only do INI management and not know what those INI files contain (not know what "Techniques" are). This way it works nicely to pick files out of the ReShade codebase and use them in other projects (and I know of projects from other people that use a subset of ReShade source code files too, like hook_manager.cpp/hpp or ini_file.cpp/hpp; these files are mostly standalone and have no other dependencies for this reason).

However, `ini_file::delete_cache` is a great addition to avoid unnecessary memory consumption for files that are not needed in cache!

seri14:
I do not like the "New Preset" button. ("+")
ReShade's preset editor is not similar to other document editors because it uses the camera shot method. The preset-file is updated each frame in the game. Even if the "Save Overwrite" button is not pressed.
The current preset path is only means it is output path of updated/captured data.

When specified a new preset name, Users needs to choose whether to keep the full-text every time.
(Maybe you will say that users are not always force required to choise it now. It is possible to say it if the checkbox is only in the settings tab.)

I always want to create a preset based on my pre-tuned anti-aliasing settings. However, there are so many items to set each time that it is impossible to remember.
It would be nice to ReShade can create a preset using a template, but it can not.

Trimming is mostly not important. It may required only in publishing to MFT record.

I think it is important to designing the editor to makes it easy for users to create a variety of preset sets with minimal re-configuration.

The "Reset" button is compatible for ReShade's preset editor, and the editor already have some another reset menu. So, I think it would be better to add a more powerful reset menu, "Reset All (delete all sections)".


I am sure that the developing favorites menu is not complete enough.
I would really like to add a default folder, but I am not sure that it would behave valid in a real environment.

My goal is to add/integrate a feature similar to the classic preset selector before the preset selector was implemented.

crosire:
You can actually create a new preset from template: For that first choose the template preset file and once it is active click on + to create a new preset and check "Duplicate current preset". That will give you a new preset with all the default settings from the template.

Creating a new preset places it in the same directory as the previous preset, so there already is behavior to save in a favorite location (provided that is where the previous preset was located). I'm not sure where adding another configurable path to the mix brings benefits to users.

I just think that the current preset editor already behaves exactly like the classic one, if things are set up so that all presets are in a dedicated directory.

crosire:
I added an option that allows one to disable automatic saving of the preset when something changes (so one has to click the save button for changes to save), which also adds a "Reset" button to restore everything to the previous preset state on disk: https://github.com/crosire/reshade/commit/485b9b96f87d4dcc0414bc95ea6fa46308af0bc8

seri14:
I think good to revert your changes.
or you should add warning dialog "Unsaved" for all leaving cases.

I do not want to increase the cases where users lost their data.

I said This is Camera shot method.

Please teach me the softwares  "Any data are not output to disk in recording"

seri14:
You must add code "Always reload preset file when enable the option"
Because reloaded data is unflushed cache. Users confuse "OK, flushed it."

PR #239, opened 2022-11-19
Fix exception when interacting with key_input_box

Mortalitas:
ImGui's BeginDisabled/EndDisabled are very sensitive to variable state changes and unfortunately were throwing an exception on first-key input here.

Also see:
https://github.com/ocornut/imgui/issues/5011#issuecomment-1037556295


crosire:
Thanks for catching this! The fix is not quite right though, since it will still fail when current input source is a gamepad (since last input was from a gamepad), but the user presses a keyboard button anyway, while the control is active. Will push a slightly different fix which handles that case as well shortly.

PR #241, opened 2023-01-27
d3d9 - Fix using ForceResolution, etc. in fullscreen mode

retroNUC:
Copy any modified presentation params back into fullscreen_desc so it matches, otherwise IDirect3D9Ex_CreateDeviceEx will fail


retroNUC:
Never mind, figured out actual issue, it's slightly different.

PR #242, opened 2023-02-11
Play sound when taking screenshot

seri14:
Probably in most cases, Users set the camera sound.

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -30,6 +30,7 @@
 #include <malloc.h>
 #include <dwmapi.h>
 #include <d3dcompiler.h>
+#include <mmsystem.h>
 
 #if RESHADE_FX
 bool resolve_path(std::filesystem::path &path, std::error_code &ec)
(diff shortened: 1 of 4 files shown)


source/runtime_gui.cpp
```diff
@@ -1672,6 +1672,9 @@ void reshade::runtime::draw_gui_settings()
 		}
 
 		modified |= imgui::directory_input_box("Screenshot path", _screenshot_path, _file_selection_path);
+		modified |= imgui::directory_input_box("Screenshot sound", _screenshot_playsound_path, _file_selection_path);
```
crosire, line 1675: This should probably be `imgui::file_input_box("Screenshot sound", "sound.wav", _screenshot_sound_path, _file_selection_path, { L".wav" });`, otherwise can't select an audio file.

source/runtime.hpp
```diff
@@ -343,6 +343,9 @@ namespace reshade
 		unsigned int _screenshot_format = 1;
 		unsigned int _screenshot_jpeg_quality = 90;
 		unsigned int _screenshot_key_data[4] = {};
+		std::filesystem::path _screenshot_playsound_path;
```
crosire, line 346: Pet peeve: I'd just rename that to `_screenshot_sound_path`.

source/runtime.cpp
```diff
@@ -859,6 +860,9 @@ void reshade::runtime::load_config()
 #endif
 
 	config.get("SCREENSHOT", "SavePath", _screenshot_path);
+	config.get("SCREENSHOT", "SoundPath", _screenshot_playsound_path);
+	config.get("SCREENSHOT", "PlayDefaultIfNotExist", _screenshot_playsound_force);
+	config.get("SCREENSHOT", "PlaySoundAsSystemNotification", _screenshot_playsound_as_system_notification);
```
crosire, line 865: I'm usually all for configurability, but in this case I'm questioning whether we really need these two extra options.
It is probably fine to always play the default ding sound if the file cannot be found (can avoid the additional file exists in our code that way) and is probably also sufficient to never play in the system notification channel?

seri14: ReShade can not adjust the volume, so we should use global volume mixer. Well, not everyone need it.

seri14: Users can specify a non-playable file for play the default sound, but the user control does not allow it. So I removed the behavior.

PR #243, opened 2023-02-11
Add option to force preset selection on startup

seri14:
Users can use this option for always select the specific preset on startup.

Use cases
* Daily preset editing
* Reset long distance exploring between base path to last selection
* Load reduction on loading screens instead of NoReloadOnInit

~~And this PR contain a bugfix
Force insert relative path to relative path is broken~~

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -826,6 +826,8 @@ void reshade::runtime::load_config()
 	config.get("GENERAL", "TextureSearchPaths", _texture_search_paths);
 	config.get("GENERAL", "IntermediateCachePath", _effect_cache_path);
 
+	config.get("GENERAL", "StartupPresetPath", _startup_preset_path);
+
 	config.get("GENERAL", "PresetPath", _current_preset_path);
 	config.get("GENERAL", "PresetTransitionDuration", _preset_transition_duration);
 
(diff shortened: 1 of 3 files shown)


source/runtime.cpp
```diff
@@ -905,12 +911,24 @@ void reshade::runtime::save_config() const
...
+
+	if (startup_preset_path.native().rfind(L"..", 0) != std::wstring::npos)
+		startup_preset_path = _startup_preset_path; // Do not use relative path if preset is in a parent directory
+	if (startup_preset_path.is_relative()) // Prefix preset path with dot character to better indicate it being a relative path
+		startup_preset_path = L"." / startup_preset_path;
+
+	config.set("GENERAL", "StartupPresetPath", startup_preset_path);
+
+	// Use ReShade DLL directory as base for relative preset paths (see 'resolve_preset_path')
+	std::filesystem::path current_preset_path = (g_reshade_base_path / _current_preset_path).lexically_proximate(g_reshade_base_path);
```
crosire, line 924: The `(g_reshade_base_path / _current_preset_path)` should not be necessary, as `_current_preset_path` is always guaranteed to be an absolute path (at load time it is made absolute via the call in `if (!resolve_preset_path(_current_preset_path, ec))`, and when changed in the overlay it is made absolute in `imgui::file_dialog()`). As such the previous code should already work.

source/runtime.cpp
```diff
@@ -839,8 +841,12 @@ void reshade::runtime::load_config()
 			LOG(ERROR) << "Failed to create effect cache directory " << _effect_cache_path << " with error code " << ec.value() << '!';
 	}
 
+	// Use startup preset instead of last selection
+	if (std::error_code ec; !_startup_preset_path.empty() && resolve_preset_path(_startup_preset_path, ec))
```
crosire, line 845: Can re-use the `std:error_code` that is declared in line 833 above here, instead of having to create a new instance.

source/runtime.cpp
seri14: correct. This is the correct process I misunderstood.

source/runtime.cpp
seri14: I didn't notice that variable. It will be fixed.

PR #246, opened 2023-02-20
Use section structure instead of ini-key value

seri14:


seri14:
Rewrote for https://github.com/crosire/reshade/commit/edf6d0330a785d0596826a82c71d10b976fa7ec5

PR #247, opened 2023-02-22
Add macro of screenshot/postsavecommand %Count% and %Rotate%

seri14:
Users can use this macro to avoid overwriting already saved file when saving

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -599,6 +599,9 @@ void reshade::runtime::on_present()
 		}
 	}
 
+	if (_should_save_screenshot)
+		_screenshot_count++;
+
 #if RESHADE_FX
 	update_effects();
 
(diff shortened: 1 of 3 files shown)


crosire:
I don't quite understand the purpose of `%Rotate%` (`%Count%` makes sense to me)?

seri14:
`Count` is the times of screenshots, so if you replace `TimeSecond` with it, filename duplication is not occur as long as times is below the average frame rate.
In the first commit, the number of digits was 1 (supports up to 10 times per second).
https://github.com/crosire/reshade/compare/0ff75fd4508a643cab9d0881a7dc4052f64de652..c3c0f0ed1c34d853ef0bcf2461dd7aa9bfab7fe7
It might make sense than that.
Actually, the number of digits has changed by my own reason.
I developing an add-on "take screenshot repeatly for the specified number of frames".

seri14:
High-end PCs run games at 144-240 FPS. If the user wants to take continuous screenshots for 1 second, 1 digits is not enough, 2 digits is not enough, 3 digits is little far to rotation, but enough.

seri14:
An actual example, the game give the rotate to the screenshot name, the number of digits is 3.
`gameshortnameyyyyMMdd_HHmmss_RRR.ext` (R is Rotate)

PR #248, opened 2023-03-05
Show preview texture when effects are disabled

seri14:


crosire:
There is probably a backstory to this? (Since the preview texture so far wasn't shown when effects are disabled intentionally, since nothing would update it then.)

PR #249, opened 2023-03-06
Add API event reshade_available_back_buffer

seri14:
This PR contains these changes

* ~~Add API event `reshade_available_back_buffer`~~
* Change overlay condition to not make movable OSD window only rendered by add-ons
* Add API `get_reshade_base_path` to resolve relative paths

Add-on side
https://github.com/seri14/reshade-addons/blob/7c376f434f1415a256d6c4e68b3e43dbe40eeacc/src/addon-screenshot/dllmain.cpp#L77-L87

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -607,6 +607,10 @@ void reshade::runtime::on_present()
 		}
 	}
 
+#if RESHADE_ADDON
+	invoke_addon_event<addon_event::reshade_available_back_buffer>(this);
+#endif
+
 #if RESHADE_FX
 	update_effects();
 
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -607,10 +607,6 @@ void reshade::runtime::on_present()
 		}
 	}
 
-#if RESHADE_ADDON
-	invoke_addon_event<addon_event::reshade_available_back_buffer>(this);
-#endif
-
 #if RESHADE_FX
 	update_effects();
 
--- a/source/runtime_gui.cpp
+++ b/source/runtime_gui.cpp
@@ -213,9 +213,6 @@ void reshade::runtime::load_config_gui(const ini_file &config)
 #endif
 	config.get("OVERLAY", "ShowFPS", _show_fps);
 	config.get("OVERLAY", "ShowFrameTime", _show_frametime);
-#if RESHADE_ADDON
-	config.get("OVERLAY", "ShowAddonsOSD", _show_addons_osd);
-#endif
 	config.get("OVERLAY", "ShowScreenshotMessage", _show_screenshot_message);
 #if RESHADE_FX
 	config.get("OVERLAY", "TutorialProgress", _tutorial_index);
@@ -309,9 +306,6 @@ void reshade::runtime::save_config_gui(ini_file &config) const
 #endif
 	config.set("OVERLAY", "ShowFPS", _show_fps);
 	config.set("OVERLAY", "ShowFrameTime", _show_frametime);
-#if RESHADE_ADDON
-	config.set("OVERLAY", "ShowAddonsOSD", _show_addons_osd);
-#endif
 	config.set("OVERLAY", "ShowScreenshotMessage", _show_screenshot_message);
 #if RESHADE_FX
 	config.set("OVERLAY", "TutorialProgress", _tutorial_index);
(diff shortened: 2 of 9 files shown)


source/runtime_gui.cpp
```diff
@@ -1095,12 +1106,12 @@ void reshade::runtime::draw_gui()
 		{
 			for (const addon_info::overlay_callback &widget : info.overlay_callbacks)
 			{
-				if (widget.title == "OSD" ? show_splash || (!show_stats_window && !_show_overlay) : !_show_overlay)
-					continue;
-
-				if (ImGui::Begin(widget.title.c_str(), nullptr, ImGuiWindowFlags_NoFocusOnAppearing))
-					widget.callback(this);
-				ImGui::End();
+				if (!show_splash && (widget.title == "OSD" ? show_addons_osd : _show_overlay))
```
crosire, line 1109: Normal overlays can show while the splash message is displayed (like the main ReShade overlay), only the OSD one cannot, because they would overlap. That's why `show_splash` was only in the OSD condition previously.

source/runtime.cpp
```diff
@@ -607,6 +607,10 @@ void reshade::runtime::on_present()
 		}
 	}
 
+#if RESHADE_ADDON
+	invoke_addon_event<addon_event::reshade_available_back_buffer>(this);
```
crosire, line 611: As far as I can tell this is basically equivalent to `addon_event::present`, except when the back buffer is multisampled. But in that case this wouldn't be useable too since `effect_runtime::get_current_back_buffer()` returns the real back buffer and not the resolved one.

seri14: How to capture "-original" image?
In develop, this event name was `reshade_original`

source/runtime_gui.cpp
seri14: I tried only under reloading the effects.

source/runtime.cpp
crosire: You can use this little hack to capture a screenshot in the `present` event (which would ve the original image before effects) despite not technically having access to the effect runtime (works because the swapchain pointer indirectly is also an effect runtime still):
```c++
static void on_present(reshade::api::swapchain *swapchain, ...)
{
    const auto runtime = static_cast<reshade::api::effect_runtime *>(swapchain);
    runtime->capture_screenshot(...);
}
```

seri14: removed. [Add-on side](https://github.com/seri14/reshade-addons/commit/d8f16709fcd4ba820b2f2d79d330b9769dd7a4b3)

seri14:
There is no way to hide the Add-on OSD like as "Show FPS".
I noticed that, but I did not think it really slow.
I tried wrote [this code](https://github.com/crosire/reshade/commit/a6f09a78439e9bfe9e82dce714006ab1b255d7a9#diff-cc7f0445b600f931e2403ccf42587d1c134e0fada768bfbec09625f09a7bdb14R684-R694) in `reshade::load_addons` and `runtime::on_init`.
Can I re-challenge this PR with these changes?
```cpp
bool show_stats_window = show_clock || show_fps || show_frametime || (_show_addons_OSD && _has_addons_OSD);
```

crosire:
Wouldn't it make more sense for the add-on to provide a setting to turn its OSD display on or off (by showing relevant checkbox(es) in the settings of the add-on in the add-on list)? Since the add-on has more fine-grained control over this, whereas ReShade would only be able to blanket turn everything on/off for all add-ons. Would still have to always register for the OSD overlay, but simply don't do anything in the callback when it is turned off.

seri14:
I do not think that is right.
If users has 10 add-ons, users must check-off 10 times if they want to hide OSD's settings window (in "Settings" tab? in "My Addon Setting Window" window?, or only in %ShowMyGoodAddon% environment variable?).
Your idea should replace with that Add setting "Show window" each add-on in the "Addons" tab apart from "ShowAddonsOSD".
(My image [x] "Main", [ ] "High-cost CPU Window", [x] "OSD")
Do not pressue ReShade GUI operations to add-ons. Add-ons can implement any code but must not have "Template".
Add-ons Developers not developing in any time.

crosire:
While I agree that it's not ideal that different add-ons could solve this differently, the same is true for everything the add-on API provides. It explicitly provides no restrictions on how ImGui can be used. An add-on could just as well draw its own OSD window if it wants to. So there isn't really benefit in restricting this one extremely specific use-case, when it can technically just be worked around anyway. Add-ons can implement any code, and yes, that code may not always be great, but ReShade won't judge that or limit it. Since one of the reasons the add-on API exists is to ensure people can modify/extend ReShade in ways I would not agree to merge into the core.

PR #250, opened 2023-03-12
Add effect scope for preprocessor definitions

seri14:
This PR contains these changes

* Add `effect` scope (See [My add-on side](https://github.com/seri14/reshade-addons/compare/main...addon-uibind))
* ~Add API `bool has_effects_loaded()` to avoid touch _effects/_techniques/_textures/_uniforms while loading effects~
  Cancelled. See https://github.com/crosire/reshade/pull/250#issuecomment-1478615159
* Increase text buffer size from 128 to 256 (for to_string(float4x4) requires at least 144 bytes)
* Display definition default value as hint instead of input text

### DIFF
--- a/include/reshade_api.hpp
+++ b/include/reshade_api.hpp
@@ -553,11 +553,18 @@ namespace reshade::api
 			return get_preprocessor_definition(name, value, &length);
 		}
 		/// <summary>
-		/// Defines a global preprocessor definition to the specified <paramref name="value"/>.
+		/// Defines a preprocessor definition to the specified <paramref name="value"/>.
 		/// </summary>
 		/// <param name="name">Name of the definition.</param>
 		/// <param name="value">Value of the definition.</param>
 		virtual void set_preprocessor_definition(const char *name, const char *value) = 0;
+		/// <summary>
+		/// *TODO*
+		/// </summary>
+		/// <param name="effect_name">*TODO*</param>
+		/// <param name="name">Name of the definition.</param>
+		/// <param name="value">Value of the definition.</param>
+		virtual void set_preprocessor_definition(const char *effect_name, const char *name, const char *value) = 0;
 
 		/// <summary>
 		/// Applies a <paramref name="technique"/> to the specified render targets (regardless of the state of this technique).
@@ -660,5 +667,11 @@ namespace reshade::api
 			size_t length = SIZE;
 			get_technique_effect_name(technique, effect_name, &length);
 		}
+
+		/// <summary>
+		/// *TODO*
+		/// </summary>
+		/// <returns>*TODO*</returns>
+		virtual bool has_effects_loaded() const = 0;
 	};
 }
--- a/include/reshade_api.hpp
+++ b/include/reshade_api.hpp
@@ -689,11 +689,5 @@ namespace reshade::api
 		/// <param name="name">Name of the definition.</param>
 		/// <param name="value">Value of the definition.</param>
 		virtual void set_preprocessor_definition(const char *effect_name, const char *name, const char *value) = 0;
-
-		/// <summary>
-		/// *TODO*
-		/// </summary>
-		/// <returns>*TODO*</returns>
-		virtual bool has_effects_loaded() const = 0;
 	};
 }
--- a/include/reshade_api.hpp
+++ b/include/reshade_api.hpp
@@ -558,13 +558,6 @@ namespace reshade::api
 		/// <param name="name">Name of the definition.</param>
 		/// <param name="value">Value of the definition.</param>
 		virtual void set_preprocessor_definition(const char *name, const char *value) = 0;
-		/// <summary>
-		/// *TODO*
-		/// </summary>
-		/// <param name="effect_name">*TODO*</param>
-		/// <param name="name">Name of the definition.</param>
-		/// <param name="value">Value of the definition.</param>
-		virtual void set_preprocessor_definition(const char *effect_name, const char *name, const char *value) = 0;
 
 		/// <summary>
 		/// Applies a <paramref name="technique"/> to the specified render targets (regardless of the state of this technique).
@@ -667,6 +660,14 @@ namespace reshade::api
 			size_t length = SIZE;
 			get_technique_effect_name(technique, effect_name, &length);
 		}
+
+		/// <summary>
+		/// *TODO*
+		/// </summary>
+		/// <param name="effect_name">*TODO*</param>
+		/// <param name="name">Name of the definition.</param>
+		/// <param name="value">Value of the definition.</param>
+		virtual void set_preprocessor_definition(const char *effect_name, const char *name, const char *value) = 0;
 
 		/// <summary>
 		/// *TODO*
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -1483,7 +1483,6 @@ bool reshade::runtime::load_effect(const std::filesystem::path &source_file, con
 		pp.add_macro_definition("BUFFER_COLOR_BIT_DEPTH", std::to_string(format_color_bit_depth(_effect_color_format)));
 
 		for (const definition &definition : preprocessor_definitions)
-			if (!definition.first.empty())
 				pp.add_macro_definition(definition.first, definition.second.empty() ? "1" : definition.second);
 
 		for (const std::filesystem::path &include_path : include_paths)
@@ -3481,6 +3480,23 @@ void reshade::runtime::update_effects()
 	// Delay first load to the first render call to avoid loading while the application is still initializing
 	if (_frame_count == 0 && !_no_reload_on_init && !(_no_reload_for_non_vr && !_is_vr))
 		reload_effects();
+
+#if RESHADE_ADDON
+	if (_reload_remaining_effects == std::numeric_limits<size_t>::max())
+	{
+		if (const size_t effect_index = _should_save_preprocessor_definitions;
+			effect_index != std::numeric_limits<size_t>::max())
+		{
+			save_current_preset();
+			_should_save_preprocessor_definitions = std::numeric_limits<size_t>::max();
+
+			if (effect_index < _effects.size())
+				reload_effect(effect_index);
+			else
+				reload_effects();
+		}
+	}
+#endif
 
 	if (_reload_remaining_effects == 0)
 	{
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -447,7 +447,7 @@ bool reshade::runtime::on_init(input::window_handle window)
 	_frame_count = 0;
 
 	_is_initialized = true;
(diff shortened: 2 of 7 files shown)


source/runtime.cpp
```diff
@@ -1317,7 +1366,7 @@ bool reshade::runtime::switch_to_next_preset(std::filesystem::path filter_path,
 	return true;
 }
 
-bool reshade::runtime::load_effect(const std::filesystem::path &source_file, const ini_file &preset, size_t effect_index, bool preprocess_required)
+bool reshade::runtime::load_effect(std::filesystem::path source_file, const ini_file &preset, size_t effect_index, bool preprocess_required)
```
crosire, line 1369: What's the reason for this change?

source/runtime.cpp
```diff
@@ -3182,9 +3228,27 @@ void reshade::runtime::reorder_techniques(std::vector<size_t> &&technique_indice
...
+
+	_preprocessor_definitions.erase(std::remove_if(_preprocessor_definitions.begin(), _preprocessor_definitions.end(),
+		[](const definition &definition) { return definition.scope == definition::scope_preset || definition.scope == definition::scope_effect; }), _preprocessor_definitions.end());
+
+	if (std::vector<std::pair<std::string, std::string >> preset_preprocessor_definitions;
+		preset.get({}, "PreprocessorDefinitions", preset_preprocessor_definitions))
+		for (std::pair<std::string, std::string> &value : preset_preprocessor_definitions)
+			_preprocessor_definitions.push_back(definition{ definition::scope_preset, "", std::move(value.first), std::move(value.second) });
+
+	if (std::vector<std::string> effect_names; preset.get(effect_names))
```
crosire, line 3241: This new `ini_file` method could be avoided as there already is a list of effect files (`effect_files`) build below, so could simply iterate over that to query the preset for each file.

source/runtime.cpp
```diff
@@ -991,15 +1003,36 @@ void reshade::runtime::load_current_preset()
...
+
+	preprocessor_definitions.erase(std::remove_if(preprocessor_definitions.begin(), preprocessor_definitions.end(),
+		[](const definition &definition) { return definition.scope == definition::scope_preset || definition.scope == definition.scope_effect; }), preprocessor_definitions.end());
+
+	if (std::vector<std::pair<std::string, std::string >> preset_preprocessor_definitions;
+		preset.get({}, "PreprocessorDefinitions", preset_preprocessor_definitions))
+		for (std::pair<std::string, std::string> &value : preset_preprocessor_definitions)
+			preprocessor_definitions.push_back(definition{ definition::scope_preset, "", std::move(value.first), std::move(value.second) });
+
+	if (std::vector<std::string> effect_names; preset.get(effect_names))
```
crosire, line 1017: And in this case can just iterate over `_effects` to get the effect file names.

source/runtime.cpp
```diff
@@ -991,15 +1003,36 @@ void reshade::runtime::load_current_preset()
...
+	std::stable_sort(preprocessor_definitions.begin(), preprocessor_definitions.end(), [](const definition &a, const definition &b) { return a.first < b.first; });
+	std::stable_sort(preprocessor_definitions.begin(), preprocessor_definitions.end(), [](const definition &a, const definition &b) { return a.name < b.name; });
+	std::stable_sort(preprocessor_definitions.rbegin(), preprocessor_definitions.rend(), [](const definition &a, const definition &b) { return a.scope < b.scope; });
 
 	// Recompile effects if preprocessor definitions have changed or running in performance mode (in which case all preset values are compile-time constants)
 	if (_reload_remaining_effects != 0) // ... unless this is the 'load_current_preset' call in 'update_effects'
 	{
-		if (_performance_mode || preset_preprocessor_definitions != _preset_preprocessor_definitions)
+		if (_performance_mode || !std::equal(preprocessor_definitions.cbegin(), preprocessor_definitions.cend(), _preprocessor_definitions.cbegin(), _preprocessor_definitions.cend(),
+			[](const definition &a, const definition &b) { return a.scope == b.scope && a.name == b.name && a.first == b.first && a.second == b.second; }))
```
crosire, line 1033: It would be better to simply use the `operator!=` of the vector as before. Might have to implement a comparion operator for the `definition` struct, but since it's so simply I'd actually expect it to work out of the box.

source/runtime.cpp
```diff
@@ -3182,9 +3228,27 @@ void reshade::runtime::reorder_techniques(std::vector<size_t> &&technique_indice
...
+		for (std::string &effect_name : effect_names)
+			if (effect_name.rfind(".fx") != std::string::npos)
+				if (std::vector<std::pair<std::string, std::string >> effect_preprocessor_definitions;
+					preset.get(effect_name, "PreprocessorDefinitions", effect_preprocessor_definitions))
+					for (std::pair<std::string, std::string> &value : effect_preprocessor_definitions)
+						_preprocessor_definitions.push_back(definition{ definition::scope_effect, effect_name, std::move(value.first), std::move(value.second) });
+
+	std::stable_sort(_preprocessor_definitions.begin(), _preprocessor_definitions.end(), [](const definition &a, const definition &b) { return a.first < b.first; });
+	std::stable_sort(_preprocessor_definitions.begin(), _preprocessor_definitions.end(), [](const definition &a, const definition &b) { return a.name < b.name; });
+	std::stable_sort(_preprocessor_definitions.rbegin(), _preprocessor_definitions.rend(), [](const definition &a, const definition &b) { return a.scope < b.scope; });
```
crosire, line 3251: These could be combined into a single sort condition.

include/reshade_api.hpp
```diff
@@ -660,5 +667,11 @@ namespace reshade::api
 			size_t length = SIZE;
 			get_technique_effect_name(technique, effect_name, &length);
 		}
+
+		/// <summary>
+		/// *TODO*
+		/// </summary>
+		/// <returns>*TODO*</returns>
+		virtual bool has_effects_loaded() const = 0;
```
crosire, line 675: Why is this needed? Events are never called during effect loading anyway, so this should not be needed. You can use `reshade_reloaded_effects` to keep track of when effects are loaded.

include/reshade_api.hpp
```diff
@@ -553,11 +553,18 @@ namespace reshade::api
...
 		/// <param name="name">Name of the definition.</param>
 		/// <param name="value">Value of the definition.</param>
 		virtual void set_preprocessor_definition(const char *name, const char *value) = 0;
+		/// <summary>
+		/// *TODO*
+		/// </summary>
+		/// <param name="effect_name">*TODO*</param>
+		/// <param name="name">Name of the definition.</param>
+		/// <param name="value">Value of the definition.</param>
+		virtual void set_preprocessor_definition(const char *effect_name, const char *name, const char *value) = 0;
```
crosire, line 567: This will break the vtable layout and crash with any existing add-ons using the `render_technique` or below functions (which do exist already).

source/runtime_objects.hpp
```diff
@@ -282,5 +282,17 @@ namespace reshade
 		};
 		std::vector<binding_data> texture_semantic_to_binding;
 	};
+
+	struct definition
+	{
+		unsigned int scope = 0;
```
crosire, line 288: Nicer to do this IMHO:
```c++
struct definition
{
    enum
    {
        scope_global,
        scope_preset,
        scope_effect
    } scope;

    std::string name;
    ...
};
```
Usage stays the same.

crosire:
This is good to be merged, only hoping that adding the `has_effects_loaded` method can be avoided.

seri14:
What your idea?

crosire:
I don't know why you need it, so I can't come up with potential alternatives.

seri14:
Did you read my comment?
What part did you not understand?

crosire:
As per https://github.com/crosire/reshade/pull/250#discussion_r1136088467, no events that pass effect handles are called while effects are loaded and as per https://github.com/crosire/reshade/blob/main/include/reshade_api.hpp#L16 these handles are only valid until a `addon_event::reshade_reloaded_effects` event is called, at which point they must be re-enumerated. If following this requirement, it is not possible to be in an event callback and access an effect handle while effects are being loaded (since that event is called before and after loading). In case one still tries that, the ReShade API will prevent it from happening (all the APIs operating on effect handles have a `is_loading` check). Therefore there is no use for a public facing `has_effects_loaded` API.

seri14:
I will say again.
call save_current_preset while loading effects.
It really save same as before loading?

crosire:
? You never said anything about `save_current_preset` before (+ this was only just added). But it shouldn't be possible to call this during effect loading anyway (it would benefit from a `is_loading` check to ensure though, that was missed when it was made public just now).

seri14:
Pending comments are now visible.

See this discussion
https://github.com/community/community/discussions/10369

crosire:
Heh, I ran into that before as well ...
I can see the comments now and see where you are coming from. It's also not great that with multiple `ui_bind` variables the same effect potentially has to be reloaded multiple times (on every `set_preprocessor_definition`). So maybe a better overall solution would be:
- Remove the `save_current_preset` call from `set_preprocessor_definition` (just like was done for `set_uniform_value_*`)
- Don't `reload_effects` from `set_preprocessor_definition`, instead only set a flag that indicates that preprocessor definitions are dirty and effects need to be reloaded at the next most convenient time
- Then just extend the `if (_frame_count == 0 && !_no_reload_on_init && !(_no_reload_for_non_vr && !_is_vr))` condition in `update_effects` to also check whether this dirty flag is set
This way `set_preprocessor_definition` works regardless of whether effects are still being loaded at the moment, so it's pretty safe to call always and less likely to cause bugs in add-ons and we don't need the public `has_effects_loaded` method. And preset saving is explicit for preprocessor definitions as well, which is more consistent.

source/runtime.cpp
```diff
@@ -4350,7 +4350,7 @@ template <> void reshade::runtime::get_uniform_value<uint32_t>(const uniform &va
 void reshade::runtime::set_uniform_value_data(uniform &variable, const uint8_t *data, size_t size, size_t base_index)
 {
 #if RESHADE_ADDON
-	if (has_addon_event<addon_event::reshade_set_uniform_value>() && !is_loading() && !_is_in_api_call)
```
seri14, line 4296: I think it seems not have effect.
I thought the has_addon_event should use if call add-on with high cost processing(e.g. make big array, call syscall, ...)
Inserted at https://github.com/crosire/reshade/blame/10e449d7e71aab07fe7b51a4abc7ea6028f7f6be/source/runtime.cpp#L4295-L4304

crosire: Yes, that is true. I only did this change for code aesthetics, there is no other practical reason behind it

seri14:
Add-on side
https://github.com/seri14/reshade-addons/commit/2a3cf075969e4886774237dedb46eb81bba5d068

PR #251, opened 2023-03-13
Fix "reset_uniform_value" no raise "reshade_set_uniform_value" event when uniform has no initializer

seri14:


crosire:
I wouldn't consider this a bug, since uniforms default to zero, so there really shouldn't be a `reshade_set_uniform_value` event?

seri14:
My image is that the event is raise when content of data storage is modifing.
so I thought current behavior is buggy.
I guess that the real name of the event is `reshade_edit_uniform_value` in your image.

seri14:
And this is bad API design.
You added `is_loading()` to the event raise condition without public `is_loading()` on API, because only you only want to get it when editing.

seri14:
If you create a new `edit` event, rename it and add a new `set` event. In that case, the add-on side behavior will same until recompile.

crosire:
Missed that `reset_uniform_value` can also be called when the user resets the value in the overlay, so raising the event then makes sense.
The reason why these are not raised when `is_loading()` is true is not because only I want that, no, it is very much rooted in reality: There cannot be a persistent `effect_uniform_variable` handle during loading. It's not a valid pointer yet, since pointing to a temporary variable or pointing to an element of the uniforms array of an effect that is not finalized yet (more uniforms can be added which can cause the vector to resize and reallocate, invalidating all previous pointers/uniform handles). Not to mention that loading happening on multiple threads makes enumerating handles impossible due to potential race conditions.

seri14:
It is right. but the event seems it was just put without lock _effects and _uniforms.
I talking about API design. It should have been avoided at that time.

PR #252, opened 2023-03-17
Added a remark to the reshade_reloaded_effects event to signal how many times it gets called

FransBouma:
I ran into this and as it's not clear what's going on, I documented it. 

~~E.g. I had 2 presets in my test setup and when clicking 'Reload' this event was called twice, once for each preset. This is logical of course but I understood from the documentation on the method it would be called once.~~

Updated, see posts below.

### DIFF
--- a/include/reshade_events.hpp
+++ b/include/reshade_events.hpp
@@ -1479,6 +1479,7 @@ namespace reshade
 		/// Called right after all ReShade effects were reloaded.
 		/// This occurs during effect runtime initialization or because the user pressed the "Reload" button in the overlay.
 		/// <para>Callback function signature: <c>void (api::effect_runtime *runtime)</c></para>
+		///	<remarks>This is called after every preset has been reloaded. This means that if the user has e.g. 3 presets loaded into ReShade, this event is called three times. </remarks>
 		/// </summary>
 		reshade_reloaded_effects,
 
--- a/include/reshade_events.hpp
+++ b/include/reshade_events.hpp
@@ -1479,7 +1479,8 @@ namespace reshade
 		/// Called right after all ReShade effects were reloaded.
 		/// This occurs during effect runtime initialization or because the user pressed the "Reload" button in the overlay.
 		/// <para>Callback function signature: <c>void (api::effect_runtime *runtime)</c></para>
-		///	<remarks>This is called after every preset has been reloaded. This means that if the user has e.g. 3 presets loaded into ReShade, this event is called three times. </remarks>
+		///	<remarks>The callback is called twice: first right after all effects have been destroyed, and second right after the effects have been reloaded and all techniques enabled in the active
+		///	preset have been enabled.</remarks>
 		/// </summary>
 		reshade_reloaded_effects,
 


FransBouma:
On second thought, isn't this actually a bug? when I add a new preset, I get yet another call to this event handler, however when I query the current preset path, I get the same value back every time. So I don't know which preset is currently the 'active' one when this event is raised. 

This makes it complicated to migrate uniform variables to their new handles as different presets might have different techniques enabled (e.g. I have 1 preset with 0 techniques enabled, which happens to be the first one that's reloaded so when I enumerate active techniques, 0 are returned), but it's unclear how to determine which 'preset' is the one that was just loaded when the call arrives.

seri14:
ReShade always works with one preset. It does not enumerate/load another preset files.
You give an example about a preset that has 0 enabled techniques, I guess you misrecognize the initial value (zeroed memory data) as data of that preset.

I guess you got a hit here.
https://github.com/crosire/reshade/blob/4fa96c4c144732315847a81f28b632e2fabebb75/source/runtime.cpp#L3240-L3246

Currently, ReShade requires that add-on developers be well versed the reshade/_gui. Some events have APIs that are not allowed to call.

I highly recommend to build an environment where you can debug both ReShade and the add-on at the same session.
Place breakpoints and track changes to runtime members.

```props
<PostBuildEvent>
  <Command>"XCOPY" "$(OutDir)$(TargetName).*" "$(ProjectDir)..\..\deps\reshade\bin\$(Platform)\$(Configuration) App\" /Y /Q</Command>
  <Message>Place the built add-on files to ReShade's debug application folder.</Message>
</PostBuildEvent>
```

FransBouma:
@seri14 yeah I get all that, but that's not the case. When I have 2 presets, my callback registered to that event is called 2 times, when I add another preset, it gets called 3 times when I click reload.

The problem is that I have stored uniform variables in my addon and I have to migrate these to their new handles after a reload of effects happens. However as I have 1 preset without any techniques enabled and 1 with techniques enabled (which is the one I have active) and the 1 without any active techniques is hit first, when I enumerate the enabled techniques in my callback to reshade_reloaded_effects, none are enabled, so I have to release all my uniform variables (as I don't know if this is the real state!). I get another call after that for my active preset with enabled techniques so my code can't do anything else but to store *that* state.

I expected just 1 call to the callback of reshade_reloaded_effects no matter how many presets I have currently defined, but that's not the case. 

> Currently, ReShade requires that add-on developers be well versed the reshade/_gui. Some events have APIs that are not allowed to call.
Not sure what that means, tbh... I am not calling anything, my callback gets called too many times

FransBouma:
I've tested again and @seri14 you're correct, it doesn't call it per preset, so reshade always calls the callback two times: once after the effects are destroyed and once after the effects have been updated. 

This isn't clear from the API comment. @crosire not sure if this has to stay this way, as it's 2 events wrapped into 1 event: effects_destroyed and effects_reloaded. An addon now has to count the calls and do work according to whether it's the first or the second call to the event callback. IMHO not ideal.

crosire:
There is no guarantee made about how many times this event is called. On top of this currently being called during destruction and after all active effects were initialized, this can also be called again an undetermined amount of times later: Whenever you enable a technique of an effect that was not previously enabled during the session, ReShade has to initialize that effect, which internally counts as a reload and will cause that event to be called as well (even though technically only a single effect was updated, not everything was reloaded from scratch).

FransBouma:
@crosire ok, so this PR can be closed I guess...

The main issue I then am left with is... I want to migrate uniforms from one set of pointers to another and that's only needed when a reload has occurred. But at the same time, I also have to keep track of which techniques are enabled so I can skip those. As the reload event is raised after all have been destroyed, nothing is enabled and therefore nothing gets migrated. By skipping that call I worked around that. But if there are other triggers for this event then I have to ignore technique states I guess and only migrate uniforms and that's it..

crosire:
You can try the `reshade_set_current_preset_path` event. That is called after a preset was loaded, so should have all the state you expect (like which techniques are enabled in the current preset at that point). And this event is also called after effects were reloaded, but before they are fully initialized (and as such this is not called again when additional effects are initialized because a new technique was enabled and it is also not called after destroying previous effects).

FransBouma:
That indeed sounds like the right event to subscribe to instead! :)

I'll close this PR as it's of no use :)

FransBouma:
@crosire the `reshade_set_current_preset_path` isn't raised when I click the 'Reload' button, so I need another way to determine when to migrate the handles... Looking at the code, the only event that will be raised in the code path when the Reload button is clicked is `reshade_reloaded_effects`, so I'll rewrite my code so it doesn't rely on techniques being disabled for migration.

PR #253, opened 2023-03-20
Removed preset saving from uniform variable setters, added save_current_preset to public API

FransBouma:
~~This commit adds `set_save_preset_on_api_uniform_change_state` which sets the new runtime flag `_save_preset_on_api_uniform_change` to the specified value. By default `_save_preset_on_api_uniform_change` is set to true to keep the API functionality the same with previous API versions. The flag `_save_preset_on_api_uniform_change` is used in the `set_uniform_variable_*` methods to allow for a way to switch preset saving off when uniform variables are changed.~~

~~API methods have been added at the bottom to keep API interfaces binary compatible. The `_save_preset_on_api_uniform_change` member variable is also added at the bottom of the runtime type for this reason.~~

~~I've not added a get_... variant as by default the flag is true and the general use case is that it's switched off only temporarily in a tight loop or single event handler.~~

As per discussion in Discord, I've reworked this PR to lift `save_current_preset` to the public API and remove the calls to `save_current_preset` in the `set_uniform_value_*` methods. 
I haven't lifted `load_current_preset` to the public API because it looks like it requires extra code around it in most current use cases which would require a wrapper method with almost the same name; additionally, it's also already covered for most use cases with `set_current_preset_path`.

I've added remarks to the `set_uniform_value_*` methods that the preset isn't saved after setting the uniform, and to do so the developer should call `save_current_preset`. 

I didn't increase the API version number as v7 hasn't been released yet.

### DIFF
--- a/include/reshade_api.hpp
+++ b/include/reshade_api.hpp
@@ -660,5 +660,12 @@ namespace reshade::api
 			size_t length = SIZE;
 			get_technique_effect_name(technique, effect_name, &length);
 		}
+
+		/// <summary>
+		/// Enables or disables the flag whether the current preset file is saved after a uniform variable was changed through the api using one of the set_uniform_variable_* methods. 
+		/// </summary>
+		/// <param name="save_preset">Set to <see langword="true"/> to enable preset file saving after a uniform variable has been changed through the api or <see langword="false"/> to disable it.</param>
+		///	<remarks>By default the flag is set to true. It's recommended to set this flag to <see langword="false"/> if you set uniform variables in a loop, or multiple times in a handler to the reshade_present event</remarks>
+		virtual void set_save_preset_on_api_uniform_change_state(bool save_preset) = 0;
 	};
 }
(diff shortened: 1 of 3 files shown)


seri14:
I thought it would be better to change it to save the preset at once when end of frame? :thinking:
I am trying to bring this idea to preprocessor.

```cpp
void set_uniform_value_data(...)
{
    ...
    // std::memcpy
    _flag = true;
}
```

```cpp
void on_present()
{
    ...
    // Maybe should complete most of the events before batch?
    _should_save_screenshot = false;

    if (_flag)
        save_current_preset();

    // Some process for next frame
    if (!_ignore_shortcuts && _input != nullptr)
    ...
}
```

FransBouma:
@seri14 you mean, regardless of the flag I added? As I don't want *any* saving of the preset file based on the changes I made (as I write every frame new values to the uniform variables as I interpolate between two states over a camera path). 

I don't think saving the preset file in the preset call chain is that great, as it's performance heavy (potentially, it's I/O after all) and present is a function called on a hot path. 

If an addon calls a uniform variable change outside present, it still should be able to save the preset file if it wants to I think.

seri14:
Ah, the transition should run on memory. It can only be solve by crosire.
so forget my proposal.

FransBouma:
@seri14 Crosire wanted to have the save preset method to be used by addon devs and to have the save in the uniform setters removed, so I reworked the pr to do that. I think this also meets your requirement to have a public way to save the preset without problems?

PR #254, opened 2023-03-24
Fixed reload effects run twice if effect files are missing for the preset

seri14:


crosire:
Should be fine, since this check was specifically added for the option to only load enabled effects and ReShade always holds an entry in `_effects` for all effects it can theoretically load (because found in the search paths).

PR #259, opened 2023-06-28
Add protection against key repeats triggering keybinds

Kaldaien:
For games that run their message loop on a different thread than presents the SwapChain, key repeats are currently altering the state of the `_keys[]` array after `reshade::input::next_frame (...)` runs.

<h2>This nullifies the logic that twiddles the bits here:</h2>

```cpp
	for (uint8_t &state : _keys)
		state &= ~0x08;
```

Thus, when the UI goes to sample `reshade::input::is_key_pressed (...)`, it sees **0x88** caused by key repeats and all keybinds in ReShade will randomly fire off multiple times causing the UI to toggle rapidly, thousands of screenshots to be taken, multiple copies of code saved when <kbd>S</kbd> is pressed, etc.

<h3>Simply holding down the <kbd>Home</kbd> key in the Microsoft Store version of Persona 5 currently results in this:</h3>

https://github.com/crosire/reshade/assets/13346629/8c70490a-47f1-42b8-b820-0a36dccc6b53

<hr>

I am fairly certain that `is_key_pressed (...)` should be excluding key repeats (that's what all this bit-twiddling is for, after all), so I have added logic to protect against this:

Each time `next_frame (...)` runs, the array of key states (as ***viewed*** during the last frame) is backed-up, and a new function (`is_key_repeated (...)`) checks whether the key is down between two successive frames. If it is, key repeats will be ignored and there is zero chance of triggering keybinds for key presses already processed on the last frame.

This changes keyboard input to behave more like gamepad input currently does, comparing a snapshot of the state from the last frame against the current state when checking for keys newly pressed.

<h3>Fixed Behavior (<kbd>Home</kbd> pressed once and held):</h3>

https://github.com/crosire/reshade/assets/13346629/d21dea9e-bad3-4ddf-bf3e-0efed107facb

### DIFF
--- a/source/input.cpp
+++ b/source/input.cpp
@@ -271,7 +271,7 @@ bool reshade::input::is_key_down(unsigned int keycode) const
 bool reshade::input::is_key_pressed(unsigned int keycode) const
 {
 	assert(keycode < ARRAYSIZE(_keys));
-	return keycode > 0 && keycode < ARRAYSIZE(_keys) && (_keys[keycode] & 0x88) == 0x88;
+	return keycode > 0 && keycode < ARRAYSIZE(_keys) && (_keys[keycode] & 0x88) == 0x88 && !is_key_repeated(keycode);
 }
 bool reshade::input::is_key_pressed(unsigned int keycode, bool ctrl, bool shift, bool alt, bool force_modifiers) const
 {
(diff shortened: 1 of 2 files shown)


crosire (approved):
Makes sense to me!

PR #263, opened 2023-07-11
Add *.spv files to .gitignore and delete all *.spv files

EndlesslyFlowering:
title


crosire:
They are actually intentionally included, because you can build ReShade without having the Vulkan SDK installed this way (this had been a pain point for several people attempting to build ReShade in the past). The Vulkan headers are taken from the submodule then. But the GLSL compiler can only exist after installing the Vulkan SDK, so there is no way to generate the SPIR-V files without that. As such the SPIR-V files are included in the repository.
But the GLSL compiler includes a version number in the SPIR-V header, meaning with updated Vulkan SDKs the contents of the SPIR-V files change every so slightly and they will show up in Git ...

Adding them to gitignore and not deleting won't help unfortunately, as gitignore doesn't apply to already indexed files. So the only good option I'm aware of is to make Git ignore them in your local repository via `git update-index --skip-worktree res/shaders/imgui_vs.spv res/shaders/imgui_ps.spv`. That way they don't constantly show up when building with a newer GLSL compiler. I'm not aware of a way to force this through the remote repository though.

EndlesslyFlowering:
Ah I see.
Gonna close this then :)

PR #264, opened 2023-07-12
Add ability to describe sliders and "drag sliders"

EndlesslyFlowering:
in action:
image

### DIFF
--- a/source/runtime_gui.cpp
+++ b/source/runtime_gui.cpp
@@ -3204,12 +3204,17 @@ void reshade::runtime::draw_variable_editor()
 					const int ui_max_val = variable.annotation_as_int("ui_max", 0, ui_type == "slider" ? 1 : std::numeric_limits<int>::max());
 					const int ui_stp_val = std::max(1, variable.annotation_as_int("ui_step"));
 
+					// Append units
+					std::string format = "%d";
+					const std::string_view units = variable.annotation_as_string("ui_units");
+					format.append(units);
+
 					if (ui_type == "slider")
-						modified = imgui::slider_with_buttons(label.data(), variable.type.is_signed() ? ImGuiDataType_S32 : ImGuiDataType_U32, data, variable.type.rows, &ui_stp_val, &ui_min_val, &ui_max_val);
+						modified = imgui::slider_with_buttons(label.data(), variable.type.is_signed() ? ImGuiDataType_S32 : ImGuiDataType_U32, data, variable.type.rows, &ui_stp_val, &ui_min_val, &ui_max_val, format.c_str());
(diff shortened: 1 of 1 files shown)


source/runtime_gui.cpp
```diff
@@ -3204,12 +3204,28 @@ void reshade::runtime::draw_variable_editor()
 					const int ui_max_val = variable.annotation_as_int("ui_max", 0, ui_type == "slider" ? 1 : std::numeric_limits<int>::max());
 					const int ui_stp_val = std::max(1, variable.annotation_as_int("ui_step"));
 
+#define DESCRIPTION_DEFAULT_STRING "%d"
+					const size_t length_of_description = sizeof(DESCRIPTION_DEFAULT_STRING) - 1;
+					const size_t max_description_length = 11; // space + 10 characters
+					const size_t max_chars_for_description = max_description_length + 1; // +1 for \0
+					char description[max_chars_for_description] = DESCRIPTION_DEFAULT_STRING;
```
crosire, line 3211: At this point might as well just make these `std::string`s and avoid the offset and copy logic.

source/runtime_gui.cpp
```diff
@@ -3204,12 +3204,28 @@ void reshade::runtime::draw_variable_editor()
...
+#undef DESCRIPTION_DEFAULT_STRING
+
+					const std::string_view description_view = ui_type == "slider" ? variable.annotation_as_string("ui_slider_desc")
+															: ui_type == "drag"   ? variable.annotation_as_string("ui_drag_desc")
+															: "";
```
crosire, line 3216: I'd consider naming the annotation `ui_units` uniformly, since that's what it should most likely be used for.

source/runtime_gui.cpp
EndlesslyFlowering: Thought about that too.
Though having a limit on the maximum characters seems reasonable to me. Let me know what you think about that.

source/runtime_gui.cpp
EndlesslyFlowering: yeah makes sense

EndlesslyFlowering:
I implemented the changes.
After some thinking: not putting a limit on how many characters you can put is OK since it is up to the shader developter to make it look good.
Though it probably can be abused? I don't know how many characters imgui can handle there and if that can cause issues if the string is too long.
Let me know how you want to handle this one.

crosire:
I wouldn't worry too much about potential for abuse in this case, the same could be said about most of the other UI options. Could even have a technique name with a thousand characters and mess things up that way. At the end of the day it's up to the user to be careful about what effect files they install, ReShade will just attempt to do what it is told to do.

PR #265, opened 2023-09-05
Add resolution filtering to the generic depth buffer addon

Liareth:
I face depth buffer detection woes all too often. The heuristics works 90% of the time then the entire scene flashes as you lose AO for a frame or two. This is especially true in Starfield where the heuristic fails from certain view angles in certain areas, likely caused by varying draw count due to frustum culling. In some scenes, it looks like the entire world is flashing as you pan your camera. This change allows you to (optionally) only consider depth buffers that match the specified resolution. It certainly fixes the issue for me in Starfield, and might be useful in other games which suffer similar problems.

image

### DIFF
--- a/examples/09-depth/generic_depth.cpp
+++ b/examples/09-depth/generic_depth.cpp
@@ -22,6 +22,10 @@ static bool s_disable_intz = false;
 static unsigned int s_preserve_depth_buffers = 0;
 // Enable or disable the aspect ratio check from 'check_aspect_ratio' in the detection heuristic
 static unsigned int s_use_aspect_ratio_heuristics = 1;
+// Enable or disable the resolution filter to only consider buffers that match the dimensions
+static unsigned int s_use_resolution_filtering = 0;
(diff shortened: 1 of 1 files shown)


crosire:
This is kind of the idea behind the existing aspect ratio heuristics already, so would be interesting to know why those cause to switch to a different buffer in your case (what resolution does it have in comparison to the main one?). Have you maybe not disabled "Use extended aspect ratio heuristics"?

BlueSkyDefender:
I can provide insight In my custom build I have two options.

image

The top one helps with this small switching issue in starfield.

The 2nd one helps in the new game DeadSpace when you need to use DLSS, FSR or other types if  scalers.

I do have to explain what each option does even when it's in the description text but, my users have gotten use to the modifications. 

I did want to add a resolution selection based on the buffers used. But, Realized it caused more issues then it solved so a combination of "Use extended aspect ratio" with a resolution check seemed to work out here.

But, the consequence of doing this it breaks if the user used Dynamic Resolution.

PR #268, opened 2023-09-18
Add option to manually prioritise draw calls over vertices

Dakraid:
Helps in games whose vertex count does not necessarily match the main scene, practical example would be Starfield where the actual depth buffer is the one with the higher draw call count

Option disabled:
Starfield_nj2U5W3Xdl

Option enabled:
Starfield_nCBncFKKGy

The order of the options in the depth plugin have been re-arranged in code, so it shows up first and not in between the others. I just forgot to recompile after making that change.


Dakraid:
Making following changes:
- Optimize the logic for selection
- Add secondary option to select buffer with highest draw calls and the highest vertices
- Set the Prioritise default to false to keep it in line with current behavior

Dakraid:
As discussed in Discord, this doesn't seem to be the right way ahead.
Closing this PR accordingly.

PR #270, opened 2023-10-10
Generic depth: resolve MSAA depth buffer if running Vulkan.

TheIronWolfModding:
This allows running depth based effects with MSAA on, tested via DXVK. VKL clean.

Thanks to DXVK guys for help

### DIFF
--- a/examples/09-depth/generic_depth.cpp
+++ b/examples/09-depth/generic_depth.cpp
@@ -207,7 +207,7 @@ struct __declspec(uuid("e006e162-33ac-4b9f-b10f-0e15335c7bdb")) generic_depth_de
 
 		desc.type = resource_type::texture_2d;
 		desc.heap = memory_heap::gpu_only;
-		desc.usage = resource_usage::shader_resource | resource_usage::copy_dest;
+		desc.usage = resource_usage::shader_resource | resource_usage::copy_dest | resource_usage::depth_stencil;
 
 		if (api == device_api::d3d9)
 			desc.texture.format = format::r32_float; // D3DFMT_R32F, since INTZ does not support D3DUSAGE_RENDERTARGET which is required for copying
@@ -236,6 +236,9 @@ struct __declspec(uuid("e006e162-33ac-4b9f-b10f-0e15335c7bdb")) generic_depth_de
 
 		depth_stencil_backup &backup = depth_stencil_backups.emplace_back();
 		backup.depth_stencil_resource = resource;
+
+		if (device->check_capability(device_caps::resolve_depth_stencil))
+			desc.texture.samples = 1; // Backup texture should either resolved or never created.
 
 		if (device->create_resource(desc, nullptr, resource_usage::copy_dest, &backup.backup_texture))
 			device->set_resource_name(backup.backup_texture, "ReShade depth backup texture");
@@ -444,7 +447,8 @@ static bool on_create_resource(device *device, resource_desc &desc, subresource_
 {
 	if (desc.type != resource_type::surface && desc.type != resource_type::texture_2d)
 		return false; // Skip resources that are not 2D textures
-	if (desc.texture.samples != 1 || (desc.usage & resource_usage::depth_stencil) == 0 || desc.texture.format == format::s8_uint)
+
+	if ((desc.texture.samples != 1 && !device->check_capability(device_caps::resolve_depth_stencil)) || (desc.usage & resource_usage::depth_stencil) == 0 || desc.texture.format == format::s8_uint)
 		return false; // Skip MSAA textures and resources that are not used as depth buffers
 
 	switch (device->get_api())
@@ -488,7 +492,7 @@ static bool on_create_resource_view(device *device, resource resource, resource_
 
 	const resource_desc texture_desc = device->get_resource_desc(resource);
 	// Only non-MSAA textures where modified, so skip all others
-	if (texture_desc.texture.samples != 1 || (texture_desc.usage & resource_usage::depth_stencil) == 0)
+	if ((texture_desc.texture.samples != 1 && !device->check_capability(device_caps::resolve_depth_stencil)) || (texture_desc.usage & resource_usage::depth_stencil) == 0)
 		return false;
 
 	switch (usage_type)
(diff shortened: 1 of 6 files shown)


source/vulkan/vulkan_impl_device.cpp
```diff
@@ -208,6 +208,8 @@ bool reshade::vulkan::device_impl::check_capability(api::device_caps capability)
 		return true;
 	case api::device_caps::shared_resource_nt_handle:
 		return !is_windows7();
+	case api::device_caps::resolve_depth_stencil:
+		return true;
```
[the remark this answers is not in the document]
crosire: It could also be implemented using render passes, instead of just dynamic rendering, but can just go with dynamic rendering only for now. It is definetely necessary to check support for that though, as e.g. there is no NV driver on Windows 7 that supports it. Can do this pretty easily:
```c++
case api::device_caps::resolve_depth_stencil:
    return _dynamic_rendering_ext;
```
Technically also need to check if the `VK_KHR_depth_stencil_resolve` extension is enabled or Vulkan 1.2 is used. ReShade only forces Vulkan 1.1, so that doesn't guarantee that this is available. But I can add that in a follow up.

source/vulkan/vulkan_impl_command_list.cpp
```diff
@@ -864,43 +865,169 @@ void reshade::vulkan::command_list_impl::resolve_texture_region(api::resource sr
...
+		rendering_info.layerCount = resolve_region.dstSubresource.layerCount;
+
+		if (src_subresource_info.aspectMask & VK_IMAGE_ASPECT_DEPTH_BIT)
+			rendering_info.pDepthAttachment = &depth_attachment;
+
+		if (dst_subresource_info.aspectMask & VK_IMAGE_ASPECT_STENCIL_BIT)
+			rendering_info.pStencilAttachment = &stencil_attachment;
+
+		vk.CmdBeginRendering(_orig, &rendering_info);
+		end_render_pass();
```
crosire, line 1029: This should simply be `vk.CmdEndRendering(_orig);` instead of `end_render_pass();` (which does more than is needed here).

source/vulkan/vulkan_impl_command_list.cpp
```diff
@@ -864,43 +865,169 @@ void reshade::vulkan::command_list_impl::resolve_texture_region(api::resource sr
...
+		VkImageView src_img_view;
+		if (vk.CreateImageView(_device_impl->_orig, &img_create_info, nullptr, &src_img_view)) {
+			assert(false);
+			return;
+		}
+
+		VkRenderingAttachmentInfo depth_attachment = { VK_STRUCTURE_TYPE_RENDERING_ATTACHMENT_INFO };
+		depth_attachment.imageView = src_img_view;
+		depth_attachment.imageLayout = VK_IMAGE_LAYOUT_DEPTH_STENCIL_ATTACHMENT_OPTIMAL;
+		depth_attachment.resolveMode = VK_RESOLVE_MODE_AVERAGE_BIT;
```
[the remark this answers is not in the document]
crosire: Implementations are only required to support `VK_RESOLVE_MODE_SAMPLE_ZERO_BIT`, `VK_RESOLVE_MODE_AVERAGE_BIT` is optional, so using it would require checking `VkPhysicalDeviceDepthStencilResolveProperties::supportedDepthResolveModes` (see https://registry.khronos.org/vulkan/specs/1.3-extensions/man/html/VkPhysicalDeviceDepthStencilResolvePropertiesKHR.html). I think it's easier to just use sample 0 to avoid that.

source/vulkan/vulkan_impl_command_list.cpp
```diff
@@ -864,43 +865,169 @@ void reshade::vulkan::command_list_impl::resolve_texture_region(api::resource sr
...
 
-	vk.CmdResolveImage(_orig,
-		(VkImage)src.handle, VK_IMAGE_LAYOUT_TRANSFER_SRC_OPTIMAL,
-		(VkImage)dst.handle, VK_IMAGE_LAYOUT_TRANSFER_DST_OPTIMAL,
-		1, &region);
+		img_create_info.image = src.handle;
+		img_create_info.subresourceRange = src_subresource_range;
+
+		VkImageView src_img_view;
+		if (vk.CreateImageView(_device_impl->_orig, &img_create_info, nullptr, &src_img_view)) {
```
crosire, line 998: This is a big nono and a huge memory leak. These resources are never freed, so right now every time this is called (every frame) additional memory is allocated and never cleaned up again.

ReShade does already create a default image view for depth-stencil images (with a subresource range that encompasses the entire image) however, which is fortunate, as that can be used in the case where no region is passed to `resolve_texture_region`. Support for region resolve would have to be dropped for now though.

You can get to the default views via `src_data->default_view` and `dst_data->default_view`.

examples/09-depth/generic_depth.cpp
```diff
@@ -786,6 +790,8 @@ static void on_begin_render_effects(effect_runtime *runtime, command_list *cmd_l
 	// Unlock while calling into device below, since device may hold a lock itself and that then can deadlock another thread that calls into 'on_destroy_resource' from the device holding that lock
 	lock.unlock();
 
+	const auto api = device->get_api();
```
crosire, line 793: This was intentionally not using `auto` to make the enum type clearer =P

examples/09-depth/generic_depth.cpp
```diff
@@ -237,6 +237,9 @@ struct __declspec(uuid("e006e162-33ac-4b9f-b10f-0e15335c7bdb")) generic_depth_de
 		depth_stencil_backup &backup = depth_stencil_backups.emplace_back();
 		backup.depth_stencil_resource = resource;
 
+		if (device->check_capability(device_caps::resolve_depth_stencil))
+			desc.texture.samples = 1; // Backup texture should either resolved or never created.
```
crosire, line 241: I don't think this needs to be conditional, can just always force sample count to 1 and move this to the rest of those setters above (line 210).

crosire (changes requested):

source/vulkan/vulkan_impl_command_list.cpp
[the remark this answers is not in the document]
crosire: `VK_RESOLVE_MODE_AVERAGE_BIT` can be used if necessary, but would need to verify if the device actually supports it first then and fall back to `VK_RESOLVE_MODE_SAMPLE_ZERO_BIT` if not.

crosire (approved):

PR #272, opened 2023-10-30
Programatic overlay control and add-on event for overlay activation

Kaldaien:
This introduces a new `addon_event::reshade_overlay_activation` that add-ons can subscribe to in order to watch for overlay activation and prevent activation state changes if necessary.

It also re-works the internal `_show_overlay` toggle in `reshade::runtime::draw_gui (...)` to use a new `reshade::runtime::activate_overlay (...)` function that is accessible to add-ons through `ReShadeActivateOverlay (...)`.

The new event differs from the existing overlay events in the add-on API in that it has nothing to do with extending ReShade's overlay, but instead allows add-ons to react to overlay activation or bring up ReShade's overlay on-demand.

<hr>

_Special K uses this in order to hide its own overlay and cease input manipulation while users are interacting with the ReShade UI, and to restore its overlay and input processing behavior once ReShade's UI is closed. This is the expected use-case._

### DIFF
--- a/include/reshade_api.hpp
+++ b/include/reshade_api.hpp
@@ -32,6 +32,18 @@ namespace reshade { namespace api
 	RESHADE_DEFINE_HANDLE(effect_uniform_variable);
 
 	/// <summary>
+	/// Input source for events triggered by user input.
+	/// </summary>
+	enum class input_source
+	{
+		none      = 0,
+		mouse     = 1,
+		keyboard  = 2,
+		gamepad   = 3,
+		clipboard = 4,
+	};
+
+	/// <summary>
 	/// A post-processing effect runtime, used to control effects.
 	/// <para>ReShade associates an independent post-processing effect runtime with most swap chains.</para>
 	/// </summary>
@@ -743,5 +755,15 @@ namespace reshade { namespace api
 		/// <param name="name">Name of the definition.</param>
 		/// <param name="value">Value of the definition.</param>
 		virtual void set_preprocessor_definition_for_effect(const char *effect_name, const char *name, const char *value) = 0;
+
+		/// <summary>
+		/// Activates ReShade's overlay programatically.
+		/// </summary>
+		/// <param name="activate">Activation state to request.</param>
+		/// <param name="source">Source of activation request.</param>
+		/// <remarks>
+		/// Returns <see langword="true"/> if the activation was accepted, otherwise returns <see langword="false"/>.
+		/// </remarks>
+		virtual bool activate_overlay(bool activate, reshade::api::input_source source) = 0;
 	};
 } }
--- a/include/reshade_events.hpp
+++ b/include/reshade_events.hpp
@@ -12,7 +12,7 @@ namespace reshade
 	enum class addon_event : uint32_t
 	{
 		/// <summary>
-		/// Called after successfull device creation, from:
+		/// Called after successful device creation, from:
 		/// <list type="bullet">
 		/// <item><description>IDirect3D9::CreateDevice</description></item>
 		/// <item><description>IDirect3D9Ex::CreateDeviceEx</description></item>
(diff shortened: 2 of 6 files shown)


include/reshade_events.hpp
```diff
@@ -1564,6 +1564,15 @@ namespace reshade
...
 		reshade_reorder_techniques,
 
+		/// <summary>
+		/// Called when the ReShade overlay is about to be activated (shown) or deactivated (hidden).
+		/// <para>Callback function signature: <c>bool (api::effect_runtime *runtime, bool* activate, api::input_source source)</c></para>
+		/// </summary>
+		/// <remarks>
+		/// To prevent overlay activation, set the value of activate and return <see langword="true"/>, otherwise return <see langword="false"/>.
+		/// </remarks>
+		reshade_overlay_activation,
```
crosire, line 1574: I'd rename this to `reshade_open_overlay`, so to be more consistent with a uniform source that exists doing something similar called `overlay_open` (which can be used to check whether the overlay is open in effects, https://github.com/crosire/reshade/blob/main/source/runtime.cpp#L1664).
And also make it behave more similar to other action events like `addon_event::draw`, in that returning true will make ReShade skip the open/close request, rather than requiring to modify a passed in parameter:
```c++
bool (api::effect_runtime *runtime, bool open, api::input_source source)
```

include/reshade_api.hpp
```diff
@@ -743,5 +755,15 @@ namespace reshade { namespace api
...
+
+		/// <summary>
+		/// Activates ReShade's overlay programatically.
+		/// </summary>
+		/// <param name="activate">Activation state to request.</param>
+		/// <param name="source">Source of activation request.</param>
+		/// <remarks>
+		/// Returns <see langword="true"/> if the activation was accepted, otherwise returns <see langword="false"/>.
+		/// </remarks>
+		virtual bool activate_overlay(bool activate, reshade::api::input_source source) = 0;
```
crosire, line 767: Similarily to the event, could then rename this `open_overlay()`. Makes it consistent with other events like `reshade_set_technique_state` or `reshade_reorder_techniques`, where the `effect_runtime` method name is the same as the event (apart from the added `reshade_` prefix) that is invoked when the action it triggers occurs.

source/runtime_gui.cpp
```diff
@@ -711,12 +711,20 @@ void reshade::runtime::draw_gui()
...
-			_show_overlay = false; // Close when pressing the escape button and not currently navigating with the keyboard
+			show = false; // Close when pressing the escape button and not currently navigating with the keyboard
 		else if (!_ignore_shortcuts && _input->is_key_pressed(_overlay_key_data, _force_shortcut_modifiers) && _imgui_context->ActiveId == 0)
-			_show_overlay = !_show_overlay;
+			show = !_show_overlay;
+
+		if (show != _show_overlay)
+		{
+			activation_source = api::input_source::keyboard;
+		}
```
crosire, line 727: I think you can avoid this branch by just defaulting `activation_source` to `keyboard` instead of `none` above, since `activate_overlay`/`open_overlay` below won't be called if the state doesn't change anyway.

source/runtime_gui.cpp
```diff
@@ -4370,4 +4384,27 @@ void reshade::runtime::destroy_imgui_resources()
 	_imgui_pipeline_layout = {};
 }
 
+bool reshade::runtime::activate_overlay(bool activate, reshade::api::input_source source)
+{
+	const bool requested = activate;
+
+	reshade::invoke_addon_event<reshade::addon_event::reshade_overlay_activation>(
```
crosire, line 4391: This ideally has protection against an add-on calling `activate_overlay`/`open_overlay` from within the `reshade_overlay_activation`/`reshade_open_overlay` event and also cannot be called when building ReShade without add-on support (also using the suggestion of using the return value to skip instead of a parameter):
```c++
bool reshade::runtime::open_overlay(bool open, api::input_source source)
{
#if RESHADE_ADDON
	if (!_is_in_api_call)
	{
		_is_in_api_call = true;
		const bool skip = invoke_addon_event<addon_event::reshade_open_overlay>(this, &open, source);
		_is_in_api_call = false;
		if (skip)
			return false;
	}
#endif

	_show_overlay = open;

	if (open)
		_imgui_context->NavInputSource = static_cast<ImGuiInputSource>(source);

	return true;
}
```

crosire (commented):
Looks good, just a have a little suggestion on the behavior of the event and naming:

crosire:
Merged with the aforementioned changes in bf0d1a9d7168af801aa55706b56e351d4cd3a307.

PR #275, opened 2023-11-27
Improve Simplified Chinese translation

liuxd17thu:
Proofread the Chinese translations:
* Fixed some mistakes like confusing left/right mouse button.
* Made some sentences smoother with the ones in my own fork before.

Todo:
* Use custom glyph range (probably GB2312) instead of `GetGlyphRangesChineseSimplifiedCommon()` for better coverage.

------
对着我之前的fork简单校对了一下，@shooterspps 供你参考。


lilidream (approved):
第179行：建议修改“字段”为“输入框”，“要修改快捷键，先点击输入框，然后按下新的快捷键，或按退格键移除快捷键。”

liuxd17thu:
Updated.
Now custom_glyph_range.hpp goes into res\fonts, and the resource file encoding is reverted back to UTF-8.

PR #277, opened 2023-11-30
Add Japanese translation

cot6:
Some additional contexts are included.


source/runtime_gui.cpp
```diff
@@ -1860,8 +1860,8 @@ void reshade::runtime::draw_gui_home()
 	}
 	else
 	{
-		ImGui::BeginChildFrame(ImGui::GetID("tutorial"), ImVec2(0, 175));
-		ImGui::TextWrapped(tutorial_text.c_str());
+		ImGui::BeginChildFrame(ImGui::GetID("tutorial"), ImVec2(ImGui::GetContentRegionAvail().x, std::min(ImGui::GetContentRegionAvail().y - ImGui::GetFrameHeight() - _imgui_context->Style.ItemSpacing.y, ImGui::CalcTextSize(tutorial_text.data(), tutorial_text.data() + tutorial_text.size(), false, ImGui::GetContentRegionAvail().x - _imgui_context->Style.FramePadding.x * 2).y + _imgui_context->Style.FramePadding.y * 2)));
+		ImGui::TextWrapped("%*s", tutorial_text.size(), tutorial_text.c_str());
```
crosire, line 1864: I kinda liked the look of the tutorial box not covering the full window. Is this really needed?

crosire (commented):
It looks like lang_ja-JP.rc2 is not encoded as UTF-8 (since Git doesn't recognize it as text). Could you resave it as UTF-8?

cot6:
The reason I handling this PR as a draft is that I added many contextual completion to the translated text which are missing in the original sentences.

The following are examples.
Regarding ID:22624, it is not a translation issue.
It is important to understand which feature is referred from the original sentence in working translation, but this feature has no name yet.

## Translation ID:3613

### English

This is the preset selection. All changes will be saved to the selected preset file.

Click on the '+' button to add a new one.
Use the right mouse button and click on the preset button to open a context menu with additional options.

### English from translated Japanese (below)

Select your preset here. Clicking on the button labeled the preset name opens a dialog box for selecting the preset file (.ini).
Clicking the '<' and '>' buttons opens the preset from the same folder as in sequence, without opening the dialog box.

ReShade's default behavior is always to automatically save the preset to a file each time you tweak the preset. Clicking the 'Auto Save On' button will turn off auto save and will not save the preset to a file each time you tweak it.
The '(picture:floppy)' button seems to simply perform a save operation, but that actually behaves differently in that it erases all entries before saving the current preset.
Also note that the effect variables which are referenced from the turned-off effects will be excluded in the saving process.

To create a new preset, click the '+' button.
The preset selector has a context menu, which pops up when you right-click it.

### Japanese

ここでプリセットを選択します。プリセット名が書かれたボタンをクリックすると、プリセットファイル(.ini)を選択するためのダイアログボックスが開きます。
'<'と'>'ボタンをクリックすると、ダイアログボックスを開くことなく、同じフォルダにあるプリセットを順送りで開きます。

ReShadeの既定の動作では、プリセットを調節するたびに自動でファイルに保存します。'自動保存をオン'ボタンをクリックすると、自動保存がオフになり、プリセットを調節してもファイルに保存されなくなります。
'(絵:フロッピー)'ボタンは単に保存の動作をするように見えますが、実際は保存先のプリセットファイルの内容をすべて消去するという点で動作が異なります。
また、オフ状態の画面効果が参照していたエフェクト変数は保存の対象外となるので注意して下さい。

新しいプリセットを作るには'+'ボタンをクリックして下さい。
プリセットセレクターにはコンテキストメニューがあり、右クリックをすると表示されます。

## Translation ID:22624

### English

Open base folder in explorer

### English from translated Japanese (below)

Open the session profile location

### Japanese

セッションプロファイルの場所を開く

## Translation ID:52454

### English

Save %s

### English from translated Japanese (below)

Save '%s' as image file.

### Japanese

'%s'を画像ファイルとして保存します。

crosire:
1. In the link you posted it is clarified further down below that UTF-8 with the resource compiler is in fact working correctly now. That's why line https://github.com/crosire/reshade/blob/main/res/resource.rc#L3 is there to enable it (see also https://learn.microsoft.com/en-us/windows/win32/menurc/pragma-directives). It works great for the other translations and is much preferred, since Git can actually diff it properly (there would be `working-tree-encoding` to solve that, but why complicate things when it works the simple way too).

2. I see, I had missed the `ImGui::CalcTextSize` part of the expression in a quick glance over it. Makes sense then, though the unfolded variant would be a lot more readable.

3. I have no problem with expanded information in the translation texts. Ideally these will eventually make themselves into the other translations as well, if needed for understanding, but that's future work.
Regarding 22624, I would've thought the translation to be similar to 33659, where you seem to include a mention of the Windows "Explorer". But you are familiar with the language and I'm not, so that's up to you.

4. I see you also addressed the problem with font atlas building not handling the case where language is set to the "System Default" correctly. Thanks!

cot6:
> In the link you posted it is clarified further down below that UTF-8 with the resource compiler is in fact working correctly now. That's why line https://github.com/crosire/reshade/blob/main/res/resource.rc#L3 is there to enable it (see also https://learn.microsoft.com/en-us/windows/win32/menurc/pragma-directives). It works great for the other translations and is much preferred, since Git can actually diff it properly (there would be working-tree-encoding to solve that, but why complicate things when it works the simple way too).

See https://github.com/crosire/reshade/pull/277/commits/21a1f04f3ca86b3ffe13b2ddd156d194e7ed9193

> I see, I had missed the ImGui::CalcTextSize part of the expression in a quick glance over it. Makes sense then, though the unfolded variant would be a lot more readable.

See https://github.com/crosire/reshade/pull/277/commits/852050bb8449f28aa09049c9c3c96d25d59f254e

> I have no problem with expanded information in the translation texts. Ideally these will eventually make themselves into the other translations as well, if needed for understanding, but that's future work.
Regarding 22624, I would've thought the translation to be similar to 33659, where you seem to include a mention of the Windows "Explorer". But you are familiar with the language and I'm not, so that's up to you.

I think the "Open the location" button has only patterns of behavior that open the drive (using explorer), access to samba server (using explorer), and open web site (using the default browser).

Translation ID:33659 seems to actually mean "open the effect file location" and translation is "エフェクトファイルの場所を開く"

PR #278, opened 2023-12-02
Add UI font type 'Latin font'

cot6:
Changes

* Add UI font type 'Latin font'
* Prefer 'Morisawa BIZ UDGothic' instead of 'Microsoft Gothic' for Japanese (Available since Win10 1809)

Use 'Microsoft Gothic' (for Japanese and latin typefaces) and 'ProggyClean' (in the code editor)
Before
Use 'Morisawa BIZ UDGothic', 'Microsoft Gothic' and 'Cascadia Mono'
After


crosire:
- +1 for the changes to `reshade::imgui::font_input_box`, nice for it to show the actual default font used!
- I'm not fully convinced there is a need for a separate latin font however, it complicates settings (and is a bit awkward to translate/explain in some languages).
- The idea to default the editor font to something simpler than the global font is nice, would just stick to built-in ProggyClean always then though. If one wants a different font like Cascadia Mono, that can be configured by the user, it shouldn't be forced by default.
- I'm still in process of considering embeding a default font for the non-latin glyphs (with a very restricted character set, to not bloat binary size beyond reason), so to avoid having to rely on whatever is available in C:\Windows\Fonts. That would therefore make selection between possible candidates in C:\Windows\Fonts moot.

cot6:
> I'm not fully convinced there is a need for a separate latin font however, it complicates settings (and is a bit awkward to translate/explain in some languages).

These are easier to read. Most latin fonts has high quality design, so they look better than our language specific typefaces.
Also, some fonts are not designed for suitable for composition like section numbering.

Example: rendering like "1。2。3。"

> The idea to default the editor font to something simpler than the global font is nice, would just stick to built-in ProggyClean always then though. If one wants a different font like Cascadia Mono, that can be configured by the user, it shouldn't be forced by default.

Now that the code editor handles UTF-8, it would be nice to be able to display it by default.

Cascadia is bundled with Windows Terminal (maybe after 2019), so it is already installed on most PCs. However, it is understandable that it is not the common choice.

> I'm still in process of considering embeding a default font for the non-latin glyphs (with a very restricted character set, to not bloat binary size beyond reason), so to avoid having to rely on whatever is available in C:\Windows\Fonts. That would therefore make selection between possible candidates in C:\Windows\Fonts moot.

cot6:
> * I'm still in process of considering embeding a default font for the non-latin glyphs (with a very restricted character set, to not bloat binary size beyond reason), so to avoid having to rely on whatever is available in C:\Windows\Fonts. That would therefore make selection between possible candidates in C:\Windows\Fonts moot.

Something like Noto?
I think it would be better that the ReShade Setup installs these rather than embed to a DLL.

cot6:
> Maybe can make the latin font option only show up when actually needed then (when a language is active that uses non-latin glyphs)? That way it doesn't clutter the settings for languages where it isn't needed, but still is an option in the languages where it's useful. Something like:

If you approve `imgui.h` to be included in `runtime.hpp`,
```cpp
static_assert(sizeof(ImWchar) == sizeof(wchar_t));
bool latin_font_visible = wcscmp((const wchar_t *)_language_glyph_ranges, (const wchar_t *)_imgui_context->IO.Fonts->GetGlyphRangesDefault()) != 0;

if (latin_font_visible &&
    imgui::font_input_box(_("Latin font"), _default_latin_font_path.c_str(), _latin_font_path, _file_selection_path, _font_size))
```
otherwise
```cpp
// building atlas
ImFontConfig cfg;
strcpy_s(cfg.Name, "Global Font");

// draw_gui_settings
static_assert(sizeof(ImWchar) == sizeof(wchar_t));
if (ImFont **global_font = std::find_if(_imgui_context->IO.Fonts->Fonts.begin(), _imgui_context->IO.Fonts->Fonts.end(),
    [](const ImFont *font) { return strcmp("Global Font", font->GetDebugName()) == 0; });
    _imgui_context->IO.Fonts->Fonts.end() &&
    wcscmp((const wchar_t *)global_font[0]->ConfigData[(global_font[0]->ConfigDataCount - 1) - 1].GlyphRanges, (const wchar_t *)_imgui_context->IO.Fonts->GetGlyphRangesDefault()) != 0 &&
    imgui::font_input_box(_("Latin font"), _default_latin_font_path.c_str(), _latin_font_path, _file_selection_path, _font_size))
```

cot6:
> Yeah, just pruned down to less glyphs, so that maybe with some compression on top it's possible to get it within 2MB, at which point it could be acceptable to embed. Having the setup tool install it instead would be nice in theory, but it goes against the design decision of the ReShade binaries working independently (one doesn't have to use the setup tool to set ReShade up, it's just a convenience option, but ReShade itself configures itself just fine when installed manually as well).

I think you are overthinking in this case.
Show fonts list of community reviewed to users, and let users install these themselves.
Pairs of screenshots and URL would be enough.

However, most people should be able to use it as is the current build, and few people care enough about how it looks to make them unable to use ReShade.

crosire:
I think you don't actually have to store the glyph ranges separately. ImGui requires the glyph ranges pointer to be alive for the lifetime of a font (see comment on `ImFontConfig::GlyphRanges`, and the data is not copied during font building). `ImFontAtlas::GetGlyphRanges*()` return a static pointer which was used as is, meaning `ImFontConfig::GlyphRanges` will point to that same pointer. And since ReShade doesn't build glyph ranges on the fly, but only uses those static ones, a simple pointer comparison (`_imgui_context->IO.Fonts->Fonts[0]->ConfigData->GlyphRanges != _imgui_context->IO.Fonts->GetGlyphRangesDefault()`) *should* work and be enough to check whether the default range is used or not, without having to actually compare the contents.

> I think you are overthinking in this case.

Probably =D. I just want to have a simple default that "just works": drop in the DLL and you are done. Installing custom fonts, having to run the setup tool etc. makes that more evolved. Could of course argue that the default simply only supports languages with latin based script, but that's not nice either ... So I guess will just have to go with a list of fonts to try for now, as you've done here.

cot6:
> I think you don't actually have to store the glyph ranges separately. ImGui requires the glyph ranges pointer to be alive for the lifetime of a font (see comment on `ImFontConfig::GlyphRanges`, and the data is not copied during font building). `ImFontAtlas::GetGlyphRanges*()` return a static pointer which was used as is, meaning `ImFontConfig::GlyphRanges` will point to that same pointer. And since ReShade doesn't build glyph ranges on the fly, but only uses those static ones, a simple pointer comparison (`_imgui_context->IO.Fonts->Fonts[0]->ConfigData->GlyphRanges != _imgui_context->IO.Fonts->GetGlyphRangesDefault()`) _should_ work and be enough to check whether the default range is used or not, without having to actually compare the contents.

See https://github.com/crosire/reshade/pull/278/commits/22bf1911d8fdf1b4e4d8e5c55d9ed32e24aadde2

PR #279, opened 2023-12-04
Add localization for the two combos in ReShade GUI.

liuxd17thu:
Now '\n' is used for seperator when in string table, and is replaced by '\0' at runtime.

TODO:
BG/DE/JP translation for input processing selections and the "failed to compile" error. The strings are still English in these tables.
(The OSD selections are probably fine, since they are simple words.)


liuxd17thu:
> It is a curious update so I also tried it. This resource content `"T\0E\0S\0T\0\0\0\0"` (lang_*.rc2) is trimmed to `"T\0E\0S\0T"` by `LoadStringW`. (I have not checked the raw dll resource yet) My idea is that if the original sentence ends with `\0`, `load_string()` adds `\0` at the end. [949a1b9](https://github.com/crosire/reshade/commit/949a1b9bd3686de888bc002beb1a42f0f96a3c92)

 Another problem is that '\0' cannot be handled well with `xgettext`. 
I have also tried using '\a' or '\v' to distinguish from '\n' but `LoadStringW()` did not like that.

crosire:
Is a nice solution IMHO. Can't reasonably have multi-line entries in combo boxes anyway, but it makes sense to separate the entries into multiple lines from a translation perspective.

Btw., the nice thing about using `LoadString` is that you don't necessarily have to have every single string translated into all languages. If an entry is missing in the string table of the active language, Windows will try to find it in the closest other string table in the resources (if no other sublanguage is embeded, will fall back to the en-US one). So could technically also leave out the not yet translated entries in the other language files.

cot6:
> > It is a curious update so I also tried it. This resource content `"T\0E\0S\0T\0\0\0\0"` (lang_*.rc2) is trimmed to `"T\0E\0S\0T"` by `LoadStringW`. (I have not checked the raw dll resource yet) My idea is that if the original sentence ends with `\0`, `load_string()` adds `\0` at the end. [949a1b9](https://github.com/crosire/reshade/commit/949a1b9bd3686de888bc002beb1a42f0f96a3c92)
> 
> Another problem is that '\0' cannot be handled well with `xgettext`. I have also tried using '\a' or '\v' to distinguish from '\n' but `LoadStringW()` did not like that.

It is just a tool problem. I think it is enough that code and runs as intended.

liuxd17thu:
> \n is a nice solution IMHO. Can't reasonably have multi-line entries in combo boxes anyway, but it makes sense to separate the entries into multiple lines from a translation perspective.
> 
> Btw., the nice thing about using `LoadString` is that you don't necessarily have to have every single string translated into all languages. If an entry is missing in the string table of the active language, Windows will try to find it in the closest other string table in the resources (if no other sublanguage is embeded, will fall back to the en-US one). So could technically also leave out the not yet translated entries in the other language files.

Do we need a function detecting combo and handling the substitution, or just manually replacing at every ImGui::Combo? It seems not easy to distinguish '\n' seperated combo with other multi-line strings.

crosire:
What you have done here is fine with me, just waiting for the draft flag to be removed. There are other cases already where the translation string is modified before being passed to imgui functions.

liuxd17thu:
> Btw., the nice thing about using LoadString is that you don't necessarily have to have every single string translated into all languages. If an entry is missing in the string table of the active language, Windows will try to find it in the closest other string table in the resources (if no other sublanguage is embeded, will fall back to the en-US one). So could technically also leave out the not yet translated entries in the other language files.
I tried keeping only zh-CN translated, then other languages would fall back to zh-CN on my PC. Is the fallback based on system locale? (That's why I retain the English placeholders. )

crosire:
I meant that if you have the en-US string table with everything included, but the zh-CN string table is missing an entry, it will fall back to the en-US string for that entry. Search order is roughly:
1. Matching primary language + sublanguage
2. Matching primary language (first sublanguage in the resources)
3. Neutral
4. English (skipped if primary language is English)
5. First language in the resources

PR #280, opened 2023-12-05
Fix missing XML tags in API docs

Nukem9:
As the title says, since intellisense was unhappy. Thanks for the excellent documentation.

P.S.
The [addon event create_pipeline_layout](https://github.com/crosire/reshade/blob/79011b27cba8880923e838ee086a5615fc64a414/include/reshade_events.hpp#L693) appears to be 100% unused and the docs imply otherwise. It's not referenced in the source code as far as I can tell. Wasn't sure where to bring this up.

### DIFF
--- a/include/reshade_api_resource.hpp
+++ b/include/reshade_api_resource.hpp
@@ -311,7 +311,7 @@ namespace reshade { namespace api
 
 	/// <summary>
 	/// An opaque handle to a resource object (buffer, texture, ...).
-	/// <para>Resources created by the application are only guaranteed to be valid during event callbacks.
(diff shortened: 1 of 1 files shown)


crosire:
Yeah, the `create_pipeline_layout` event isn't really functional, since D3D9-11/GL use a fixed layout description (making the event useless, since it cannot be modified) and in D3D12/VK changing the description would add a lot of code (e.g. rebuilding the DXIL root signature blob) that it didn't feel worth it. So it mainly exists as a placeholder for the enum value, in case it is at some point implemented for D3D12/VK. Should probably be mentioned in the documentation that it isn't currently called though.

PR #281, opened 2023-12-05
Add French translation

AlucardDH:
I've started to translate to French. I will complete it tomorrow probably.

I may need some adjustments for the line 5 of the [res/lang_fr-FR.rc2] file (I see other languages use explicit value sand no constants).


liuxd17thu:
The script `tools/create_laguage_template.ps1` gives out `LANGUAGE 12, 1` for fr-FR.

AlucardDH:
I just added  13075 "Latin font" translation.

crosire:
Me, not actively having spoken French in several years, am by no means qualified, looks good overall.
Just noticed that 10646 sepaks of deleting the preprocessor, rather than deleting the preprocessor definition. Also TIL about espace insécable, interesting. There appears to be an unintended extra space in 53844 though.

PR #282, opened 2023-12-06
Add a button 'Save and complete adjustments'

cot6:
This button helps saving adjustments the depth settings as preprocessor definitions.
However, I do not understand why the author of [DisplayDepth.fx](https://github.com/crosire/reshade-shaders/blob/8ebf737b9f8ab55b129f7729c05c20f9a56d2369/Shaders/DisplayDepth.fx) banned the interface to change setting values.

### DIFF
--- a/source/runtime_api.cpp
+++ b/source/runtime_api.cpp
@@ -216,7 +216,7 @@ bool reshade::runtime::get_annotation_bool_from_uniform_variable([[maybe_unused]
 	{
 		const std::string_view ann_name = std::string_view(name);
 		if (const auto it = std::find_if(variable->annotations.cbegin(), variable->annotations.cend(),
-			[ann_name](const reshadefx::annotation &annotation) { return annotation.name == ann_name; });
+				[ann_name](const reshadefx::annotation &annotation) { return annotation.name == ann_name; });
 			it != variable->annotations.cend())
 		{
 			for (size_t i = 0; i < count; ++i)
@@ -237,7 +237,7 @@ bool reshade::runtime::get_annotation_float_from_uniform_variable([[maybe_unused
 	{
 		const std::string_view ann_name = std::string_view(name);
 		if (const auto it = std::find_if(variable->annotations.cbegin(), variable->annotations.cend(),
-			[ann_name](const reshadefx::annotation &annotation) { return annotation.name == ann_name; });
+				[ann_name](const reshadefx::annotation &annotation) { return annotation.name == ann_name; });
 			it != variable->annotations.cend())
 		{
 			for (size_t i = 0; i < count; ++i)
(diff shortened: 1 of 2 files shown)


crosire:
ReShade as a rule shouldn't know anything about potential effects being used with it, including that something called DisplayDepth.fx exists, nor that there are `RESHADE_DEPTH_INPUT_[...]` definitions (exception is the setup tool, since it does a little bit more game-specific setup). Core ReShade doesn't actually know what "depth" is at all, that is fully handled by a (albeit built-in) add-on nowadays.

The other additions/modifications I get behind (apart from there being some seemingly unnecessary formatting changes). The `string_view` ones should prevent a bunch of repeated `strlen` call, that's a good catch!

cot6:
> There could be an add-on that listens for changes to DisplayDepth.fx and applies them as preprocessor definitions automatically or something along those lines, but core ReShade shouldn't rely on knowledge of external effects IMHO.

By the way, is it possible that I create UI layout events of the Variable Editor, or/and implement that in the GenericDepth add-on?

crosire:
In the Generic Depth add-on maybe, though I feel like it goes beyond its scope too (it's already a nightmare to maintain in its current form). Trying to think of a useful event called from the variable editor that's not too niche, but could be useful in other instances too.
Maybe something that would also allow integrating custom widgets for variables? Like this (making it possible for an add-on to completely override the widget used to display a variable by returning true, or just add additional stuff before a variable by returning false):
```c++
...
case reshadefx::type::t_float:
{
	float data[16];
	get_uniform_value(variable, data, 16);

	is_default_value = std::memcmp(data, variable.initializer_value.as_float, variable.type.components() * sizeof(float)) == 0;

	if (!invoke_addon_event<addon_event::reshade_overlay_uniform_variable>(api::effect_uniform_variable { reinterpret_cast<uintptr_t>(&variable) }, data, sizeof(data), modified)
	{
		const float ui_min_val = variable.annotation_as_float("ui_min", 0, ui_type == "slider" ? 0.0f : std::numeric_limits<float>::lowest());
		...
		if (ui_type == "slider")
			modified = imgui::slider_with_buttons(label.data(), ImGuiDataType_Float, data, variable.type.rows, &ui_stp_val, &ui_min_val, &ui_max_val, precision_format.c_str());
		else if (ui_type == "drag")
		...
	}


	if (modified)
		set_uniform_value(variable, data, 16);
	break;
}
```
For this case in the add-on could then just look for a variable "iUIPresentType" in "DisplayDepth.fx" and add the save button in the event callback.

crosire:
Since the setup tool now has support for installing add-ons, this could even be made a required add-on (similar to how the "Standard effects" package is required). Just want to keep it separate from a code point of view.

crosire:
Added the proposed event in https://github.com/crosire/reshade/commit/1f246faa58094b4dcc666860b8d5a66a973fdbff.

PR #285, opened 2023-12-07
Display 'Home' as 'Home/Pos1' in German only

cot6:


crosire:
I had thought about doing this before, but ultimately had decided against it since language and keyboard layout are technically two different things. I can have my system language set to German, but still use a US keyboard (layout or even physical, is pretty common for developers for instance). So it's a bit misleading to tie the key names to the set language.
At the same time the previous code was also weird from the other perspective, where I had translated the key names to German when a German keyboard layout was active, even though everywhere else in the UI they were referenced by their English names ... Plus I wanted to get rid of code soly used for one singular language, now that ReShade is more localized, hence why https://github.com/crosire/reshade/commit/44326328ea0efa118167ccf6e6435662afae17de happened.
Need to think about this some more ...

PR #286, opened 2023-12-08
Add event 'reshade_overlay_uniform_variable'

cot6:
See my add-on side https://github.com/cot6/reshade-addons/compare/addon-adjustdepth#diff-ea12865d5d197112b8d8c060751e5a0dfdeb5571ccb29f1e7400ef97904052a2R40

### DIFF
--- a/source/ini_file.cpp
+++ b/source/ini_file.cpp
@@ -119,7 +119,7 @@ bool ini_file::save()
 
 	std::error_code ec;
 	const std::filesystem::file_time_type modified_at = std::filesystem::last_write_time(_path, ec);
-	if (!ec && modified_at > _modified_at)
+	if (!ec && (modified_at - _modified_at) >= std::chrono::seconds(2))
 		return false; // File exists and was modified on disk and therefore may have different data, so cannot save
 
 	std::stringstream data;
@@ -200,11 +200,12 @@ bool ini_file::save()
 	file.imbue(std::locale("en-us.UTF-8"));
 
 	const std::string str = data.str();
-	if (!file.write(str.data(), str.size()))
+	const bool fail = !(file.write(str.data(), str.size()) && file.flush());
+	
+	// Flush stream to disk before updating last write time
+	if (file.close(); fail)
 		return false;
 
-	// Flush stream to disk before updating last write time
-	file.close();
 	_modified_at = std::filesystem::last_write_time(_path, ec);
 
 	assert(!ec && std::filesystem::file_size(_path, ec) > 0);
--- a/source/ini_file.cpp
+++ b/source/ini_file.cpp
@@ -119,7 +119,7 @@ bool ini_file::save()
 
 	std::error_code ec;
 	const std::filesystem::file_time_type modified_at = std::filesystem::last_write_time(_path, ec);
-	if (!ec && (modified_at - _modified_at) >= std::chrono::seconds(2))
+	if (!ec && (modified_at - _modified_at) > std::chrono::seconds(2))
 		return false; // File exists and was modified on disk and therefore may have different data, so cannot save
 
 	std::stringstream data;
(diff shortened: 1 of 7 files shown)


crosire:
Oh, whoops, appears we did the same thing at the same time 🙃: https://github.com/crosire/reshade/commit/1f246faa58094b4dcc666860b8d5a66a973fdbff

crosire:
I didn't pass in the value itself in that implementation, since the add-on can also just query it via `runtime->get_uniform_value_[...]()`. But it might be a nicer interface to combine that approach with what you did, in the sense that the `bool &modified` parameter could be removed to instead check whether the data itself has changed.

source/runtime_api.cpp
```diff
@@ -1019,62 +1019,122 @@ void reshade::runtime::set_preprocessor_definition_for_effect([[maybe_unused]] c
...
+				if (const auto it = std::remove_if(preset_it->second.begin(), preset_it->second.end(),
+					[name = std::string_view(name)](const std::pair<std::string, std::string> &definition) { return definition.first == name; }); it != preset_it->second.cend())
+					preset_it->second.erase(it, preset_it->second.end()), updated = 0x010;
 		}
+		if (find_flag & 0b100)
+		{
+			if (const auto it = std::remove_if(_global_preprocessor_definitions.begin(), _global_preprocessor_definitions.end(),
+				[name = std::string_view(name)](const std::pair<std::string, std::string> &definition) { return definition.first == name; });
+				it != _global_preprocessor_definitions.end())
+				_global_preprocessor_definitions.erase(it, _global_preprocessor_definitions.end()), updated = 0x100;
```
crosire, line 1051: This was probably supposed to be `updated = 0b100;`, rather than `update = 0x100;` (same for the two above). Maybe it would help to make this an flags enum, for readability?
Btw., I find code becomes easier to read with some more curly braces and semicolons 😉:
```c++
if (const auto it = std::remove_if(_global_preprocessor_definitions.begin(), _global_preprocessor_definitions.end(),
		[name = std::string_view(name)](const std::pair<std::string, std::string> &definition) { return definition.first == name; });
	it != _global_preprocessor_definitions.end())
{
	_global_preprocessor_definitions.erase(it, _global_preprocessor_definitions.end());
	updated = 0b100;
}
```

source/ini_file.cpp
```diff
@@ -119,7 +119,7 @@ bool ini_file::save()
 
 	std::error_code ec;
 	const std::filesystem::file_time_type modified_at = std::filesystem::last_write_time(_path, ec);
-	if (!ec && modified_at > _modified_at)
+	if (!ec && (modified_at - _modified_at) > std::chrono::seconds(2))
```
crosire, line 122: Could you explain what prompted this change?

crosire (approved):

source/ini_file.cpp
cot6: See my 'screenshot' add-on https://github.com/cot6/reshade-addons/blob/c1c7c6cb1f7b94c6491ce226c69c7f467b51ff37/src/addon-screenshot/dllmain.cpp#L126-L140

This is a special behavior to avoid that effects bundled with this add-on are not loaded if 'Load only enabled effects' is enabled.

This add-on does load and modify the preset's contents before the host loading the effects.
However, the host has a cache of INI files, and if it detects changes written by another of these, the save will fail because of the condition preferring the latest changes.
The reason why 2 seconds rather than 1, because FAT32 can handle the last write time with 2-second precision.
And only having a cache of 0.5 seconds is too short.
This allows the add-on have an INI cache, and the host can still handle own successfully.

cot6: However, reviewing the code now, it appears to be dependent on Windows timing, it does not close the INI file before try to reload (set dummy preprocessor definition).
When I looked at it in Procmon while working on this issue, if it does not close the file byself, the syscall is called by System process, not the EXE.

PR #287, opened 2023-12-10
Add some specific behaviors for add-on developers can easily develop add-on controlled effects

cot6:
Changes
* Add 'nosave' annotation to exclude specific values when saving values to the preset
* Add '.addonfx' extension to always load the effect regardless 'Load only enabled effects' settings
* Changed the behavior of empty technique list after first edit


cot6:
Add-on side https://github.com/cot6/reshade-addons/compare/addon-adjustdepth...addonfx

crosire:
:+1:

Only thinking that the annotation could be more consistent with a different name, maybe `unsaved` or `ignored` (since there is a similar `hidden` annotation, rather than `noshow`, https://github.com/crosire/reshade-shaders/blob/slim/REFERENCE.md#uniform-variables)?

cot6:
The naming of `nosave` is certainly not correct.
The `nosave` has the following unintended behavior.

1. The variable cannot be reset from the UI.
2. The variable can only be controlled from add-ons.
3. The variable is never overwritten by reading presets.
4. The `nosave` is also use for not saving the enable/disable of a technique.
5. In others, The variable/technique works fine as usual.

Maybe `unsaved` and `ignored` is not intuitive. 
How do you think `unhandled`, `addon_controlled`, and `noevent`?.

crosire:
I like `unhandled` or `addon_controlled`. Maybe should even disable the widget in the variable list (`ImGui::Begin/EndDisabled`) for them so users don't change them if they are intended to be fully controlled via an add-on.

cot6:
> I like `unhandled` or `addon_controlled`. ~Maybe should even disable the widget in the variable list (`ImGui::Begin/EndDisabled`) for them so users don't change them if they are intended to be fully controlled via an add-on.~ Nevermind, that defeats your usecase I believe.

Yes. It is good to use 'hidden' annotation.
The unintented behavior "The variable cannot be reset from the UI" is to just protect self from being included in the bulk reset.

`noedit` might be good too.

cot6:
I will retrying to rename the annotation to something like `nosaved` or `unsaved` rather than including means 'handled' and 'controlled'.

crosire:
Having them as separate annotations is useful. I'm still not 100% happy with the naming though, but agree that it's less confusing than the other ideas, so maybe `disable_edit`, `disable_reset` and `disable_save` (basically introducing `disable_` as another special prefix, similar to the `ui_` annotations)?

cot6:
> Having them as separate annotations is useful. I'm still not 100% happy with the naming though, but agree that it's less confusing than the other ideas, so maybe `disable_edit`, `disable_reset` and `disable_save` (basically introducing `disable_` as another special prefix, similar to the `ui_` annotations)?

I referred to the names of enums from ImGui. If the ReShade naming does not match, feel free to decide yourself.

cot6:
I cannot list the prefix mean 'false' excluding `no`.
If you just want the scope, these are in the preset, so maybe we can use `preset_no`. (Add-ons cannot handling preset. These are just states.)

PR #288, opened 2023-12-16
Define the default latin fonts for the languages with non-latin typefaces

cot6:


cot6:
Added ASCII button to fallback to ProggyClean.
ascii_button_before
ascii_button_after

PR #291, opened 2023-12-17
Fix wrong value comparisons

cot6:


source/runtime_gui.cpp
```diff
@@ -3412,7 +3412,8 @@ void reshade::runtime::draw_variable_editor()
 			{
 			case reshadefx::type::t_bool:
 				get_uniform_value(variable, reinterpret_cast<bool *>(value.as_uint), variable.type.components());
```
crosire, line 3414: Best to not cast to boolean here (and line 3416 below), but rather simply get the integer value as-is and compare that against zero.

cot6:
By the way, why the bool variable can only handle a single component?

https://github.com/crosire/reshade/blob/5bb95d2dc26bce28be08d2bf797314b7ba9df1cb/source/runtime_gui.cpp#L3408-L3416

crosire:
Mainly because the widget used was a single checkbox, so can only handle a single component. But now that the `reshade_overlay_uniform_variable` event exists, it should definetely be extended to check all components for changes (since add-ons may show widgets with more than one component).

crosire (approved):

PR #294, opened 2023-12-19
Expand area of pop-up the tooltip

cot6:


crosire:
I believe this is also addressed by ee86e82de5d40211cd91355a651083a5dc3bbec7, unless I missed a widget?

cot6:
It seems to work as well in my side.

PR #297, opened 2023-12-22
Add UI type "check"

cot6:


crosire:
IMHO it would make more sense for the effect to be changed to use a `bool` uniform variable, rather than a `int` uniform variable here, instead of introducing a checkbox widget for integer uniform variables in ReShade.

cot6:
See https://github.com/crosire/reshade-shaders/pull/247/commits/d3bac8ad444eaaec147c6a3db7292b6f34de3687 of https://github.com/crosire/reshade-shaders/pull/247

PR #300, opened 2023-12-25
Fixed compile errors for the add-on marker on some targets

cot6:


crosire:
Didn't have access to GitHub at the time, so had fixed this in a3609e9eae69281197647af0cd360a35c35a52fe already.

PR #302, opened 2024-01-04
Added API 'reset_uniform_value' to allow add-ons can reset variables

cot6:
See my add-on side https://github.com/cot6/reshade-addons/commit/c398194608edfc621dbdf0478f990c2e37abcbad?diff=unified&w=1

### DIFF
--- a/source/runtime.hpp
+++ b/source/runtime.hpp
@@ -208,6 +208,7 @@ namespace reshade
 		void save_texture(const texture &texture);
 		void update_texture(texture &texture, uint32_t width, uint32_t height, uint32_t depth, const void *pixels);
 
+		void reset_uniform_value(api::effect_uniform_variable variable);
 		void reset_uniform_value(uniform &variable);
 
 		void get_uniform_value_data(const uniform &variable, uint8_t *data, size_t size, size_t base_index) const;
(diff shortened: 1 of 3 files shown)


source/runtime_api.cpp
```diff
@@ -1417,3 +1417,9 @@ bool reshade::runtime::open_overlay(bool /*open*/, api::input_source /*source*/)
 	return false;
 }
 #endif
+
+void reshade::runtime::reset_uniform_value([[maybe_unused]] api::effect_uniform_variable handle)
+{
+	if (auto variable = reinterpret_cast<uniform *>(handle.handle))
```
crosire, line 1423: This is missing a `#if RESHADE_FX` guard.

source/runtime.hpp
```diff
@@ -208,6 +208,7 @@ namespace reshade
 		void save_texture(const texture &texture);
 		void update_texture(texture &texture, uint32_t width, uint32_t height, uint32_t depth, const void *pixels);
 
+		void reset_uniform_value(api::effect_uniform_variable variable);
```
crosire, line 211: This needs to be defined above in the `public` section (where `set_uniform_value_bool` etc. are defined), otherwise the project will fail to compile when `RESHADE_FX` is not defined (due to the `reset_uniform_value` virtual not being implemented then).

PR #303, opened 2024-01-06
Always refer English if language fallback occurs

cot6:
This is for add-ons.
If the resource exists for desktop language but not equals with selected language or default, LoadString choices desktop one by language fallback. So ImGui will display characters outside of the glyph range.


crosire:
Mmh, realized this makes the "System Default" option useless, since it always falls back to English now. So I will probably need to revert this.

PR #304, opened 2024-01-06
Display instructions if compiling effects failed

cot6:
effect_errors

### DIFF
--- a/source/runtime_gui.cpp
+++ b/source/runtime_gui.cpp
@@ -2797,6 +2797,11 @@ void reshade::runtime::draw_gui_log()
 	std::filesystem::path log_path = global_config().path();
 	log_path.replace_extension(L".log");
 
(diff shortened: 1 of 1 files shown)


crosire:
Nice idea! I think I'll just move the text slightly, for consistency with the "Add-ons" page:
grafik

cot6:
I thinking about current situations of supporting users in Discord, what do you think about reimplement this with first composition?
Or/And show the modal like font scaling UI to show the complete error log?
Problems only occurs when user does something wrong. I think it is good to allow to ignore compile error by flag at least the runtime is alive.

crosire:
I don't think moving the message would help much, it's the red "failed to compile" that gets people either way. There could be an option to just ignore errors and hide effects that failed to compile, but I'm afraid that will lead to different problems with people wondering why an effect does not show up in their list and not realizing its because it failed to compile for some reason ...

PR #305, opened 2024-01-07
Always refer English if language fallback occurs (Re-challenge)

cot6:


cot6:
It should work now.

cot6:
> I don't quite understand what this solves which [dd7f252](https://github.com/crosire/reshade/commit/dd7f2525f214455095a68cf38cf39304b829ed15) did not address? E.g. when I remove a bunch of string table entries from the French resource file, set my system language to German and select French as language in the overlay, those missing entries are now shown in English (where before they were shown in German). The same thing is happening with this PR applied. Am I missing a use case?

That is because the glyph range is limited to selected language. For example, fallbacks to Japanese should not be displayed correctly. (like '????????????')
If you want to solve this, make `get_current_language()` returns multiple languages, rewrite the `build_font_atlas()` as a loop, and merge all matching languages.

cot6:
The language fallback cause by language, not by glyph range. These languages display correctly because there are uses same gryph in ReShade but LoadString is not know.

crosire:
But the latin glyph range is built into the font atlas for all languages, so even if the system language is Japanese and the user has selected Chinese in the overlay, if the add-on does not have Chinese resources it will fall back to English resources for which there is a glyph range.
It would only fail if it doesn't have English resources as well. But at that point I'd argue that's a fault of the add-on and it should at the very least have English resources for things to work properly.

cot6:
> But the latin glyph range is built into the font atlas for all languages, so even if the system language is Japanese and the user has selected Chinese in the overlay, if the add-on does not have Chinese resources it will fall back to English resources for which there is a glyph range. It would only fail if it doesn't have English resources as well. But at that point I'd argue that's a fault of the add-on and it should at the very least have English resources for things to work properly.

Well, it is probably a bug that the English resource does not exist.

Support the language drawing equivalent to Windows desktop application is requires merging all languages. It seems requires allocating a large amount of memory. So, I reduce the meaning of 'System Default' to 'Automatic Selection'.

cot6:
The code changed to expect cases that does not exist.
There is no need.

PR #307, opened 2024-01-12
Add variable annotation 'ui_category_toggle'

cot6:
See https://github.com/crosire/reshade-shaders/pull/247/commits/c3f1e5b20089054a6fd2a17459c62bba4b7a3e06 of https://github.com/crosire/reshade-shaders/pull/247

### DIFF
--- a/source/runtime_gui.cpp
+++ b/source/runtime_gui.cpp
@@ -3462,11 +3462,10 @@ void reshade::runtime::draw_variable_editor()
 						category_label += "###" + current_category; // Ensure widget ID does not change with varying width
 					}
 
-					if (variable.annotation_as_uint("ui_category_toggle") != 0)
+					if (category_visible = true;
+						variable.annotation_as_uint("ui_category_toggle") != 0)
 						get_uniform_value(variable, &category_visible);
-					else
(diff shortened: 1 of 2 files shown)


crosire:
As far as I can tell the additional `category_visible` variable outside the loop isn't needed and this can be simplified to just 7 added lines:
```c++
if (!current_category.empty())
{
>	if (bool category_visible = true;
>		variable.annotation_as_uint("ui_category_toggle") && (
>		get_uniform_value(variable, &category_visible, 1), !category_visible))
>	{
>		category_closed = true;
>	}
>	else
	{
		std::string category_label(get_localized_annotation(variable, "ui_category", _current_language));
		...
		category_closed = !ImGui::TreeNodeEx(category_label.c_str(), flags);
		...
	}
}
else
{
	category_closed = false;
}
```

cot6:
I feel bad to integrate the meanings within the category_closed annotation. But, permitted by you now, I will do this.

cot6:
I don't know how to avoid the first variable is invisible. I give up.

crosire:
Ah, whoops, I had missed that the variable itself would go invisible with my suggestion too.

PR #309, opened 2024-01-16
Fix various Vulkan tools (namely the SDK) complaining about the layer manifest being invalid

EndlesslyFlowering:
gets rid off this message when opening vkconfig from the Vulkan SDK:

vkconfig_2024-01-16_21-01-07

so now debugging via the SDK possible!

if the naming is unwanted you can change it obviously :)

hopefully this is not intentional 🤞

### DIFF
--- a/setup/ReShade32.json
+++ b/setup/ReShade32.json
@@ -1,7 +1,7 @@
 {
 	"file_format_version": "1.0.0",
 	"layer": {
-		"name": "VK_LAYER_reshade",
+		"name": "VK_LAYER_CROSIRE_reshade",
 		"type": "GLOBAL",
 		"library_path": ".\\ReShade32.dll",
 		"api_version": "1.3.268",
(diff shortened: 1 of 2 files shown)


crosire:
It isn't, but also is. The name was chosen before vkconfig started expecting a specific naming schema. But changing it now is difficult, since AFAIK there is software out there relying on the existing name to detect whether ReShade is loaded or not (e.g. X-Plane). Locally I'm just changing it during development, so that vkconfig doesn't complain, but haven't touched the files in the repository for these historic reasons.

EndlesslyFlowering:
> It isn't, but also is. The name was chosen before vkconfig started expecting a specific naming schema. But changing it now is difficult, since AFAIK there is software out there relying on the existing name to detect whether ReShade is loaded or not (e.g. X-Plane). Locally I'm just changing it during development, so that vkconfig doesn't complain, but haven't touched the files in the repository for these historic reasons.

I see 😄 

well time to close it then...

PR #311, opened 2024-01-28
Create lang_ko-KR.rc2

FriedG:
Add Korean Localization


res/lang_ko-KR.rc2
```diff
@@ -0,0 +1,195 @@
...
+39535 "저장 후 명령 창 숨기기"
+58617 "오버레이 & 스타일"
+11496 "언어"
+195 "튜토리얼 다시 시작"
+58246 "스크린샷 메시지 표시"
+58370 "사전 설정 전환 메시지 표시"
+54153 "사전 설정 전환 메시지가 표시되려면 사전 설정 전환 시간이 0이 아니어야 합니다."
+45894 """모든 이펙트 강제 로드"" 버튼 표시"
+45223 "그룹 이펙트 파일 트리 대신 탭으로 보기"
+761 "공통 스타일
```
crosire, line 116: One missing " at the end here

crosire (commented):
Thanks a lot for the translation!
Since this will also need adding loading of a font with support for Korean glyphs (I can do that in a follow up though), do you know what the default Korean font is on Windows? It looks like "Malgun Gothic" is preinstalled, but just want to confirm.

FriedG:
> Thanks a lot for the translation! Since this will also need adding loading of a font with support for Korean glyphs (I can do that in a follow up though), do you know what the default Korean font is on Windows? It looks like "Malgun Gothic" is preinstalled, but just want to confirm.

Hello, you are correct. The windows default font for Korean is "Malgun Gothic". Thank you for your interest in our Korean support :D

PR #312, opened 2024-02-02
Use sleep with constants to fix easily unstable framerate issue

cot6:


crosire:
To be fair, that example is not exactly how a good framerate limiter should look like anyway, it mainly exists for demonstration purposes =P

PR #314, opened 2024-02-12
Add create_pipeline_layout addon event

clshortfuse:
Adds event which allows addons to modify pipeline layouts (D3D12 Root Signatures).

### DIFF
--- a/include/reshade_api_pipeline.hpp
+++ b/include/reshade_api_pipeline.hpp
@@ -199,6 +199,14 @@ namespace reshade { namespace api
 	};
 
 	/// <summary>
+	/// Describes a pipeline layout.
+	/// </summary>
+	struct pipeline_layout_desc {
+		uint32_t count;
+		pipeline_layout_param *params;
+	};
+
+	/// <summary>
 	/// An opaque handle to a pipeline layout object.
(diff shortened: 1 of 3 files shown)


crosire:
How do you handle memory allocation for the layout parameters when changing the count? Similar to the shader replace example with allocation in `create_pipeline_layout` and subsequent freeing in `init_pipeline_layout`?

clshortfuse:
Uh... now I do. 😄 

The code is available here: https://github.com/clshortfuse/renodx/blob/9bf0a6fed4e0fec3d451b46ab3ea5a9b66676aa8/src/cp2077/addon.cpp

I don't have an issue with injecting into the compute pipeline during `bind_pipeline`, but the graphics pipeline isn't working at the moment. It's likely related to the `stage` that's being injected. On compute pipeline it's always `compute`/`all_compute` but graphics pipeline doesn't work that way. That's probably more a user-side issue than the changes to the reshade.

crosire:
@clshortfuse I modified the event signature slightly to `bool (api::device *device, uint32_t &param_count, api::pipeline_layout_param *&params)` instead of introducing an additional struct (4120e3a390590a6559fbfc809ddab87b182dca37).

clshortfuse:
@crosire That's very sensible. I'm not too experienced by-ref / by-value syntax. Thanks for the review, clean-up, and merge.

PR #316, opened 2024-03-29
Add the API to read all elements of value from ReShade configuration

cot6:
This PR also includes the following changes:
[Fixed an issue where the file may be cached by different paths separately](https://github.com/crosire/reshade/commit/b94b1ee604871ddea30260a855c51e5c3c781d78)

### DIFF
--- a/source/addon.cpp
+++ b/source/addon.cpp
@@ -53,35 +53,6 @@ bool ReShadeGetConfigValue(HMODULE, reshade::api::effect_runtime *runtime, const
 
 	const std::string section_string = section != nullptr ? section : std::string();
 	const std::string key_string = key != nullptr ? key : std::string();
-	std::string value_string;
-
-	if (!config.get(section_string, key_string, value_string))
-	{
-		*size = 0;
-		return false;
-	}
-
-	if (value == nullptr)
-	{
-		*size = value_string.size() + 1;
-	}
-	else if (*size != 0)
-	{
-		*size = value_string.copy(value, *size - 1);
-		value[*size] = '\0';
-	}
-
-	return true;
-}
-bool ReShadeGetConfigValue2(HMODULE, reshade::api::effect_runtime *runtime, const char *section, const char *key, char *value, size_t *size)
-{
-	if (size == nullptr)
-		return false;
-
-	ini_file &config = (runtime != nullptr) ? ini_file::load_cache(static_cast<reshade::runtime *>(runtime)->get_config_path()) : reshade::global_config();
-
-	const std::string section_string = section != nullptr ? section : std::string();
-	const std::string key_string = key != nullptr ? key : std::string();
 	std::vector<std::string> elements;
 	std::string value_string;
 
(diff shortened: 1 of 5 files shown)


cot6:
See add-on side: https://github.com/cot6/reshade-addons/compare/config-getsetarray

source/runtime_api.cpp
```diff
@@ -178,7 +178,7 @@ void reshade::runtime::get_uniform_variable_name([[maybe_unused]] api::effect_un
 		else if (*size != 0)
 		{
 			*size = variable->name.copy(value, *size - 1);
-			value[*size++] = '\0';
+			value[*size] = '\0';
```
crosire, line 181: Please don't change this behavior: When passing a `nullptr` for value the functions return the buffer size (including the null-terminator), so it is confusing if after passing in a buffer the behavior of the size parameter suddenly changes. It should still return the actual buffer size (including the null-terminator).

source/ini_file.cpp
```diff
@@ -11,7 +11,7 @@
 #include <shared_mutex>
 
 static std::shared_mutex s_ini_cache_mutex;
-static std::unordered_map<std::wstring, std::unique_ptr<ini_file>> s_ini_cache;
+static std::unordered_map<std::wstring, std::shared_ptr<ini_file>> s_ini_cache;
```
crosire, line 14: Simply keep this a `unique_ptr` and only ever add the canonical representation of a path the map (don't add multiple paths pointing to the same `ini_file` object). That will incur an additional lookup cost if the path passed to `load_cache`/`flush_cache` is not already canonical, since it first has to be transformed to that, but simplifies this cache.

source/addon.cpp
```diff
@@ -68,11 +68,48 @@ bool ReShadeGetConfigValue(HMODULE, reshade::api::effect_runtime *runtime, const
 	else if (*size != 0)
 	{
 		*size = value_string.copy(value, *size - 1);
-		value[*size++] = '\0';
+		value[*size] = '\0';
 	}
 
 	return true;
 }
+bool ReShadeGetConfigValue2(HMODULE, reshade::api::effect_runtime *runtime, const char *section, const char *key, char *value, size_t *size)
```
crosire, line 76: Returning the value list as a buffer of consecutive of null-terminated items is a nice solution, it would actually allow making this a drop-in change without having to introduce a new export! Since previously `ReShadeGetConfigValue` already returned the buffer size including the null-terminator, expanding it to be able to return a buffer with multiple null-terminators would not be a problem, nor would it break existing add-ons, since those would only operate on settings with scalar values (and therefore behavior would stay the same).
So easy to move the new implementation of `ReShadeGetConfigValue2` into `ReShadeGetConfigValue` and get rid of the added `ReShadeGetConfigValue2`.

source/runtime_api.cpp
cot6: I thought you might say that, but it is true.
No, it is safe. This commit ([No required to increment the pointer](https://github.com/crosire/reshade/pull/316/commits/aab2429aa1ea528283c0f36c8e62ee618609141a)) will not change this odd behavior.

The story begins when I noticed the behavior that calling this function repeated calls truncate the tail of the string by one byte.
In the size report, it returns the required size up to the null byte assignment, but in the assignment, it returns the size of the string excluding the null byte.
So what is the increment operator incrementing?
You should have written `(*size)++` (at [Change API functions returning a string to automatically null-terminate them again](https://github.com/crosire/reshade/blame/8986e2a814dc9d1acc6cf882061493fccb75f32a/source/addon.cpp#L40)). This is API error.
So, this commit will not change this odd behavior.

source/ini_file.cpp
cot6: I actually don't know about the behavior of unique and shared pointers, and I'm not sure what problem will be solved by these classes.
If the total costs of the syscalls by running `canonical()` in usually is so small, then I think it is fine your proposal.

source/addon.cpp
cot6: You can understand what I mean as an example how to write the list in the Win32 API's PrivateProfile functions and ImGui.
Arrays end with '\0' , and strings end with '\0', so if you expand to handle multiple elements, the terminator will be "\0\0", even if contains a single element.
This would change the specification of the size advisement, but maybe it is a good time to change it now, depending the how actual users of `get/set_config_value`?

cot6:
Currently no changes due to feedback.

source/addon.cpp
crosire: I had been coming from the assumption that the null-terminator was already included in the returned size, in which case having an additional one in there wouldn't really be a problem. Since that isn't actually the case, this does change behavior. Though looking through the list of add-ons the setup tool advertises, it doesn't look like any of them would break from such a change, so probably still ok. The other overloads of `get/set_config_value` use `std::from_chars`, which I believe can handle a null character at the end.

PR #317, opened 2024-04-12
Some changes to the behavior of preset saving.

liuxd17thu:
1. Now preset will be saved immediately after auto save is toggled on.

2. Differentiate "Save preset without cleaning" & "Save preset with cleaning". Now the behaviors are:
* Auto Save On: Clicking the button always cleans the preset and saves it.
* Auto Save Off: Clicking the button behaves the save as "auto save" (without cleaning). Use Shift + Click for a save with cleaning.

Sometimes we prefer keeping values for those optional or temporarily disabled effects. However save button always performs cleaning in earlier designs, even if auto save is disabled.

Besides, the icon is changed to `ICON_FK_REFRESH` for save without cleaning. I have not come up with a better icon so far.


crosire:
I'd just leave the icon untouched

liuxd17thu:
> I'd just leave the icon untouched

Updated.

cot6:
This makes more confusing.
It is not clear to me why the trim is switched on/off when toggled the autosave.

cot6:
Sorry, but the Japanese tutorial needs additional work because the explanation is outdated.

PR #319, opened 2024-04-30
Reduce/Optimize total size of installer/setup.

goodusername123:
Changes are documented in the commits attached.
No lossy operations where used only lossless.
Expending on what I said/noted in one of the commits: Deflate64 is only supported in newer versions of .Net or more specifically newer versions of [System.IO.Compression](https://github.com/dotnet/runtime/tree/main/src/libraries/System.IO.Compression), so If Reshade's setup is ever upgraded to a modern version of .Net Core then Deflate64 can be used instead of Deflate for a nice compression gain.
The update to 7za.exe is not 100% necessary but I felt it was fine to include it.

Tools/Things used:
- [IcoTools/IcoCrush](https://github.com/jtippet/IcoTools?tab=readme-ov-file#icocrush)
- [oxipng](https://github.com/shssoichiro/oxipng/)
- chm help file included in [7-Zip](https://www.7-zip.org/)


crosire:
ReShade is Windows Vista+, so that's certainly enough compatibility for the icon. .NET can't currently be updated unfortunately to retain that compatibility with older Windows versions.
7-zip 18.05 and 23.01 seem to compress to the same size with the added options though, so I'd skip the 7za.exe update since it doesn't really add value and just increases the Git repository site (albeit by a negligible amount).

PR #323, opened 2024-07-08
Show date in clock

BoYanZh:


crosire:
Was there a specific reason the date was needed? The clock is there to more easily keep track of play time, but I'm not sure what the date could help with, so trying to understand the motivation behind it.

BoYanZh:
Sometimes I find some random clips on my disk or in the discord. But if there is no clue in the filename or modified date, I have no way of knowing when the clip is recorded. I think the date may help.

source/runtime_gui.cpp
```diff
@@ -1283,7 +1283,19 @@ void reshade::runtime::draw_gui()
...
+					sprintf_s(temp_clock, "%02u:%02u", tm.tm_hour, tm.tm_min)
+					break;
+				case 1:
+					sprintf_s(temp_clock, "%02u:%02u:%02u", tm.tm_hour, tm.tm_min, tm.tm_sec)
+					break;
+				case 2:
+					sprintf_s(temp_clock, "%.4d-%.2d-%.2d %02u:%02u:%02u", tm.tm_year + 1900, tm.tm_mon + 1, tm.tm_mday, tm.tm_hour, tm.tm_min, tm.tm_sec);
+					break;
+			}
+			const int temp_size = ImFormatString(temp, sizeof(temp), "%s", temp_clock);
```
crosire, line 1298: This can cause a buffer overun if the clock format is not 0, 1 or 2 (which is possible, as this can be overriden in the config file). It's also not necessary to do two string formats here.
Since this is a pretty simple change, I'll just push a commit with this fixed directly.

BoYanZh: Thanks!

PR #324, opened 2024-08-16
Add API 'require_reload_effect' to set flag to reload effects

cot6:
See my add-on side https://github.com/cot6/reshade-addons/pull/6 

TODO

- [x] Supports to reload multiple effects
- [x] Abort if effect is broken (solved by add-on side)
- [x] Other error patterns (solved by add-on side)

### DIFF
--- a/include/reshade_api.hpp
+++ b/include/reshade_api.hpp
@@ -804,6 +804,6 @@ namespace reshade { namespace api
 		/// TODO
 		/// </summary>
 		/// <param name="effect_name">TODO</param>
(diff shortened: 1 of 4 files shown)


crosire:
Sufficient for me, I'd just consider renaming `require_reload_effect()` to e.g. `reload_effect_next_frame()` (technically doesn't necessarily happen the next frame if there are multiple effects to reload, but better describes what the function does anyway I think).

cot6:
The remaining required behaviors are quality issues on my add-on side.

PR #327, opened 2024-09-14
Fix linebreak performed by fgets() when loading .cube LUTs

liuxd17thu:
fgets() generates a string ending with '\n', then all the empty lines cannot be skipped.


crosire:
There was a bit more wrong then just what you had initially changed, so I just fixed the whole thing in one go.

PR #328, opened 2024-09-14
Halved the size of the spinner (back to the size it originally was)

CeeJayDK:
Halved the size of the spinner (back to the size it originally was)

It's now about the same size (42 x 42px) as a taskbar icon in Windows 10 (48x48px) which I feel is about the perfect size.

If I knew how I would also add a 1 or 2 second wait before it shows up, because if the compile time is THAT short, why even bother people with a message since it's near instantaneous anyways.

Maybe we should also add it to the initial message - since it is the logo after all.

### DIFF
--- a/source/runtime_gui.cpp
+++ b/source/runtime_gui.cpp
@@ -1141,7 +1141,7 @@ void reshade::runtime::draw_gui()
 #if RESHADE_FX
 		if (show_spinner)
 		{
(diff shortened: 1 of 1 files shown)


CeeJayDK:
Hmm hang on .. I think I might have forgotten to fix the width too.
I'll go check

CeeJayDK:
OK now I'm happy with the commit.

crosire:
Not sure I agree with this, it is awfully small now and heavily aliased (even if it matches the taskbar icon size).

CeeJayDK:
I'll see if I can make a compromise we are both happy with.

PR #330, opened 2024-10-14
Error occurs due to a filename containing characters which does not match Unicode with current codepage

cot6:
e.g. `reshade-shaders\Shaders\Deband - コピー.fx`
In the case of a Japanese user copied a file to the same folder using Windows Explorer.

### DIFF
--- a/source/runtime_api.cpp
+++ b/source/runtime_api.cpp
@@ -71,7 +71,7 @@ void reshade::runtime::enumerate_uniform_variables([[maybe_unused]] const char *
 
 	for (const effect &effect : _effects)
 	{
-		if (effect_name != nullptr && effect.source_file.filename() != effect_name)
+		if (effect_name != nullptr && effect.source_file.filename().u8string() != effect_name)
 			continue;
 
 		for (const uniform &variable : effect.uniforms)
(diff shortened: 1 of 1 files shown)


cot6:
I found some more bugs around this changes.

source/runtime_api.cpp
```diff
@@ -255,19 +262,20 @@ bool reshade::runtime::get_annotation_float_from_uniform_variable([[maybe_unused
 		values[i] = 0.0f;
 	return false;
 }
-bool reshade::runtime::get_annotation_int_from_uniform_variable([[maybe_unused]] api::effect_uniform_variable handle, [[maybe_unused]] const char *name, int32_t *values, size_t count, [[maybe_unused]] size_t array_index) const
+bool reshade::runtime::get_annotation_int_from_uniform_variable([[maybe_unused]] api::effect_uniform_variable handle, [[maybe_unused]] const char *name_in, int32_t *values, size_t count, [[maybe_unused]] size_t array_index) const
 {
 #if RESHADE_FX
-	if (const auto variable = reinterpret_cast<const uniform *>(handle.handle))
+	if (handle.handle != 0 && name_in != nullptr)
```
cot6, line 268: To avoid write like
```cpp
if (const auto variable = reinterpret_cast<const uniform *>(handle.handle) && name_in != nullptr)
```

PR #331, opened 2024-10-17
Use the recycle bin for user files rather than permanently deletion

cot6:


crosire:
I'm not convinced this is needed here: The setup tool does not touch user-generated files (like presets or textures), so the only files being deleted during uninstallation can also be restored by just installing ReShade again, so having them in the Recycle Bin doesn't add value as far as I can see.

crosire:
The setup tool explicitly **only** deletes *.fx and *.fxh files in the search paths relative to the application directory (see https://github.com/crosire/reshade/blob/main/setup/MainWindow.xaml.cs#L1411) and then the directories that are **empty** as a result of that (see https://github.com/crosire/reshade/blob/main/setup/MainWindow.xaml.cs#L1418). Image files such as "lut.png" are not touched, irrespective of their names.

PR #332, opened 2024-10-30
Add GetRawInputBuffer hook to handle games that do not use WM_INPUT

Kaldaien:
# Buffered Raw Input

I have added a new hook on `GetRawInputBuffer (...)` that classifies all input events as keyboard, mouse or gamepad and removes events from the buffer if ReShade is blocking that type of input.

This API does not use `WM_INPUT`, and cannot be captured through a game's message queue. It is intended for devices with high throughput where pulling `WM_INPUT` out of a window's message queue one by one is inefficient.

### On the topic of efficiency

I would really like to see implementations of `is_blocking_mouse_input (...)` and `is_blocking_keyboard_input (...)` that do not acquire a lock. The fast-path to exit `GetRawInputBuffer (...)` when ReShade is not blocking input (99% of the time during gameplay!) is crippled by potential lock contention on an API designed specifically to alleviate locking overhead on extremely high polling rate devices.

# Keyboard APIs (sometimes used for mouse input)

Hooks have also been introduced for `GetAsyncKeyState (...)`, `GetKeyState (...)` and `GetKeyboardState (...)`, since some games use these APIs for checking mouse button state or in lieu of processing keyboard/mouse messages.

# Cursor Position and Clipping Rectangle

I have changed mouse cursor blocking to get the immediate cursor position at the time blocking starts, rather than waiting for the game to call `SetCursorPos (...)` first.

Last, I added a check for an active window when ReShade blocks calls to `ClipCursor (...)`.

* If the calling thread has a window, instead of allowing the cursor to move to secondary monitors, the cursor will be restricted to the window. This makes finding the cursor easier on multi-monitor systems because it won't drift off to a different monitor and when the overlay is first opened, the cursor will automatically be moved inside the window.

This new behavior may be better as a user-configurable parameter, I'll leave that up to you however. _It is how Special K's own UI works (and is not configurable)._

### DIFF
--- a/source/input.cpp
+++ b/source/input.cpp
@@ -494,6 +494,8 @@ void reshade::input::immobilize_cursor(bool enable)
 			s_last_cursor_position.y = cursor_pos.y;
 		}
 	}
+
+	_immobilize_cursor = enable;
 }
 
 void reshade::input::block_mouse_input(bool enable)
@@ -512,11 +514,15 @@ void reshade::input::block_mouse_input(bool enable)
 		// Restore previous clipping rectangle when not blocking mouse input
 		ClipCursor(&s_last_clip_cursor);
 	}
+
+	_block_mouse = enable;
 }
 void reshade::input::block_keyboard_input(bool enable)
 {
 	if (enable)
 		_block_keyboard_time = std::chrono::high_resolution_clock::now();
+
+	_block_keyboard = enable;
 }
 
 bool is_immobilizing_cursor(reshade::input::window_handle target = reshade::input::any_window)
--- a/source/input.cpp
+++ b/source/input.cpp
@@ -776,10 +776,6 @@ extern "C" BOOL WINAPI HookClipCursor(const RECT *lpRect)
 	// Some applications clip the mouse cursor, so disable that while we want full control over mouse input
 	if (is_blocking_mouse_input() || is_immobilizing_cursor())
 	{
-		// A non-zero clipping rect would move the cursor, and may be a component of a game's mouselook.
-		if (lpRect != nullptr)
-			s_last_cursor_warp = std::chrono::high_resolution_clock::now();
-
 		lpRect = nullptr;
 	}
 
--- a/source/input.cpp
+++ b/source/input.cpp
@@ -457,9 +457,24 @@ std::string reshade::input::key_name(const unsigned int key[4])
 
 extern "C" BOOL WINAPI HookGetCursorPosition(LPPOINT lpPoint);
 
+void reshade::input::immobilize_mouse(bool enable)
+{
+	bool originally_immobilized = std::exchange(_immobile_mouse, enable);
+	if (enable)
+	{
+		// Update the initial cursor position as soon as blocking starts
+		static const auto trampoline = reshade::hooks::call(HookGetCursorPosition);
+		POINT cursor_pos;
+		if (!originally_immobilized && trampoline(&cursor_pos)) {
+			s_last_cursor_position.x = cursor_pos.x;
+			s_last_cursor_position.y = cursor_pos.y;
+		}
+	}
+}
+
 void reshade::input::block_mouse_input(bool enable)
 {
-	bool originally_blocked = std::exchange(_block_mouse, enable);
+	_block_mouse = enable;
 
 	// Some games setup ClipCursor with a tiny area which could make the cursor stay in that area instead of the whole window
 	if (enable)
@@ -468,14 +483,6 @@ void reshade::input::block_mouse_input(bool enable)
 		const RECT last_clip_cursor = s_last_clip_cursor;
 		ClipCursor(nullptr);
 		s_last_clip_cursor = last_clip_cursor;
-
-		// Update the initial cursor position as soon as blocking starts
-		static const auto trampoline = reshade::hooks::call(HookGetCursorPosition);
-		POINT cursor_pos;
-		if (!originally_blocked && trampoline(&cursor_pos)) {
-			s_last_cursor_position.x = cursor_pos.x;
-			s_last_cursor_position.y = cursor_pos.y;
-		}
 	}
 	else if ((s_last_clip_cursor.right - s_last_clip_cursor.left) != 0 && (s_last_clip_cursor.bottom - s_last_clip_cursor.top) != 0)
 	{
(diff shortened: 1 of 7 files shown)


Kaldaien:
# Mouse Immobilization

I have added an additional type of input manipulation to ReShade, a subset of `block_mouse_input (...)` that will only prevent the `Get/SetCursorPos (...)` APIs from altering anything.

When the cursor is immobilized (any time the overlay or traditional input blocking is active), all calls to `SetCursorPos (...)` are discarded, but the requested position is saved. Then when `GetCursorPos (...)` is called while blocking input or immobilizing the mouse, the last value the game tried to set the cursor pos to will be returned.

This goes above and beyond the old system, it allows interacting with UI elements like normal (typically games use `WM_MOUSEMOVE` for that), but prevents awful mouse warping when input blocking is set to only happen when the cursor is over the UI.

Kaldaien:
## Added (LowLevel) Keyboard/Mouse Hook Proxy code.
I added a fully working solution to block input to games that use Mouse/Keyboard hooks, adapted from Special K's code.

It is definitely not suitable for integration like this, but it shows how to accomplish this reliably. I have been using it in SK for 4 or 5 years now, never had any problems.

Kaldaien:
## Added DirectInput 8 support
Surprisingly this was absent and causing games like Sonic X Shadow Generations to receive keyboard and mouse input when they should not.

source/input.cpp
```diff
@@ -738,3 +831,777 @@ extern "C" BOOL WINAPI HookGetCursorPosition(LPPOINT lpPoint)
...
+{
+	static const auto trampoline = reshade::hooks::call(HookGetAsyncKeyState);
+	SHORT sKeyState = trampoline(vKey);
+
+	// Valid (Keyboard) Keys:  8 - 255
+	if ((vKey & 0xF8) != 0)
+	{
+		if (is_blocking_keyboard_input())
+		{
+			sKeyState = 0;
```
crosire, line 845: This could be simplified to return zero immediately, before even calling the trampoline function, àla:
```c++
if ((vKey & 0xF8) != 0 && is_blocking_keyboard_input())
    return 0;
if (vKey < 8 && is_blocking_mouse_input())
    return 0;

static const auto trampoline = reshade::hooks::call(HookGetAsyncKeyState);
return trampoline(vKey);
```

source/input.cpp
```diff
@@ -738,3 +831,777 @@ extern "C" BOOL WINAPI HookGetCursorPosition(LPPOINT lpPoint)
...
+{
+	static const auto trampoline = reshade::hooks::call(HookGetKeyState);
+	SHORT sKeyState = trampoline(vKey);
+
+	// Valid (Keyboard) Keys:  8 - 255
+	if ((vKey & 0xF8) != 0)
+	{
+		if (is_blocking_keyboard_input())
+		{
+			sKeyState = 0;
```
crosire, line 872: Same as with `HookGetAsyncKeyState`.

source/input.cpp
```diff
@@ -738,3 +831,777 @@ extern "C" BOOL WINAPI HookGetCursorPosition(LPPOINT lpPoint)
...
+				std::memset(lpKeyState, 0, 7);
+			}
+		}
+	}
+
+	return bRet;
+}
+
+// Adapted from Special K, but not needed since ReShade already processes WM_INPUT
+#if 0
```
crosire, line 924: Probably best to just remove this code block.

source/input.cpp
```diff
@@ -738,3 +831,777 @@ extern "C" BOOL WINAPI HookGetCursorPosition(LPPOINT lpPoint)
...
+    DWORD, HOOKPROC > _RealKeyboardProcs;
+           HOOKPROC   _RealKeyboardProc = nullptr;
+  concurrency::concurrent_unordered_map <
+    DWORD, HHOOK >    _RealKeyboardHooks;
+           HHOOK      _RealKeyboardHook = nullptr;
+} __hooks;
+
+//
+// Some typedefs because I cannot figure out what the requirements to actually
+// get this to build are, WinUser.h seems gimped by something :)
```
crosire, line 1187: This is because of https://github.com/crosire/reshade/blob/main/deps/Windows.props#L7, which reduces the amount of stuff pulled in via Windows headers. I think `NOWH` is the culprit specifically here, so that can just be removed from the props file and all the definitions should appear.

source/input.cpp
```diff
@@ -478,23 +517,47 @@ void reshade::input::block_keyboard_input(bool enable)
...
+		return std::any_of(s_windows.cbegin(), s_windows.cend(), predicate);
+	}
+
+	return s_immobilizing_global_cursor_move.load();
 }
-bool is_blocking_keyboard_input()
+bool is_blocking_mouse_input(reshade::input::window_handle target = reshade::input::GlobalQueue)
 {
-	const std::shared_lock<std::shared_mutex> lock(s_windows_mutex);
+	if (target != reshade::input::GlobalQueue)
```
crosire, line 536: I'm not sure I understand the need to distinguish `GlobalQueue` and `AnyWindow`. It seems to me that in both cases it would be sufficient to just check the global atomic bool?

source/input.cpp
```diff
@@ -704,9 +767,33 @@ extern "C" BOOL WINAPI HookClipCursor(const RECT *lpRect)
...
+	{
+		// Application tried to set a restrictive clip rect, expand it to the entire window if applicable
+		if (lpRect != nullptr)
+		{
+			lpRect = nullptr;
+
+			// Most of the time when the ReShade UI is open, users will want the mouse confined to the window.
+			// 
+			//   nb: Add a config parameter for this?
+			//     (Previous to this, ReShade was letting the cursor move to other monitors)
```
crosire, line 781: I do indeed use this behavior from time to time (e.g. having a code editor on my second monitor and moving back and forth between that and the game window on the main monitor), so would be in favor of leaving out this change.

crosire (commented):
Thanks a lot again! I have some small nitpicks and questions, but generally am good with all the additions. There are some differences in code style that I'd like to bring in line with the rest of the ReShade codebase =P, but I can do that in a follow up commit.

source/input.cpp
Kaldaien: The motivation for calling the function first originally was that this removes a bit indicating whether the input has been seen or not. But that aspect of `Get*KeyState` is unreliable anyway.

I'll change it.

source/input.cpp
Kaldaien: Yeah, that's a good point.

source/input.cpp
Kaldaien: I have reverted this.

PR #333, opened 2024-11-18
Initial working implementation of HDR screenshots (HDR10 and scRGB)

Kaldaien:
### I have written a small header-only library to handle HDR10 and scRGB screenshots using PNG.

* **The only special requirement for HDR10 PNG is an encoder capable of 16-bpc.**

Unfortunately, fpng and stb are both incapable of this. As such, the `sk_hdr_png` library uses WIC ***[(Windows Imaging Component)](https://learn.microsoft.com/en-us/windows/win32/wic/-wic-about-windows-imaging-codec)*** to encode 48-bpp PNGs and then adds additional PNG chunks necessary to declare the image as Bt.2020 + St.2084 (HDR10) and assist ICC-aware software to tone-map the HDR image to SDR.

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; To the best of my knowledge, there is no existing library that can do all of those things.

_I may continue development on a general-purpose HDR imaging library in the future, but what I have provided here is not that. It is a minimal implementation, unconditionally free for anyone to use and has questionable value beyond getting HDR image capture support into ReShade as simply as possible._
<br><br>&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; **NOTE:&nbsp;&nbsp;&nbsp;*The library uses DirectXMath and enables AVX intrinsics for select code paths.***

It is my strong belief that AVX is acceptable; it only applies to the hand-written SIMD code used for colorspace conversion and luminance frequency distribution, and that code only runs for users with HDR active.

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; *The odds that an HDR user has a CPU from the paleolithic era are sufficiently close to zero for me not to care* 😸
<br>
* ***Whenever a user is running a game in HDR, screenshots will default to HDR PNG.***

I have not implemented HDR->SDR tone-mapping that would be necessary to capture screenshots using either JPG or BMP.

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; I am open to implementing tone-mapping, if that is a thing that you want.

Special K's HDR->SDR tone-map implementation is a massive improvement over the garbage offered by Game Bar, so it would be a QoL improvement for users, but I can see this being considered bloat... ***in fact, I almost see that as something that should be implemented by a specialized Add-on.***

<hr>

### New config options have been added to:

1.  **Control HDR PNG quality**
 `SCREENSHOT.HDRBitDepth` (11-bit by default)
2.  **Copy captured screenshots to Clipboard**
 `SCREENSHOT.CopyToClipboard` (enabled by default)

The latter option makes sharing HDR screenshots in Discord and forums much simpler, allowing users to simply take a screenshot in ReShade and then paste... never having to fumble around in the screenshots directory.

<hr>

### I fixed one miscellaneous (***bug?***) while I was poking around in the screenshot code.

* Users may press the screenshot key multiple times in one second.

With the speed that `sk_hdr_png` is able to encode PNGs, assigning screenshots names that only contain Hours, Minutes and Seconds results in two rapid-fire screenshot triggers overwriting the previous screenshot.

If it simply overwrote the file that would be fine, but it tends to corrupt the file with a partial copy of both images!

* *I changed the default screenshot filename to **`"%AppName% %Date% %Time%_%TimeMS%"`***

&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;> _That was the simplest mitigation I could think of._

<hr>

### Miscellaneous Odds and Ends

I am unfamiliar with ReShade's string table for localization, so the new control panel settings are hard-coded in English.

I hope this meets constraints for "minimal bloat" (it is roughly 1000 lines of code and adds 10 KiB to the compiled release DLL, with no additional dependencies.)

I will be happy to make any changes necessary... I am adamant that this functionality should be core in ReShade, it will make life much easier for players of anti-cheat protected games who would otherwise have to run additional software.


deps/sk_hdr_png/include/sk_hdr_png.hpp
```diff
@@ -0,0 +1,993 @@
...
+
+#include <shellapi.h>
+#include <ShlObj.h>
+#include <wincodec.h>
+
+#pragma comment (lib, "windowscodecs.lib")
+
+namespace sk_hdr_png
+{
+#ifndef SK_HDR_PNG_RESHADE
```
crosire, line 35: Curious if the implementation is shared with something else or this was used for debugging?

deps/sk_hdr_png/include/sk_hdr_png.hpp
```diff
@@ -0,0 +1,993 @@
...
+		}
+
+		else
+		{
+			hr = E_OUTOFMEMORY;
+		}
+	}
+
+	if (SUCCEEDED(hr)) hr = bitmap_frame->Commit();
+	if (SUCCEEDED(hr)) hr = encoder->Commit();
```
crosire, line 978: This would crash if factory creation succeeded, but the `_aligned_malloc` allocations above failed. The `hr = E_OUTOFMEMORY` code path isn't attached to the right `if` statement to prevent that. Alternatively could just move these three `if (SUCCEEDED(hr))` blocks into the `if (png_buffer != nullptr && rgba32_scanline != nullptr && luminance_array != nullptr)` block like the others before.

source/runtime_gui.cpp
```diff
@@ -2266,6 +2266,16 @@ void reshade::runtime::draw_gui_settings()
 		modified |= ImGui::Checkbox(_("Save before and after images"), &_screenshot_save_before);
 #endif
 		modified |= ImGui::Checkbox(_("Save separate image with the overlay visible"), &_screenshot_save_gui);
+		//
+		// TODO: Add string table entries for this stuff...
```
crosire, line 2270: Localization works by wrapping strings in `_("...")` and then running `tools\create_language_template.ps1` with the various language identifiers ReShade has localization for (e.g. `tools\create_language_template.ps1 en-US`). That will update the language resources with the new strings, which can then be translated. But I can do that in post.

crosire (commented):
Nice, thanks! This looks like a reasonable solution to me.
If you do continue with a general purpose HDR imaging library shared with other projects, I'd be interested in picking that up as a dependency instead in the future.
A few comments:

deps/sk_hdr_png/include/sk_hdr_png.hpp
Kaldaien: That was for testing in a standalone app. I do want to eventually write something that can be shared with other projects, there's a severe lack of support for anything HDR-related at the moment.

deps/sk_hdr_png/include/sk_hdr_png.hpp
```diff
@@ -0,0 +1,993 @@
...
+
+	if (SUCCEEDED(hr))
+	{
+		WriteHDRChunks(image_path, luminance_array, width, height, format, quantization_bits);
+
+		if (copy_to_clipboard)
+			CopyToClipboard(image_path);
+	}
+
+	if (png_buffer      != nullptr) _aligned_free(png_buffer);
```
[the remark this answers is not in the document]
Kaldaien: You're right, it brings me comfort to always check for nullptr though :)

source/runtime.cpp
```diff
@@ -5015,17 +5036,33 @@ bool reshade::runtime::get_texture_data(api::resource resource, api::resource_us
...
+				// HDR10: Keep the original data, do not convert to 8-bpc
+				else
+					std::memcpy(pixels, mapped_pixels, pixels_row_pitch);
+				break;
+			case api::format::r16g16b16a16_float:
+				// FP16 is implicitly always scRGB
+				if (_back_buffer_format == api::format::r16g16b16a16_float &&
+				    _back_buffer_color_space != api::color_space::extended_srgb_linear)
+				{
+					_back_buffer_color_space = api::color_space::extended_srgb_linear;
```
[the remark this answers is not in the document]
Kaldaien: Yes, that can be removed. I was worried that having no colorspace assigned would cause issues with other things, as some software never calls `SetColorSpace1 (...)` for scRGB-based HDR. But it looks like everything handles the implicit colorspace correctly.

Kaldaien:
I apologize, I seem to have based this on the codebase before you merged my previous input-related changes.

I have had to add additional code to input.cpp to handle `WM_DISPLAYCHANGE` messages in order to invalidate display caching when users change the position of a monitor on their desktop, or turn displays on/off.

--
### I made all of the changes discussed to sk_hdr_png, and added display mastering metadata.

In order to implement the display mastering metadata, ReShade now keeps track of all of the displays on the Windows desktop and tracks changes to window position for on-screen runtimes.

This required the addition of a new api object, `reshade::api::display`.

It encapsulates `IDXGIOutput`, and keeps track of additional per-display details that DXGI does not support, such as refresh rate, the monitor's human-readable name, and critically, it allows checking if a display supports HDR and can enable/disable HDR for a display.

I also added an event that Add-ons can register for to be notified when the display changes. They will receive that notification when a runtime is initially created, when the window associated with a runtime moves across monitors, or when the display's settings change.

crosire:
If you don't mind, I would prefer to have the HDR screenshot changes separate from the display additions and instead move those to a new PR, to better distinguish the two features.

Kaldaien:
> If you don't mind, I would prefer to have the HDR screenshot changes separate from the display additions and instead move those to a new PR, to better distinguish the two features.

Sure, they're related in the sense that the HDR PNG screenshots aren't feature complete without the HDR display capabilities. But I guess I can comment that stuff out for now and re-visit this in another pull request.

PR #334, opened 2024-11-27
Allow add-on controlled effects to render regardless of effects state

cot6:
This work was to make my screenshot add-on always perform the depth capture.

.addonfx effects will no longer being affected by user choice directly, but you can still make it work like before by using the effects state API.

### DIFF
--- a/source/runtime.cpp
+++ b/source/runtime.cpp
@@ -1558,6 +1558,8 @@ bool reshade::runtime::load_effect(const std::filesystem::path &source_file, con
 		effect = {};
 		effect.source_file = source_file;
 		effect.source_hash = source_hash;
+
+		effect.is_addonfx = source_file.extension() == L".addonfx";
 	}
 
 	if (_effect_load_skipping && !force_load)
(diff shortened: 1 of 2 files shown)


source/runtime.cpp
```diff
@@ -1556,6 +1568,9 @@ bool reshade::runtime::load_effect(const std::filesystem::path &source_file, con
 		effect = {};
 		effect.source_file = source_file;
 		effect.source_hash = source_hash;
+
+		if (effect.source_file.extension() == L".addonfx")
+			effect.rendering = std::numeric_limits<unsigned int>::max();
```
crosire, line 1573: This effectively forces "*.addonfx" effects to always get uniform variables etc. updated, even if no technique inside has been enabled by the add-on (and thus it not actually rendering), which is not ideal.
I have attempted an alternative implementation in db5c3b8b31cb96c5079e4e0403f7e3780559a8d5, which should achieve the same goal, but without this drawback. I hope this solves this for you too?

cot6: Capture is working. However, we can identify the extension at the effect is loaded, so it may be better to create a member that `bool is_addonfx`.

PR #336, opened 2024-11-27
Add reshade::api::display in order to implement HDR display mastering metadata.

Kaldaien:
In order to finish HDR PNG screenshot integration, the capabilities of the display the image is captured on are required.

Thus, I have added `reshade::api::display` as a new api object to encapsulate display devices (its current implementation has a backing `IDXGIOutput` that exposes most of its functionality) and an event for AddOns to subscribe to so that they can be notified when display settings change or when the SwapChain's window moves to a different display.

Display Change is detected by checking the position of the SwapChain's window each frame for movement, and checking the associated monitor for the HWND on any frame where the window moves.

There is a broader indication of display change, sent to all top-level windows in the form of `WM_DISPLAYCHANGE`. It does not indicate which display has changed, but it is a signal that at least one display on the desktop has changed and any cached display data should be invalidated.

 * I handled `WM_DISPLAYCHANGE` from **input.cpp** since it is the only code in ReShade that is actually processing Window Messages, I hope that is not a problem.

### DIFF
--- a/include/reshade_api_display.hpp
+++ b/include/reshade_api_display.hpp
@@ -1,0 +1,153 @@
+/*
+ * Copyright (C) 2024 Patrick Mours
+ * SPDX-License-Identifier: BSD-3-Clause OR MIT
+ */
+
+#pragma once
+
+#include "reshade_api_device.hpp"
+#include <unordered_map>
+#include <memory>
+
+namespace reshade { namespace api
+{
+	/// <summary>
+	/// Describes quantities defined precisely as the ratio of two integers, such as refresh rates.
+	/// </summary>
+	struct rational
+	{
+		uint32_t numerator = 0;
+		uint32_t denomenator = 0;
+
+		constexpr float as_float() const { return denomenator != 0 ? static_cast <float>(numerator)/static_cast<float>(denomenator) : 0.0f; }
+	};
+
+	/// <summary>
+	/// Describes the colorimetry of a colorspace.
+	/// </summary>
+	struct colorimetry
+	{
+		float red [2] = { 0.0f, 0.0f };
+		float green [2] = { 0.0f, 0.0f };
+		float blue [2] = { 0.0f, 0.0f };
+		float white [2] = { 0.0f, 0.0f };
+	};
+
+	/// <summary>
+	/// Describes the dynamic range of a display or image.
+	/// </summary>
+	struct luminance_levels
+	{
+		float min_nits = 0.0f;
+		float max_nits = 0.0f;
+		float max_avg_nits = 0.0f;
+	};
+
+	/// <summary>
+	/// An output display.
+	/// <para>Functionally equivalent to a 'IDXGIOutput'.</para>
+	/// </summary>
+	struct __declspec(novtable) display : public api_object
+	{
+		/// <summary>
+		/// Indicates if cached properties are valid or potentially stale (i.e. refresh rate or resolution have changed).
+		/// </summary>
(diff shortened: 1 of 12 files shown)


include/reshade_api_display.hpp
```diff
@@ -0,0 +1,153 @@
...
+#include "reshade_api_device.hpp"
+#include <unordered_map>
+#include <memory>
+
+namespace reshade { namespace api
+{
+	/// <summary>
+	/// Describes quantities defined precisely as the ratio of two integers, such as refresh rates.
+	/// </summary>
+	struct rational
```
crosire, line 17: For consistency with other parts of the API, I'd forego this type. `swapchain_desc::fullscreen_refresh_rate` e.g. is a floating point value describing the refresh rate, not a rational type (and there is some logic at https://github.com/crosire/reshade/blob/main/source/dxgi/dxgi.cpp#L21 to convert between the two), so it would be strange if the display type suddenly expressed the refresh rate in a different manner.

source/dxgi/dxgi_impl_display.cpp
```diff
@@ -0,0 +1,317 @@
...
+		return api::color_space::srgb_nonlinear;
+	case DXGI_COLOR_SPACE_RGB_FULL_G10_NONE_P709:
+		return api::color_space::extended_srgb_linear;
+	case DXGI_COLOR_SPACE_RGB_FULL_G2084_NONE_P2020:
+		return api::color_space::hdr10_st2084;
+	case DXGI_COLOR_SPACE_RGB_FULL_G22_NONE_P2020:
+		return api::color_space::hdr10_hlg;
+	default:
+		return api::color_space::unknown;
+	}
```
crosire, line 143: You could include `d3d12/d3d12_impl_type_convert.hpp` and just call `return d3d12::convert_color_space(_desc.ColorSpace);` here to avoid some code duplication.

source/runtime.cpp
```diff
@@ -943,6 +947,82 @@ void reshade::runtime::on_present(api::command_queue *present_queue)
...
+						luminance_caps.min_nits, luminance_caps.max_nits, luminance_caps.max_avg_nits);
+#endif
+			}
+		}
+	}
+
+	else
+	{
+		log::message(log::level::warning, "Unable to map runtime %p's swap chain to a containing display output!", this);
+	}
```
crosire, line 1025: Can this be decoupled from the effect runtime and instead be put in the swapchain implementation, or maybe just into `dxgi/dxgi_swapchain.cpp` directly? There can technically be swap chains without an effect runtime initialized for them, in which case the events wouldn't fire.

source/dxgi/dxgi_impl_display.hpp
```diff
@@ -0,0 +1,132 @@
...
+		/// </summary>
+		virtual monitor get_monitor() const final;
+
+		/// <summary>
+		/// Gets the (GDI) device name (i.e. \\DISPLAY1) of the the monitor; do not use this as a persistent display identifier!
+		/// </summary>
+		/// <remarks>
+		/// This device name is not valid as a persistent display identifier for storage in configuration files, it may not refer to the same display device after a reboot.
+		/// </remarks>
+		virtual const wchar_t* get_device_name() const final;
```
crosire, line 40: The ReShade API uses UTF-8 strings everywhere else, so ideally these should be converted too. I'm still in favor of a `reshade::api::device::get_property` alike instead of all the individual getter functions, but maybe that's just me =P

PR #337, opened 2024-12-12
Some small GUI improvements

elbadcode:
These changes add a couple small usability settings to the GUI including machine translated localizations. There may also be a few unchanged lines that are just reformatted by the godawful intellisense features in VS, sorry if those snuck through, but I believe all changes are merged and I've tested everything thoroughly.

New checkbox item, "Escape closes overlay", enabled by default to preserve usage. When disabled you can only close the overlay with the open/close key, as there can be cases where you want the overlay to stay open but you need to press escape to close a game menu. I didn't want to mess with the input blocking modes so I just made it a separate option. It takes some getting used to and honestly is a bit annoying not to be able to press escape but I think its a necessary option to have. Possible improvement could be to add an option to hold escape to close but I think this is fine for now.

The other option sorts screenshots into folders by AppName which is great for people who save everything to a unified screenshot directory.  Previously I used a python script to do this post save but I think this is useful enough to add to the runtime and I've seen a few requests for it. I thought about also making Before/After folders if that option is selected and could go ahead and add that if desirable but I think for now this makes sense for a core runtime feature.

Users can put %AppName% in their screenshot name still (e.g. "Date Time AppName" works), however I made it so if they have either the new default or the old default option "AppName Date Time" it will remove the %AppName% and leading whitespace from the name whenever a screenshot is saved to ensure they don't get screenshots like "AppName/AppName" which is redundant and bad for organization (instead they get "AppName\Date Time Ms").  

Maybe this feels a bit too heavy handed to change the user's setting, but I think most people won't mind. While writing this I went ahead and changed it to use a new local variable so the actual config value won't be changed. This looks a bit awkward to have two local vars for screenshot name and there is a downside that it may not convey the change properly to users since it doesn't save, but it does ensure that the default setting will be preserved if they disable the option. If you like the feature enough to make it default to on we could also just have the setup tool change everyone's default screenshot path. Although in that case we might need to do the reverse and add the AppName to the screenshot name if this setting gets disabled. Lots of ways to do it and none of them feel quite right so its up to your judgement.

I could add something to the tutorial text or the macro string tooltip to explain this, actually I also tried to add a tooltip to this option but it appears blank. I'm sure I calculated the CRC hash correctly but maybe you just can't have a tooltip on a checkbox? It still shows a tooltip but its blank. The option is probably self explanatory enough so its not a big deal and if its not fixable we can just remove the tooltip.

Last change here is to the post save command feature. I made it accept file extensions for runnable scripts with specific handling for python and powershell scripts and anything else being run by calling "cmd /C ". Truthfully users could already do this themselves by running "C:\Windows\System32\cmd.exe" as their post save command and "/C <script path>" as their arguments but this makes it a bit more intuitive and clean. Up to you if you want it but I don't really see a reason not to since it doesn't actually add any new feature.

### DIFF
--- a/setup/MainWindow.xaml.cs
+++ b/setup/MainWindow.xaml.cs
@@ -559,6 +559,12 @@ namespace ReShade.Setup
 			DownloadCompatibilityIni();
 
 			string executableName = Path.GetFileName(currentInfo.targetPath);
+			if (compatibilityIni?.GetString(executableName, "Banned") == "1")
+			{
+				UpdateStatusAndFinish(false, "The target application is known to have blocked or banned the usage of ReShade. Cannot continue installation.");
+				return;
+			}
+
 			if (compatibilityIni != null && compatibilityIni.HasValue(executableName, "RenderApi"))
 			{
 				if (compatibilityIni.HasValue(executableName, "InstallTarget"))


source/runtime_gui.cpp
```diff
@@ -2280,14 +2285,16 @@ void reshade::runtime::draw_gui_settings()
 		modified |= ImGui::Checkbox(_("Save before and after images"), &_screenshot_save_before);
 #endif
 		modified |= ImGui::Checkbox(_("Save separate image with the overlay visible"), &_screenshot_save_gui);
-
+		modified |= ImGui::Checkbox(_("Split into folders by app name"),&_screenshot_path_split_appname);
+	if (ImGui::IsItemHovered(ImGuiHoveredFlags_ForTooltip))
+		{
+        ImGui::SetItemTooltip(_("Split save paths into folders by App Name.\nUseful if you have multiple games sharing the same screenshot directory"));
+		}
```
crosire, line 2292: With the above proposal this option would go away and instead would just add the same tooltip on the "Screenshot name" widget to the "Screenshot path" widget too.

source/runtime_gui.cpp
```diff
@@ -362,6 +362,7 @@ void reshade::runtime::load_config_gui(const ini_file &config)
 	config_get("INPUT", "KeyFPS", _fps_key_data);
 	config_get("INPUT", "KeyFrameTime", _frametime_key_data);
 	config_get("INPUT", "InputProcessing", _input_processing_mode);
+        config_get("INPUT", "EscClosesOverlay", _esc_closes_overlay);
```
crosire, line 365: This feels like an option that should be in the `OVERLAY` section, since it's not really a hotkey, more like `[OVERLAY] EscapeToClose=<0,1>`.

source/runtime_gui.cpp
```diff
@@ -2168,8 +2173,8 @@ void reshade::runtime::draw_gui_settings()
 				"Block all input when overlay is visible\n");
 			std::replace(input_processing_mode_items.begin(), input_processing_mode_items.end(), '\n', '\0');
 			modified |= ImGui::Combo(_("Input processing"), reinterpret_cast<int *>(&_input_processing_mode), input_processing_mode_items.c_str());
-
-			modified |= imgui::key_input_box(_("Overlay key"), _overlay_key_data, *_input);
+            modified |= ImGui::Checkbox(_("Escape Closes Overlay"),&_esc_closes_overlay);
```
crosire, line 2176: I'm not convinced this needs to be a GUI option, since it's very niche seems sufficient to just be a config file option only.

elbadcode: that's totally reasonable! my one concern is how to communicate that option to anyone (I'm sure I can't be the only one) who might have been frustrated by old behavior. I guess if its included in the changelog and shows up in the ini without having to be explicitly added that should be plenty. However in that case I do think it should stay as an input option rather than overlay. I would guess that anyone wanting this feature is using addons and has an overlay section a bit like below. Reason I say that is that REST already behaves like this whether intentional or not, which is also the reason I noticed the issue to begin with.

image

Personally I wouldn't be able to single out a new option in that section, here its easy to see and doesn't add to an already large section.

image

...Actually wait a second, the solution was already available in reshade, we just needed to make a behavior more consistent.

`if (_show_overlay && !_ignore_shortcuts && !_imgui_context->IO.NavVisible && _input->is_key_pressed(0x1B /* VK_ESCAPE */) &&
       ((_input_processing_mode == 2 || (_input_processing_mode == 1  &&  (_input->is_blocking_any_mouse_input() || _input->is_blocking_any_keyboard_input())))))	
        show_overlay = false;`
there's a shorter way to check if we're

source/runtime.cpp
```diff
@@ -4794,31 +4795,64 @@ static std::string expand_macro_string(const std::string &input, std::vector<std
...
+  const unsigned int screenshot_count = _screenshot_count;
+  std::filesystem::path _current_screenshot_path;
+    if (_screenshot_path_split_appname) {
+    std::string _screenshot_name_split =
+        (_screenshot_name == "%AppName% %Date% %Time%_%TimeMS%" ||
+         _screenshot_name == "%AppName% %Date% %Time%")
+            ? "%Date% %Time%_%TimeMS%"
+            : _screenshot_name;
+	 // This will remove the appname and leading whitespace if people have not changed their target path but select this option. Also updates to new default
+        // if (_screenshot_name == "%AppName% %Date% %Time%_%TimeMS%" || _screenshot_name == "%AppName% %Date% %Time)  _screenshot_name = "%Date% %Time%_%TimeMS%";
```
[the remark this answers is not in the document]
elbadcode: Great idea! But if you want more powerful could we make one little amendment to the macro tooltip to add %BeforeAfter% as a macro? I have updated the tooltip hash number for all languages so they still function but havent added any localizations. IMO its too useful not to include and I think it fits in well to the existing menu

image

As for the main implementation, I'm not sure if this is the most optimal handling as string manipulation in C++ is not my favorite, but at least from a functionality standpoint its working great and I think matches your intent. I added some handling to eliminate trailing and leading whitespaces and it generally does what its supposed to. The only outstanding issue is that Having `\\%BeforeAfter%\\`  in your path and then disabling the option to save before and after images leads to a double slash despite a brief attempt to handle it. However you can type an indefinite number of slashes in the directory boxes and they're ignored, so ultimately not a huge problem.  Everything aside from this code block is just some minor changes

`
void reshade::runtime::save_screenshot(const std::string_view postfix) {
  const unsigned int screenshot_count = _screenshot_count;
//instantiate reusable vector excluding count for path because that would be literally insane to want that
 std::vector<std::pair<std::string, std::string>> macros = {
      {"BeforeAfter", std::string(postfix)  },
      {"AppName", g_target_executable_path.stem().u8string()},
#if RESHADE_FX
      {"PresetName", _current_preset_path.stem().u8string()},
#endif
    };

  // Expand macros in path
  std::filesystem::path expanded_path  = std::filesystem::u8path(  expand_macro_string(_screenshot_path.u8string(), macros));


#if RESHADE_FX
 macros.push_back({"Count", std::to_string(screenshot_count)});
#endif

  // Expand macros in filename with count included
  std::string screenshot_name = expand_macro_string(_screenshot_name, macros);


  // trim leading and trailing whitespaces from the name
  screenshot_name.erase(0, screenshot_name.find_first_not_of(" \t\n\r\f\v"));
  screenshot_name.erase(screenshot_name.find_last_not_of(" \t\n\r\f\v") + 1);


screenshot_name += (_screenshot_format == 0 ? ".bmp" : _screenshot_format == 1 ? ".png" : ".jpg");

const std::filesystem::path screenshot_path = g_reshade_base_path / expanded_path  / std::filesystem::u8path(screenshot_name);

`

I'll push these commits along in just a moment but I have a few more things ready or almost ready that I'll try to catch you in discord to discuss

PR #343, opened 2025-01-22
Screenshot macros unified changes

elbadcode:
All changes now unified to name instead of path, beforeafter macro added and naming scheme changed to reflect, escape closes overlay refactored to not need option and instead just uses input processing mode (blocking input only while hovered can double press escape which is good middle ground)

I think this should all be good to go!


res/lang_en-US.rc2
```diff
@@ -108,7 +106,7 @@ BEGIN
...
+12781 "Macros you can add that are resolved during command execution:\n  %%AppName%%         Name of the application (%s)\n  %%BeforeAfter%%     Before and after rendering effects\n  %%PresetName%%      File name without extension of the current preset file (%s)\n  %%Date%%            Current date in format '%s'\n  %%DateYear%%        Year component of current date\n  %%DateMonth%%       Month component of current date\n  %%DateDay%%         Day component of current date\n  %%Time%%            Current time in format '%s'\n  %%TimeHour%%        Hour component of current time\n  %%TimeMinute%%      Minute component of current time\n  %%TimeSecond%%      Second component of current time\n  %%TimeMS%%          Milliseconds fraction of current time\n  %%TargetPath%%      Full path to the screenshot file (%s)\n  %%TargetDir%%       Full path to the screenshot directory (%s)\n  %%TargetFileName%%  File name of the screenshot file (%s)\n  %%TargetExt%%       File extension of the screenshot file (%s)\n  %%TargetName%%      File name without extension of the screenshot file (%s)\n  %%Count%%           Number of screenshots taken this session\n"
```
cot6, line 109: "%BeforeAfter%" has the following meaning "The term that describes what was rendered on the frame."

I think “Render stage at capture” or "Render stage of the frame" is more understandable.

I am not sure if these terms are correct in graphics programming.

elbadcode: I'm okay with changing the wording if there's something more clear, especially the localizations I provided with machine translation, but I think for a UI tooltip its more important to be easily understood than to be technically correct. To clarify are we good with the macro being called %BeforeAfter% and only want to change the tooltip wording? I think it makes sense given the UI already used that wording as does your screenshot addon (which could now also get rid of the "original" category I suppose). 

“Render stage at capture” or "Render stage of the frame" is perhaps more accurate but I think this becomes less understandable at least in english. Its a bit confusing to not use the exact terms before and after or direct equivalents. Actually at first glance render stage at capture and render stage of the frame sound like identical concepts to me.

Maybe something like this would be best to convey meaning
"Before: without effects, After: with effects"

cot6: Hmm, I am not very good at English, so I will trust native speakers and crosire's sense of word choice.
I think the current sentence doesn't make sense in Japanese, so I may commit it after this changes.

The "original" timing seems useless, but I think it can be helpful in an add-ons environment. (e.g. the add-on drawn something at the event before rendering effects whether effects are enabled or not, and you want to capture the original)

source/runtime.cpp
```diff
@@ -189,7 +189,10 @@ reshade::runtime::runtime(api::swapchain *swapchain, api::command_queue *graphic
 #endif
 	_config_path(config_path),
 	_screenshot_path(L".\\"),
-	_screenshot_name("%AppName% %Date% %Time%_%TimeMS%"), // Use a timestamp down to the millisecond because users may request more than one screenshot per-second
+	// this is what I've been rocking but probably too drastic a change for a default setting
+	// _screenshot_name("%AppName%\\%DateMonth%\\%BeforeAfter%\\%Count%_%PresetName%_%Time%_%TimeMS%")
+	// so here's a less drastic change that would be useful for most people to introduce new feature if desired. note that beforeafter gets ignored if they have the save_before setting off
+	_screenshot_name("%AppName%\\%BeforeAfter%\\%Date% %Time%_%TimeMS%"), // Use a timestamp down to the millisecond because users may request more than one screenshot per-second
```
crosire, line 195: Please don't change the default, this is something users can do themselves if desired.

source/runtime.hpp
```diff
@@ -356,8 +356,12 @@ namespace reshade
...
 		#pragma region Screenshot
 #if RESHADE_FX
-		bool _screenshot_save_before = false;
-		bool _screenshot_include_preset = false;
+	bool _screenshot_save_before = false;
+	// I'm writing an addon that allows for a simple scrolling gallery view of screenshots that have associated presets
+	// with the ability to preview/apply the preset, making it easier to interact with and choose a preset
+	// plan is to support preset authors bundling images with their presets as well as display your own
+	// so it would be nice to have this default to on as otherwise many users won't have data to use with the addon!
+	bool _screenshot_include_preset = true;
```
crosire, line 364: You could just change the setting in the add-on, so to not having to change the default:
```c++
reshade::set_config_value(runtime, "SCREENSHOT", "SavePresetFile", true)
```

source/runtime_gui.cpp
```diff
@@ -2309,8 +2329,8 @@ void reshade::runtime::draw_gui_settings()
 
 		modified |= imgui::file_input_box(_("Screenshot sound"), "sound.wav", _screenshot_sound_path, _file_selection_path, { L".wav" });
 		ImGui::SetItemTooltip(_("Audio file that is played when taking a screenshot."));
-
-		modified |= imgui::file_input_box(_("Post-save command"), "command.bat", _screenshot_post_save_command, _file_selection_path, { L".exe", L".bat", L".cmd", L".ps1", L".py" });
+		// forgot pyw on original pull request, that is a big one since its commonly used for windowless scripts
+		modified |= imgui::file_input_box(_("Post-save command"), "command.bat", _screenshot_post_save_command, _file_selection_path, { L".exe", L".bat", L".cmd", L".ps1", L".py", L".pyw"});
```
crosire, line 2333: I had intentionally removed `.pyw` actually, since that file extension is generally wired up to pythonw.exe (WinMain), not python.exe (console).

source/runtime.cpp
```diff
@@ -4882,21 +4892,49 @@ static std::string expand_macro_string(const std::string &input, std::vector<std
...
-			}));
+  };
+
+ // Could likely consolidate this with some inlining
+ std::string expanded_screenshot_name = expand_macro_string(_screenshot_name, macros);
+	
+ // Trim leading and trailing whitespaces from the name.
+ // Was more of an issue with old split implementation so this could stay or go
+ expanded_screenshot_name.erase(0, expanded_screenshot_name.find_first_not_of(" \t\n\r\f\v"));
+ expanded_screenshot_name.erase(expanded_screenshot_name.find_last_not_of(" \t\n\r\f\v") + 1);
```
crosire, line 4917: I have a helper function for that in ini_file.hpp which is already included, so can simply call:
```c++
expanded_screenshot_name = trim(expanded_screenshot_name);
```

source/runtime.cpp
```diff
@@ -4882,21 +4892,49 @@ static std::string expand_macro_string(const std::string &input, std::vector<std
...
+
+
+  // If save before/after option is checked but not used as a macro, append it to the filename (old behavior)
+  bool before_after_used = (expanded_screenshot_name.find(std::string(postfix)) != std::string::npos);
+  // If we want to fix tiny issue of overlay being included we'd do another check here I suppose
+  if (_screenshot_save_before && !before_after_used) {
+    expanded_screenshot_name += "_" + std::string(postfix);
```
crosire, line 4924: Technically previous behavior was to append with space, rather than underscore =P

source/runtime.cpp
```diff
@@ -4882,21 +4892,49 @@ static std::string expand_macro_string(const std::string &input, std::vector<std
...
+  }
+
+  expanded_screenshot_name += (_screenshot_format == 0 ? ".bmp" : _screenshot_format == 1 ? ".png" : ".jpg");
+
+  // Normalize path by removing duplicate slashes, handling mixed slashes, again likely unnecessary use of local var
+  std::string normalized_name = expanded_screenshot_name;
+  std::replace(normalized_name.begin(), normalized_name.end(), '/', '\\');
+  while (normalized_name.find("\\\\") != std::string::npos) {
+    normalized_name.replace(normalized_name.find("\\\\"), 2, "\\");
+  }
```
crosire, line 4934: This can be removed by taking advantage of std::filesystem::path functionality: Can just do `const std::filesystem::path screenshot_path = g_reshade_base_path / std::filesystem::u8path(normalized_name).lexically_normal()` below.

res/lang_en-US.rc2
crosire: How about a combination: `Term dependent on moment the screenshot was taken: can be "Before", "After" or "Overlay"`

crosire (commented):
The change to allow path in the screenshot name + escape key behavior make sense to me. I just have reservations about changing defaults.

source/runtime_gui.cpp
elbadcode: Generally the usecase of pyw is to suppress the terminal window for gui apps or silent scripts. However that only applies when running them from pythonw.exe. Since we're explicitly calling python.exe I believe it should be functionally identical to executing a .py

Not a big deal since one can always rename their script but I think it should be fine

source/runtime.cpp
elbadcode: Understood, only considered doing so because I noticed the recent addition of TimeMs to the default.

source/runtime.cpp
elbadcode: Whitespaces are icky 🤢 but I really have no horse in this race since I will always be using the macro

source/runtime.hpp
elbadcode: Yep already have that, but the user would still need to go take some screenshots after installing the addon to be able to use the full feature set. they can still use it to review screenshots and I could include some preset files with accompanying screenshots to demonstrate the feature, but they won't have any screenshot/preset pairs of their own right away. Then again maybe its safe to assume anyone who has presets they'd consider worthwhile already had the setting enabled. I also haven't fully implemented my addon yet so not a big deal

res/lang_en-US.rc2
elbadcode: That makes sense and is now implemented but for the tooltip should I write 

`%Before%, %After%, %Overlay%      Term dependent on moment the screenshot was taken: can be "Before", "After" or "Overlay"`
or just
`Term dependent on moment the screenshot was taken: can be "Before", "After" or "Overlay"`

the second one is more in line with the length of other tooltips but it would be the only one breaking the two column layout. I also actually included the option to type %postfix% as a shortcut for all 3 so it could be used there

`%postfix%    Term dependent on moment the screenshot was taken: can be "Before", "After" or "Overlay"`

but maybe there's a better single word term than postfix in that case

PR #345, opened 2025-02-04
Add API to export the current preset to another path

cot6:


cot6:
See my add-on side
https://github.com/cot6/reshade-addons/commit/00f14d6db4b62551b1f25fdec3f3d152695bd78b

PR #346, opened 2025-02-17
Open text links using ImGui functions

cot6:
See https://github.com/crosire/reshade/pull/329#issue-2555739833


cot6:
imgui-textlink-1
imgui-textlink-2

cot6:
Content width is specified by the user code because there is no content has width of 1.

PR #347, opened 2025-03-03
Implement MaximumRuntimes option

Yuki-Codes:
This PR implements a config option "MaximumRuntimes" that can be set in ReShade.ini to limit how many effect runtimes reshade will create. a Value of 0 (default) is unlimited runtimes.

Very useful for games that can produce secondary, non-game windows.

I'm not 100% on the naming and implementation details, this is my first PR to reshade.

### DIFF
--- a/source/runtime_manager.cpp
+++ b/source/runtime_manager.cpp
@@ -12,7 +12,6 @@
 
 static std::shared_mutex s_runtime_config_names_mutex;
 static std::unordered_set<std::string> s_runtime_config_names;
-static int s_max_num_runtimes = 0;
 
 void reshade::create_effect_runtime(api::swapchain *swapchain, api::command_queue *graphics_queue, bool is_vr)
 {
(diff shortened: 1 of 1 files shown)


crosire:
Fine idea, just the implementation is using the wrong configuration file right now. This would be a global option, rather than being specific to an effect runtime: right now you are reading the option from the config file of the first effect runtime, instead you'll want to read it from the global config file (they typically both map to ReShade.ini in the same location, but don't have to). So really this becomes a simple one-liner:
```c++
if (size_t max_runtimes = std::numeric_limits<size_t>::max();
    global_config().get("INSTALL", "MaxEffectRuntimes", max_runtimes) &&
    s_runtime_config_names.size() >= max_runtimes)
    return;
```

Yuki-Codes:
Ah, excellent, that's much cleaner. thanks!

Changes made. 👍

PR #349, opened 2025-03-13
Don't fire addon events for swap chains that do not have effect runtimes

Yuki-Codes:
This change is a bit spooky, its primary goal is to fix a crash in the [ReshadeEffectShaderToggler](https://github.com/4lex4nder/ReshadeEffectShaderToggler) addon caused by multiple windows.

What I've done is suppressed the addon events `init_swapchain` `present` `destroy_swapchain` and  `execute_command_list` for any swapchains that do not have an effect runtime, by returning a bool from the runtime_manager when its updated.

However, a possibly significant change is that the addon events now trigger _after_ the runtime_manager and I'm not sure if that will have and significant issues. I wasn't able to notice anything misbehaving in my somewhat limited testing.

If this is too scary of a change then we can discard it here and I'll try solve the crash in the addon code instead.

### DIFF
--- a/source/d3d12/d3d12_command_queue_downlevel.cpp
+++ b/source/d3d12/d3d12_command_queue_downlevel.cpp
@@ -24,11 +24,13 @@ D3D12CommandQueueDownlevel::~D3D12CommandQueueDownlevel()
 {
 	if (_back_buffers[0] != nullptr)
 	{
-		reshade::reset_effect_runtime(this);
+		if (reshade::reset_effect_runtime(this))
+		{
 
 #if RESHADE_ADDON
-		reshade::invoke_addon_event<reshade::addon_event::destroy_swapchain>(this, false);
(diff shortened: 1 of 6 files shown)


crosire:
ReShade intentionally fires events for all swapchains, since there can be valid reasons for a swapchain to not have an effect runtime associated with it (be it disabled or ...), yet still being of interest of add-ons. That's why the swapchain and effect runtime was split up into separate object instances in the first place (they were originally tied together, but no longer are).
So this crashing is indicative of a bug in REST (potentially means it needs to keep track of which swapchain the effect runtime it works with is associated with and ignore events from other swapchains), I'd rather not throw out functionality in ReShade over it.

Yuki-Codes:
ah yep, thought that might be the case. I'll go dig in to REST instead. Thanks for the feedback!

PR #350, opened 2025-03-13
get_is_loading api

Yuki-Codes:
this is a simple change that introduces a `get_is_loading()` api to the effect runtime that add-ons can use to query the loading state:
```
static void on_present(effect_runtime* runtime)
{
	s_isLoading = runtime->get_is_loading();
}
```

I'm unsure if the API level needs to be incremented for this?

### DIFF
--- a/include/reshade_api.hpp
+++ b/include/reshade_api.hpp
@@ -812,5 +812,11 @@ namespace reshade { namespace api
 		/// </summary>
 		/// <param name="path">File path to the preset to save to.</param>
 		virtual void export_current_preset(const char *path) const = 0;
+
+		/// <summary>
+		/// Gets a boolean indicating whether effects are being loaded.
+		/// </summary>
+		/// <returns><see langword="true"/> if the the effect runtime is loading effects, <see langword="false"/> otherwise.</returns>
+		virtual bool get_is_loading() const = 0;
 	};
 } }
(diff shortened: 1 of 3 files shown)


crosire:
Do you have a concrete usecase in mind for this? ReShade otherwise already takes care of not invoking certain events during loading and preventing stale handles from being used during loading, so there ideally should be no need for knowing whether ReShade is currently loading or not.

Yuki-Codes:
I actually use this to support taking screenshots at resolutions higher than the windows resolution by changing the size of the swap-chain, then waiting for reshade to compile all the effects, then taking a screenshot before setting everything back.

I just need to reliably wait for the effects to finish loading before I take my screenshot.

crosire:
You can just register for the `reshade_reloaded_effects` event for that.

Yuki-Codes:
alright, I came up with a solution that works reliably, I end up counting the number of reshade_finish_effects events my add-on receives after changing resolution and just make sure I get about 120 (2 seconds-ish at 60Hz) before saving the screenshot and setting everything back. 👍

crosire:
When a reload is triggered, `reshade_reloaded_effects` should be called exactly 2 times: The first time to indicate effects are being unloaded and then after effects were recreated. So ideally just skipping the first invocation and triggering on the second one *should* be more stable.

PR #355, opened 2025-04-24
Fix height match on depth addon

Sarenya:
Hello,

Found a tiny little bug when I tried to select a specific depth buffer.

In the depth addon when "Match custom width and height exactly" is selected, the height filter is applied to the width


crosire:
Nice catch!

PR #359, opened 2025-06-16
Wrap DXGIFactory to support vtable hooking systems

clshortfuse:
Nvidia's Smooth Motion will create vtable hooks that intercept swapchain requests from the application. By using a wrapped DXGIFactory, commands from the application will hit Reshade first, allowing Nvidia Smooth Motion to get the true swapchain format.

### DIFF
--- a/source/d3d10/d3d10.cpp
+++ b/source/d3d10/d3d10.cpp
@@ -4,10 +4,10 @@
  */
 
 #include "d3d10_device.hpp"
+#include "dxgi/dxgi_factory.hpp"
 #include "dll_log.hpp" // Include late to get 'hr_to_string' helper function
 #include "hook_manager.hpp"
 #include "addon_manager.hpp"
-#include "dxgi/dxgi_factory.hpp"
 
 extern thread_local bool g_in_dxgi_runtime;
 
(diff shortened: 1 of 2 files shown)


source/dxgi/dxgi.cpp
```diff
@@ -568,9 +31,10 @@ extern "C" HRESULT WINAPI CreateDXGIFactory(REFIID riid, void **ppFactory)
 }
 extern "C" HRESULT WINAPI CreateDXGIFactory1(REFIID riid, void **ppFactory)
 {
-	// Do NOT skip in case this is called internally for D3D10/D3D11 (otherwise the 'IDXGIFactory::CreateSwapChain' call in 'D3D10/11CreateDeviceAndSwapChain' will not be redirected)
+	// External hooks may want to create a DXGI factory and rewrite the vtables
+	if (g_in_dxgi_runtime)
+		return reshade::hooks::call(CreateDXGIFactory1)(riid, ppFactory);
```
crosire, line 36: It looks like this breaks the case where `D3D11CreateDeviceAndSwapChain` is called to create the swapchain, rather than creating only a device and a factory manually and then swapchain from that (note the comment that was here before, I made that mistake before hence putting that comment here =D).

It can probably still be made working by modifying `D3D11CreateDeviceAndSwapChain` to create a temporary instance of `DXGIFactory` with the factory object it queries and call `CreateSwapChain` on that (https://github.com/crosire/reshade/blob/main/source/d3d11/d3d11.cpp#L149).

Same thing also applies to `D3D10CreateDeviceAndSwapChain`.

source/dxgi/dxgi_factory.cpp
```diff
@@ -0,0 +1,849 @@
...
+	if (FAILED(hr))
+	{
+		reshade::log::message(reshade::log::level::warning, "IDXGIFactory::CreateSwapChain failed with error code %s.", reshade::log::hr_to_string(hr).c_str());
+		return hr;
+	}
+
+	DXGI_SWAP_CHAIN_DESC orig_desc = *pDesc;
+	if (!modified)
+		orig_desc.BufferCount = 0;
+	orig_desc.BufferDesc.Format = desc.BufferDesc.Format;
```
crosire, line 584: This line is new, is it intentional (on first glance seems weird to return the modified format, rather than the original, when this was originally designed to return the original description)?

clshortfuse: You're right, must have been from testing.

PR #360, opened 2025-06-19
Add support for Vulkan templates

clshortfuse:


crosire:
Vulkan has the nifty VK_EXT_private_data extension, which can be used to avoid lookup maps (like the `_descriptor_update_templates`) as you can just associate data with a Vulkan object directly. This is used everywhere in the ReShade Vulkan code, so that you can just define a struct with the data you want to hold, register it for an object via `device_impl->register_object<VK_OBJECT_TYPE_[...]>(handle)` and later retrieve it via `device_impl->get_private_data_for_object<VK_OBJECT_TYPE_[...]>(handle)`.

I'll just go ahead and modify the code to use that (since I had started this in parallel already anyway), otherwise looks good (I had actually missed the existence of `vkCmdPushDescriptorSetWithTemplateKHR` myself, so good thing you worked on this too).

PR #365, opened 2025-07-25
Add DXGI (and DX9) present flags to the present addon function

Filoppi:
This is required to properly add VRR support to games.

Addon developers can tell the actual flags based on the swapchain/device API.
The present interval param also needs to be added as the allow tearing flag gives an error if set when the interval isn't 0, and while ReShade already allows forcing a specific interval, ideally we'd make the whole thing dynamic and allow the game to engage v-sync if it wanted it (from its own settings), without forcing it to ever be off or on through ReShade.

A second (optional) commit further changes the API in a breaking way, but adds functionality to skip calling the original present call. This should be useful when we override the original swapchain with a new one, for games cross API presentation, e.g. DX9 to DX11, VK to DX12 etc.

### DIFF
--- a/examples/09-depth/generic_depth_addon.cpp
+++ b/examples/09-depth/generic_depth_addon.cpp
@@ -846,7 +846,7 @@ static void on_execute_secondary(command_list *cmd_list, command_list *secondary
 	}
 }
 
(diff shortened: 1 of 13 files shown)


pmnoxx:
@crosire It would be nice to have an API to skip present entirely.

PR #369, opened 2025-08-06
Add DXGIAdapter proxy to track swapchain creation

clshortfuse:
This partially reverts https://github.com/crosire/reshade/commit/4a3bd6f9a7eb8ba14cb9719accd99c11d0337c7d by implementing a DXGIAdapter proxy. This solves games such as Persona 5 Royal from not being hooked at all by Reshade.


crosire:
Only reservation I have right now is moving all the `IDXGIFactory_CreateSwapChain_*` implementations from dxgi.cpp to dxgi_swapchain.cpp. That makes total sense logically, but I think I'd rather do it separately at some point later, and also keep the vtable hooking code alive for now (just `#if 0` out). Until this has soaked for a while. And to keep the changes here are bit more straightforward.

clshortfuse:
@crosire Rebased and updated with `RESHADE_IDXGIFACTORY_VTABLE` in `dxgi_factory.hpp`. Removing the define makes it proxy only (no vtable). P5R works with either configuration.

crosire:
Thanks!

source/dxgi/dxgi_factory.cpp
```diff
@@ -127,20 +537,89 @@ HRESULT STDMETHODCALLTYPE DXGIFactory::GetWindowAssociation(HWND *pWindowHandle)
 }
 HRESULT STDMETHODCALLTYPE DXGIFactory::CreateSwapChain(IUnknown *pDevice, DXGI_SWAP_CHAIN_DESC *pDesc, IDXGISwapChain **ppSwapChain)
 {
+#ifndef RESHADE_IDXGIFACTORY_VTABLE
+	if (g_in_dxgi_runtime)
```
crosire, line 541: Realized now that this check here should not be necessary, since these proxy objects should never be used within the DXGI runtime, it's really only needed for the vtable hooks.

PR #372, opened 2025-08-19
Optimize D3D12 GPU to original CPU descriptor handle

clshortfuse:
Avoid iteration per heap. Use ordered map instead

### DIFF
--- a/source/d3d12/d3d12_impl_device.cpp
+++ b/source/d3d12/d3d12_impl_device.cpp
@@ -2213,6 +2213,14 @@ void reshade::d3d12::device_impl::register_descriptor_heap(D3D12DescriptorHeap *
 	const auto it = _descriptor_heaps.push_back(heap);
 
 	heap->initialize_descriptor_base_handle(std::distance(_descriptor_heaps.begin(), it));
+#if RESHADE_ADDON >= 2
+	D3D12_DESCRIPTOR_HEAP_DESC desc = heap->_orig->GetDesc();
+
+	auto base_gpu = heap->_orig_base_gpu_handle.ptr;
+	auto end_gpu = base_gpu + desc.NumDescriptors * _descriptor_handle_size[desc.Type];
+	std::unique_lock<std::shared_mutex> lock(_heap_gpu_ranges_mutex);
+	_heap_gpu_ranges[base_gpu] = { end_gpu, heap };
+#endif
 }
 void reshade::d3d12::device_impl::unregister_descriptor_heap(D3D12DescriptorHeap *heap)
 {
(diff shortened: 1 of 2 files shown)


crosire:
Awesome!
I'm a bit concerned about multi-threading though: if an app would create a new descriptor heap, while submitting commands with a different descriptor heap (which based on https://github.com/crosire/reshade/commit/bfd8b10f29bbaf399d0d0c21a7c63527e7954d08 I did encounter in the past), this might crash due to race between reallocation/usage of the map. And unfortunately just adding a mutex lock here isn't an option (as per https://github.com/crosire/reshade/commit/200281d4172840d179e0aef210f6e749b68394fd, hence the concurrent vector for the heap list). Not sure how to best solve that, maybe an atomic + spin lock.

clshortfuse:
You're right. I somehow missed that. You need a mutex in this case. Though, using an ordered `map` will keep the locks pretty short once you've find the item (and `--it` called) as opposed to iteration. I'd suggest a mutex just for the value instead of using the mutex used by all resources (creating and removing a resource view shouldn't affect any related to heaps).

Personally, I've moved all the map systems to <https://github.com/greg7mdp/parallel-hashmap> (though the C++20 version). Seem it'll add mutexes for the ordered map though via submaps and have the benefit of threading. It defaults to `shared_mutex` which is SRWLOCK in C++17 and above.

clshortfuse:
I updated with a lock. I noticed two different spin locks in greg's library:

```cpp
#include <atomic>
class spinlock {
    std::atomic_flag flag = ATOMIC_FLAG_INIT;
public:
    void lock()   { while(flag.test_and_set(std::memory_order_acquire)); }
    void unlock() { flag.clear(std::memory_order_release); }
};
```
https://github.com/greg7mdp/parallel-hashmap/blob/def20389e8143be4d86a0e299c603311bae754f6/examples/bench.cc#L43-L50C29

```cpp
class spinlock_mutex {
   // https://rigtorp.se/spinlock/
   // https://vorbrodt.blog/2019/02/12/fast-mutex/
public:
   void lock() noexcept {
      for (;;) {
         if (!lock_.exchange(true, std::memory_order_acquire))
            break;
         while (lock_.load(std::memory_order_relaxed))
            __builtin_ia32_pause();
      }
   }
   void unlock() noexcept {
      lock_.store(false, std::memory_order_release);
   }
private:
   alignas(4 * sizeof(std::max_align_t)) std::atomic_bool lock_ = false;
};
```
https://github.com/greg7mdp/parallel-hashmap/blob/def20389e8143be4d86a0e299c603311bae754f6/examples/llil4map.cc#L92-L109C3

Though, I think SRWLock checks an atomic flag. Comparison to spinlock is relevant when it does lock, which would only be on descriptor heap creation, which should be rare; and if and when it does fire I'd assume would be in loading areas.

crosire:
Moving the `_descriptor_heaps` access within the mutex still kills performance in Far Cry 6, but this change alone does not, so merging.

PR #377, opened 2025-08-23
Add input blocking API (mouse/keyboard) + “Input Control” example addon

pmnoxx:
Summary: Introduces an addon API to block forwarding mouse and keyboard input from ReShade to the host application, plus an example addon demonstrating usage.
API surface:
C API exports:
ReShadeBlockMouseInput(reshade::api::effect_runtime *runtime, bool enable)
ReShadeBlockKeyboardInput(reshade::api::effect_runtime *runtime, bool enable)
C++ convenience wrappers in include/reshade.hpp:
reshade::block_mouse_input(reshade::api::effect_runtime *runtime, bool enable)
reshade::block_keyboard_input(reshade::api::effect_runtime *runtime, bool enable)
Behavior:
When enabled for a given effect_runtime, ReShade consumes the specified input class and does not forward it to the application.
Default behavior is unchanged (blocking is off). Addons must opt-in.
Implementation:
Wires the new API through runtime and GUI integration so the block state is applied each frame.
Touches: include/reshade.hpp, source/addon.cpp, source/runtime_api.cpp, source/runtime_gui.cpp.
Example addon:
New project examples/18-input_control with ImGui toggles to block mouse and/or keyboard input at runtime.
Includes a short README describing usage.
Compatibility:
No breaking changes; blocking is disabled by default and scoped to the provided effect_runtime.
Testing:
Verified toggling from the example addon correctly prevents game/app from receiving the chosen input while ReShade continues to function normally.
Notes:
The GUI already manages mouse cursor warping separately; this PR focuses on explicit mouse/keyboard blocking for addons.


pmnoxx:
> From a code perspective this is fine now!
> 
> I do still wonder if the separation into mouse/keyboard/warping is really necessary however. I fixed `block_input_next_frame()` so that it can block all input even when the overlay is closed (in [d93ee3c](https://github.com/crosire/reshade/commit/d93ee3cdee889789d49aac0e36d368784e33a6e4)). Could you try replacing
> 
> ```c++
> runtime->block_mouse_input(block_mouse);
> runtime->block_keyboard_input(block_keyboard);
> runtime->block_mouse_cursor_warping(block_warp);
> ```
> 
> with just
> 
> ```c++
> if (block_input)
>   runtime->block_input_next_frame();
> ```
> 
> and check how that fares? Seems like it could be sufficient. Unless there is a good argument for e.g. blocking keyboard, but not mouse input, which I couldn't think of?

block_mouse_input and block_mouse_cursor_warping can be combined together, I don't see any reason to separate them.

I like the default behavior of Special-K, where keyboard is disabled when game is in background, but mouse is enabled always.
This always you to see the game cursor, even when you are alt-tabbed.

However, in some games this isn't desirable, because moving mouse can cause undesired sound effects.

In summary, default settings for Special-K are:
Mouse - enabled always
Keyboard - disabled in background
Gamepad - enabled always (special-k doesn't support disabling gamepad, but that's ok). Some people want to have gamepad enabled, so they can play games on one monitor, while using mouse keyboard on another. Having mouse enabled, won't do anything if cursor is on another monitor.

pmnoxx:
My proposal would be to add following api:


 		/// <summary>
 		/// Flags to control input suppression behavior for <see cref="suppress_input"/>.
 		/// </summary>
 		enum input_block_flags : uint32_t
 		{
 			/// <summary>
 			/// Do not block any input.
 			/// </summary>
 			input_block_none = 0x00000000,
 			/// <summary>
 			/// Block all mouse input from reaching the application.
 			/// </summary>
 			input_block_mouse = 0x00000001,
 			/// <summary>
 			/// Block all keyboard input from reaching the application.
 			/// </summary>
 			input_block_keyboard = 0x00000002,
 			/// <summary>
 			/// Reserved for future use to block gamepad input.
 			/// </summary>
 			input_block_gamepad_reserved = 0x00000004,
 			/// <summary>
 			/// When blocking mouse input, still allow mouse move events to reach the application.
 			/// </summary>
 			input_block_exception_mouse_move = 0x00000008,
 		};
		void suppress_input(uint32_t flags) final;

crosire:
> I like the default behavior of Special-K, where keyboard is disabled when game is in background, but mouse is enabled always.
This allows you to see the game cursor, even when you are alt-tabbed.
In this case block_mouse_cursor_warping is blocked always, but mouse_input such as position isn't blocked, but mouse_input such as clicking is blocked always.

ReShade cannot currently block mouse clicking and mouse moving independently either (cursor warping is just preventing the game from forcefully resetting the cursor to the screen center): when mouse input is blocked, all mouse input is blocked. So this behavior is not achievable even when keyboard/mouse blocking were to be separated in the API. With these changes right now you could make it so keyboard input is blocked in the background, but have the cursor still move and click/interact, which seems undesirable. That doesn't really convince me of the advantage of splitting these out I'm afraid.

pmnoxx:
> > I like the default behavior of Special-K, where keyboard is disabled when game is in background, but mouse is enabled always.
> > This allows you to see the game cursor, even when you are alt-tabbed.
> > In this case block_mouse_cursor_warping is blocked always, but mouse_input such as position isn't blocked, but mouse_input such as clicking is blocked always.
> 
> ReShade cannot currently block mouse clicking and mouse moving independently either (cursor warping is just preventing the game from forcefully resetting the cursor to the screen center): when mouse input is blocked, all mouse input is blocked. So this behavior is not achievable even when keyboard/mouse blocking were to be separated in the API. With these changes right now you could make it so keyboard input is blocked in the background, but have the cursor still move and click/interact, which seems undesirable. That doesn't really convince me of the advantage of splitting these out I'm afraid.

I can add support for this in this PR.

pmnoxx:
> > I like the default behavior of Special-K, where keyboard is disabled when game is in background, but mouse is enabled always.
> > This allows you to see the game cursor, even when you are alt-tabbed.
> > In this case block_mouse_cursor_warping is blocked always, but mouse_input such as position isn't blocked, but mouse_input such as clicking is blocked always.
> 
> ReShade cannot currently block mouse clicking and mouse moving independently either (cursor warping is just preventing the game from forcefully resetting the cursor to the screen center): when mouse input is blocked, all mouse input is blocked. So this behavior is not achievable even when keyboard/mouse blocking were to be separated in the API. With these changes right now you could make it so keyboard input is blocked in the background, but have the cursor still move and click/interact, which seems undesirable. That doesn't really convince me of the advantage of splitting these out I'm afraid.

I think runtime->block_input_next_frame() is good enough for now.

I think it may be better design wise to add functionality in separate PR to subscribe to Mouse/Keyboard events.

bool OnKeyboardEvent(...)
bool OnMouseEvent(...)

This would be more generic solution allowing to achieve, what we talked about. And also add support for custom keyboard shortcuts.

pmnoxx:
Thanks, I'll test out block_input_next_frame api.

crosire:
I'm wary of adding event invocation to the input handling code itself, due to it being the hot path. You can however do custom keyboard shortcuts by way of polling the current input state already (which is also how ReShade handles shortcuts everywhere):
```c++
if (runtime->is_key_pressed(VK_SPACE) && runtime->is_key_down(VK_CONTROL))
   do_action_for_ctrl_plus_space_shortcut();
```

pmnoxx:
Thanks, block_input_next_frame() works. I'll close this PR.

is_key_pressed / is_key_down should be sufficient. I'll try those out.

PR #378, opened 2025-08-23
Feature/hide hdr

pmnoxx:
## Summary
This PR adds a new feature to hide HDR capabilities from D3D11 applications, preventing them from detecting and enabling HDR rendering.


crosire:
This doesn't really fit the ReShade API design principles as-is, which is to expose graphics API actions as events and let add-ons deal with them as they desire, rather than exposing explicit functionality. Plus an effect runtime is not a global object, and as such should not really contain API that performs global actions (the other functions in `effect_runtime` only affect that particular effect runtime instance, of which there can be multiple, or the window associated with that effect runtime instance).

One could make a case for adding a new event like `addon_event::check_color_space_support` though, which would then e.g. be invoked when the game calls `IDXGISwapChain::CheckColorSpaceSupport`, giving add-ons the opportunity to override support behavior.

pmnoxx:
Good idea, I'll rewrite the PR.

PR #380, opened 2025-08-25
Add support for latency markers

pmnoxx:
Fps limiters such as Reflex / Special-K FPS  limited require latency markers to work properly. This PR adds support for latency markers to reshade to allow for development of more advanced fps limiters as reshade addons.

Currently fps limiters such as renodx/fpslimiter don't have access to those latency markers.

### DIFF
--- a/source/d3d11/d3d11_device_context.cpp
+++ b/source/d3d11/d3d11_device_context.cpp
@@ -847,10 +847,26 @@ void    STDMETHODCALLTYPE D3D11DeviceContext::ExecuteCommandList(ID3D11CommandLi
 #if RESHADE_ADDON
 	// Use secondary command list event here, since this may be called on deferred contexts as well
 	reshade::invoke_addon_event<reshade::addon_event::execute_secondary_command_list>(this, command_list_proxy);
+
+	// Emit RENDERSUBMIT_START/END for immediate context execution path. For deferred contexts this represents submission to immediate.
+	const bool is_immediate = (this == _device->_immediate_context);
+	if (is_immediate)
+	{
+		const uint64_t ts_begin = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::high_resolution_clock::now().time_since_epoch()).count();
+		reshade::invoke_addon_event<reshade::addon_event::latency_rendersubmit_start>(this, nullptr, ts_begin);
+	}
 #endif
 
 	// Get original command list pointer from proxy object and execute with it
 	_orig->ExecuteCommandList(command_list_proxy->_orig, RestoreContextState);
+
+#if RESHADE_ADDON
+	if (is_immediate)
+	{
+		const uint64_t ts_end = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::high_resolution_clock::now().time_since_epoch()).count();
+		reshade::invoke_addon_event<reshade::addon_event::latency_rendersubmit_end>(this, nullptr, ts_end);
+	}
+#endif
 }
 void    STDMETHODCALLTYPE D3D11DeviceContext::HSSetShaderResources(UINT StartSlot, UINT NumViews, ID3D11ShaderResourceView *const *ppShaderResourceViews)
 {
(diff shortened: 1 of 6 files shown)


source/dxgi/dxgi_swapchain.cpp
```diff
@@ -297,6 +298,29 @@ HRESULT STDMETHODCALLTYPE DXGISwapChain::GetDevice(REFIID riid, void **ppDevice)
...
+			reshade::invoke_addon_event<reshade::addon_event::latency_present_start>(
+				static_cast<D3D11Device *>(static_cast<ID3D11Device *>(_direct3d_device))->_immediate_context,
+				_impl,
+				ts_ns);
+			break;
+		case reshade::api::device_api::d3d12:
+			reshade::invoke_addon_event<reshade::addon_event::latency_present_start>(
+				static_cast<D3D12CommandQueue *>(_direct3d_command_queue),
+				_impl,
+				ts_ns);
```
crosire, line 319: This add-on is doing the same thing as `reshade::addon_event::present`, so can also just use that instead of introducing an additional one.

source/dxgi/dxgi_swapchain.cpp
```diff
@@ -313,6 +337,30 @@ HRESULT STDMETHODCALLTYPE DXGISwapChain::Present(UINT SyncInterval, UINT Flags)
 	const HRESULT hr = _orig->Present(SyncInterval, Flags);
 	g_in_dxgi_runtime = false;
 
+#if RESHADE_ADDON
+	// Emit PRESENT_END latency marker right after present returns
+	if (_is_initialized)
+	{
+		const uint64_t ts_ns = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::high_resolution_clock::now().time_since_epoch()).count();
```
crosire, line 344: Add-ons can query a timestamp in the event callback(s), not much of a benefit to ReShade doing this.

source/d3d12/d3d12_command_queue.cpp
```diff
@@ -168,6 +169,14 @@ void    STDMETHODCALLTYPE D3D12CommandQueue::ExecuteCommandLists(UINT NumCommand
 	// Synchronize access to this command queue while events are invoked and the immediate command list may be accessed
 	std::unique_lock<std::recursive_mutex> lock(_mutex);
 
+#if RESHADE_ADDON
+	// Emit RENDERSUBMIT_START marker at the beginning of a batch submission
```
crosire, line 173: This is not correct: Applications do submit multiple times per frame, and also to multiple different command queues. The `RENDERSUBMIT_START/END` markers however are supposed to be called only once per frame, before all the submissions and after all the submissions. This can't easily be mapped to just hooking `ExecuteCommandLists`.

source/d3d11/d3d11_device_context.cpp
```diff
@@ -847,10 +847,26 @@ void    STDMETHODCALLTYPE D3D11DeviceContext::ExecuteCommandList(ID3D11CommandLi
...
 	// Use secondary command list event here, since this may be called on deferred contexts as well
 	reshade::invoke_addon_event<reshade::addon_event::execute_secondary_command_list>(this, command_list_proxy);
+
+	// Emit RENDERSUBMIT_START/END for immediate context execution path. For deferred contexts this represents submission to immediate.
+	const bool is_immediate = (this == _device->_immediate_context);
+	if (is_immediate)
+	{
+		const uint64_t ts_begin = std::chrono::duration_cast<std::chrono::nanoseconds>(std::chrono::high_resolution_clock::now().time_since_epoch()).count();
+		reshade::invoke_addon_event<reshade::addon_event::latency_rendersubmit_start>(this, nullptr, ts_begin);
+	}
```
crosire, line 857: Command list execution is not a very common occurence in D3D11 titles, and in those that make use of it, will be called multiple times per frame with various command lists. As in D3D12, this is not the right place to issue the `RENDERSUBMIT_START/END` marker every time.

crosire (commented):
Reflex integration is unfortunately more complicated than hooking graphics API calls (also hence why it's a SDK and can't just be done by the driver beyond the NVIDIA Ultra Low Latency driver mode, which is effectively looking at the graphics API command and present calls to try and get as close as possible).

PR #398, opened 2025-11-07
Handle pid 0 case for register_window function

tdlbugreport:
Added a check for zero process ID and make a warning instead directly giving up.


tdlbugreport:
> In what case did you experience a PID of zero occuring here?

mw:19 with iw8-mod installed.

PR #414, opened 2026-04-23
vulkan: Compatibility adjustments. 

MohannedElfatih:
- Clear Image format list after addons events and add mutable format bit.
- Fix framebuffer creation to exclude unused bindings, ensuring smaller attachment counts.
- Enable api::primitive_topology::triangle_fan pipelines (Fixes No man's Sky UI bugs)

AI generated but reviewed logic.

### DIFF
--- a/source/vulkan/vulkan_hooks_device.cpp
+++ b/source/vulkan/vulkan_hooks_device.cpp
@@ -24,13 +24,6 @@ extern lockfree_linear_map<void *, vulkan_instance, 16> g_vulkan_instances;
 lockfree_linear_map<void *, reshade::vulkan::device_impl *, 8> g_vulkan_devices;
 
 #if RESHADE_ADDON
-void clear_image_format_list(const void *pNext)
-{
-	if (const auto format_list_info = find_in_structure_chain<VkImageFormatListCreateInfo>(
-			pNext, VK_STRUCTURE_TYPE_IMAGE_FORMAT_LIST_CREATE_INFO))
-		const_cast<VkImageFormatListCreateInfo *>(format_list_info)->viewFormatCount = 0;
-}
-
 void create_default_view(reshade::vulkan::device_impl *device_impl, VkImage image)
 {
 	if (image == VK_NULL_HANDLE)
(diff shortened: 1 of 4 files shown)


source/vulkan/vulkan_hooks_device.cpp
```diff
@@ -1206,16 +1214,16 @@ VkResult VKAPI_CALL vkCreateImage(VkDevice device, const VkImageCreateInfo *pCre
 	if (reshade::invoke_addon_event<reshade::addon_event::create_resource>(device_impl, desc, nullptr, pCreateInfo->initialLayout == VK_IMAGE_LAYOUT_PREINITIALIZED ? reshade::api::resource_usage::cpu_access : reshade::api::resource_usage::undefined))
 	{
 		reshade::vulkan::convert_resource_desc(desc, create_info);
+		create_info.flags |= VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT;
```
crosire: This is intended to be triggered by add-ons by setting the format to a typeless variant (to behave like D3D11/D3D12, which also use typeless format for resources and then typed format for views). When done this way, `convert_resource_desc` above will already automatically add the mutable flag: https://github.com/crosire/reshade/blob/main/source/vulkan/vulkan_impl_type_convert.cpp#L1344. So this shouldn't be necessary.

For an example see also https://github.com/crosire/reshade/blob/main/examples/09-depth/generic_depth_addon.cpp#L601 + https://github.com/crosire/reshade/blob/main/examples/09-depth/generic_depth_addon.cpp#L637 (it's scoped to just D3D10/11 there, because I implemented a separate method for D3D12/Vulkan, but the code would otherwise also work this way in D3D12/Vulkan).

source/vulkan/vulkan_hooks_device.cpp
```diff
@@ -1269,6 +1277,10 @@ VkResult VKAPI_CALL vkCreateImageView(VkDevice device, const VkImageViewCreateIn
 	if (reshade::invoke_addon_event<reshade::addon_event::create_resource_view>(device_impl, reshade::api::resource { (uint64_t)create_info.image }, reshade::api::resource_usage::undefined, desc))
 	{
 		reshade::vulkan::convert_resource_view_desc(desc, create_info);
+
+		if (auto resource_data = device_impl->get_private_data_for_object<VK_OBJECT_TYPE_IMAGE, true>(create_info.image))
+			resource_data->create_info.flags |= VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT;
```
crosire: I don't understand the need for this, the image has already been created at this point, so adding the flag here doesn't serve a purpose?

source/vulkan/vulkan_hooks_swapchain.cpp
```diff
@@ -95,6 +96,11 @@ VkResult VKAPI_CALL vkCreateSwapchainKHR(VkDevice device, const VkSwapchainCreat
 		}
 	}
 
+#if RESHADE_ADDON
+	if (reshade::has_addon_event<reshade::addon_event::create_resource_view>())
+		clear_image_format_list(create_info.pNext);
+#endif
```
crosire: This also doesn't make sense to me. I do understand the need for it after the `create_swapchain` below (though should ideally also use the same logic that was in vkCreateImage to only clear format list when format was actually overriden), but not this one?

crosire (commented):
Framebuffer and triangle fan changes look good, I just have issues with the mutable format flag changes:

MohannedElfatih:
Removed unnecessary code, and gated it behind VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT in createImage hook.

source/vulkan/vulkan_hooks_device.cpp
```diff
@@ -1206,16 +1213,9 @@ VkResult VKAPI_CALL vkCreateImage(VkDevice device, const VkImageCreateInfo *pCre
 	if (reshade::invoke_addon_event<reshade::addon_event::create_resource>(device_impl, desc, nullptr, pCreateInfo->initialLayout == VK_IMAGE_LAYOUT_PREINITIALIZED ? reshade::api::resource_usage::cpu_access : reshade::api::resource_usage::undefined))
 	{
 		reshade::vulkan::convert_resource_desc(desc, create_info);
+		if ((create_info.flags & VK_IMAGE_CREATE_MUTABLE_FORMAT_BIT) != 0)
+			clear_image_format_list(create_info.pNext);
```
crosire: I'm still not clear on what this does better over the code removed below. That already cleared the format list when it was no longer valid because the format was overriden, so should solve the same thing?

MohannedElfatih: Initially I added it just to unify clear_image_format_list logic. Functionality is the same.

Now we only have it in two places so I'll revert it

MohannedElfatih:
Removed clear_image_format_list

PR #415, opened 2026-04-23
Modify pipeline and layout creation to preserve original createInfo if modified by addons

MohannedElfatih:
Any modifications to pipelines/layouts can cause unexpected errors because it forces it to go through internal functions instead of trampoline. I attempted to modify on top of the original createInfo instead of throwing it all together.

The changes are too extensive (& AI generated obviously) but it's more as a proof of concept to increase reshade's robustness. Please feel free to close it if it doesn't fit reshade's intended approach.


crosire:
I'm generally not opposed to something like this, similar stuff is already done elsewhere (e.g. D3D12 root signature creation gets original flags passed through in since they don't translate well). Need to disect the code a bit though =P

source/vulkan/vulkan_impl_device.cpp
```diff
@@ -1238,21 +1239,57 @@ void reshade::vulkan::device_impl::update_texture_region(const api::subresource_
...
-		spec_map.push_back(VkSpecializationMapEntry { desc.spec_constant_ids[i], i * 4, sizeof(uint32_t) });
+	const auto *base = static_cast<const VkBaseInStructure *>(pnext);
+	while (base != nullptr)
+	{
+		if (base->sType == VK_STRUCTURE_TYPE_SHADER_MODULE_CREATE_INFO)
+		{
+			// Stripping an inline module create info from the middle of the chain would require cloning
+			// the prefix nodes, so drop the preserved chain instead of passing an invalid combination of
+			// explicit module handle plus inline module source to Vulkan.
+			return base == pnext ? base->pNext : nullptr;
```
crosire, line 1252: I thinking that it could be easier if we avoid shader module creation entirely and always pass along the code via module create info in the structure chain now (this used to not be possible when I had first written the code, but has been a thing for a while now)? This would avoid some of the ugly clean up code and also don't have to worry about rewriting the existing chain, since just appending new module create info to the end should make that become the one used (since in Vulkan the last structure of a type in the structure chain wins).

source/vulkan/vulkan_impl_device.cpp
```diff
@@ -1734,22 +2009,23 @@ bool reshade::vulkan::device_impl::create_pipeline(api::pipeline_layout layout,
...
-		VkPipelineViewportStateCreateInfo viewport_state_info { VK_STRUCTURE_TYPE_PIPELINE_VIEWPORT_STATE_CREATE_INFO };
+		VkPipelineViewportStateCreateInfo viewport_state_info = create_info.pViewportState != nullptr ? *create_info.pViewportState : VkPipelineViewportStateCreateInfo { VK_STRUCTURE_TYPE_PIPELINE_VIEWPORT_STATE_CREATE_INFO };
 		create_info.pViewportState = &viewport_state_info;
 		viewport_state_info.scissorCount = viewport_count;
 		viewport_state_info.viewportCount = viewport_count;
 
-		VkPipelineRasterizationStateCreateInfo rasterization_state_info { VK_STRUCTURE_TYPE_PIPELINE_RASTERIZATION_STATE_CREATE_INFO };
+		VkPipelineRasterizationStateCreateInfo rasterization_state_info = create_info.pRasterizationState != nullptr ? *create_info.pRasterizationState : VkPipelineRasterizationStateCreateInfo { VK_STRUCTURE_TYPE_PIPELINE_RASTERIZATION_STATE_CREATE_INFO };
+		rasterization_state_info.pNext = nullptr;
 		create_info.pRasterizationState = &rasterization_state_info;
```
crosire, line 2029: For some state create info, `pNext` is cleared, for some not. Is there a reason for this?

crosire (commented):
I've fixed the merge conflicts with top of tree and moved the code around a bit so that the changes become easier to see in the diff. Functionally it should be the same still (only removed the pipeline cache parameter, since I'd expect modified pipelines to not end up in the original cache).

PR #422, opened 2026-05-28
Fix compilation error

mxtherfxcker:
Fixed a compilation error related to **type conversion**

Edited: This error prevented the build of `ReShade64.dll`

### DIFF
--- a/source/effect_codegen_dxil.cpp
+++ b/source/effect_codegen_dxil.cpp
@@ -67,7 +67,7 @@ public:
 
 		com_ptr<IDxcResult> result;
 
-		HRESULT hr = compiler->Compile(&hlsl_buffer, arguments, std::size(arguments), nullptr, IID_PPV_ARGS(&result));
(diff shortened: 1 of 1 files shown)


crosire:
What compiler are you building with? MSVC does not consider this a narrowing cast because it's constant and can be resolved at compile, so there is neither a warning, nor an error.

mxtherfxcker:
> What compiler are you building with? MSVC does not consider this a narrowing cast because it's constant and can be resolved at compile, so there is neither a warning, nor an error.

MSVC consider this warning as an error.

crosire:
Which version? Because MSVC shipping with VS2022+ don't, see also successful GitHub Actions builds: https://github.com/crosire/reshade/actions/runs/26671484044/job/78615435121

mxtherfxcker:
> Which version? Because MSVC shipping with VS2022+ don't, see also successful GitHub Actions builds: https://github.com/crosire/reshade/actions/runs/26671484044/job/78615435121

vs2017 and vs2019

crosire:
Ah, haven't used those in a while. There are more changes necessary to make things work for those again though (e.g. jxl not building), will apply them all in one go.

PR #423, opened 2026-06-04
MSVC: Enable Unwind Semantics

LordNyriox:
Fixes warning [C4530](https://learn.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-1-c4530?view=msvc-170):

```
warning C4530: C++ exception handler used, but unwind semantics are not enabled. Specify /EHsc [D:\a\reshade\reshade\ReShade.vcxproj]
```


crosire:
These are intentionally not enabled in release builds (see also ```<ExceptionHandling>false</ExceptionHandling>```).

PR #427, opened 2026-06-16
Unify create_pipeline overloads switch statements

MohannedElfatih:
create_pipeline vkGraphicsCreateInfo missed max_vertex_count switch case. Instead of just adding it, I opted to centralize the switch statements to avoid future bugs.


crosire:
I've changed it so the `create_pipeline` without orig_create_info just forwards always now instead, including for ray tracing pipelines, and doesn't parse the subobject list (wasn't a fan of capturing and passing around another big struct of create info): https://github.com/crosire/reshade/commit/bc6fc73e5511977c4310924ec1b261ffdf68a002#diff-93ced83fd0c6a44d39165fdc823c39ec1566b98f8e6d66233f1352722c4e4c15

PR #428, opened 2026-06-26
Fix missing return value in on_end_render_pass

HYPER-THEORY:
I tried to build the examples in VS2022, but the compilation failed. One reason is that the on_end_render_pass function in 13-effects_during_frame is missing a return value.


crosire:
Merged as part of #429.

PR #431, opened 2026-07-13
Add option to set the depth-buffer-clear position counting from the last one

microdee:
This PR gives a tool for Source games to help with correctly copying the depth-buffer there. There are well known problems with reliably getting the depth buffer in Source games:

* Rendering the GUI clears the depth buffer
* The same depth-buffer is used several times to render 3D skybox or render-targets
* The statistical methods may incorrectly determine to copy the 3D skybox content if both draw-call-count and vertex count are higher than the main scene.

To get around all that I introduced the option to manually set the copy-depth-before-clear position counting from the last clear operation, since the count of the UI depth-buffer clears are much more stable than the depth-buffer-clears done during actual rendering.

Depth buffer without this fix:

https://github.com/user-attachments/assets/ce15f491-467e-4c08-8dff-1adf21a47d3a

Depth buffer with this fix:

https://github.com/user-attachments/assets/b3e38c6f-0d47-4d88-a772-d0d6a71629ad

Another example:

https://github.com/user-attachments/assets/7cbea17f-e249-4ee1-8404-b3c433376b57

Unfortunately since we rely on data from previous frame (the number of depth-buffer-clears in one frame) there may be flickering if the clear operations change often, but in Source games at least it's barely noticeable.

### DIFF
--- a/examples/09-depth/generic_depth_addon.cpp
+++ b/examples/09-depth/generic_depth_addon.cpp
@@ -944,6 +944,12 @@ static void on_present(command_queue *, swapchain *swapchain, const rect *, cons
 			info.last_frame_stats = frame_stats;
 			info.last_used_in_frame = device_data->frame_index;
 
+			// Save the last clear count on depth_stencil_backup
+			if (depth_stencil_backup *const depth_stencil_backup = device_data->find_depth_stencil_backup(depth_stencil))
+			{
+				depth_stencil_backup->last_clear_count = static_cast<uint32_t>(info.last_frame_stats.clears.size());
+			}
+
 			if (std::numeric_limits<uint64_t>::max() == info.first_used_in_frame)
 				info.first_used_in_frame = device_data->frame_index;
 		}
@@ -1390,9 +1396,8 @@ static void draw_settings_overlay(effect_runtime *runtime)
 			if (depth_stencil_backup == nullptr || depth_stencil_backup->backup_texture == 0)
 				continue;
 
-			depth_stencil_backup->last_clear_count = static_cast<uint32_t>(info.last_frame_stats.clears.size());
-
-			for (uint32_t clear_index = 1; clear_index <= depth_stencil_backup->last_clear_count; ++clear_index)
+			uint32_t last_clear_count = static_cast<uint32_t>(info.last_frame_stats.clears.size());
+			for (uint32_t clear_index = 1; clear_index <= last_clear_count; ++clear_index)
 			{
 				const clear_stats &clear_stats = info.last_frame_stats.clears[clear_index - 1];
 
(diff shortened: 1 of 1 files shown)


crosire:
This Source Engine case is theoretically already supposed to be addressed by the "Choose last clear operation with high number of draw calls" option (see also b1effa70646e7a63a8f89ca9cd1bfa30acc9b734). However I see in the videos that the option is not shown here, presumably because the condition for showing it is too restrictive (`sorted_item_list.size() == 1`, but there are two items here). Could you try by removing that `sorted_item_list.size() == 1` condition and enabling the option, to see if it has the same effect? Based on the statistics I'm seeing in the second window, I suspect it should.

microdee:
Oh! I'm going to try with that commit, thanks for pointing it out, do you remember what was the reason to restrict "Choose last clear operation with high number of draw calls" like that in the first place?

microdee:
While it's indeed more stable with "Choose last clear operation with high number of draw calls" it can still fail in circumstances when the 3D skybox is a lot busier than the actual foreground:

Screenshot 2026-08-02 191008

Whereas counting from last clear solves this situation as well

Screenshot 2026-08-02 191119

The results are the same if I check "Copy depth buffer before fullscreen draw calls"

crosire:
Makes sense, I guess this is fine to merge.

There is just one issue I spotted: Existing configs will have "DepthCopyAtClearIndex" with value 4294967295 (aka UINT_MAX) and the setup tool also currently sets that value for some games (e.g. https://github.com/crosire/reshade-shaders/blob/list/Compatibility.ini#L517). This breaks now, because the code here changed that magic value to `std::numeric_limits<int32_t>::max()` and also, `get_config_value` is called with a signed integer, so parsing of that value fails completely.
This could be solved by changing
```c++
if (s_preserve_depth_buffers)
	reshade::get_config_value(nullptr, "DEPTH", "DepthCopyAtClearIndex", depth_stencil_backup->force_clear_index);
else
	depth_stencil_backup->force_clear_index = 0;
```
to handle this case explicitly like so:
```c++
char clear_index_string[32]; size_t clear_index_string_length = sizeof(clear_index_string) - 1;
if (!s_preserve_depth_buffers || !reshade::get_config_value(nullptr, "DEPTH", "DepthCopyAtClearIndex", clear_index_string, &clear_index_string_length))
	depth_stencil_backup->force_clear_index = 0;
else if (std::strcmp(clear_index_string, "4294967295") == 0)
	depth_stencil_backup->force_clear_index = std::numeric_limits<int32_t>::max();
else
	std::from_chars(clear_index_string, clear_index_string + clear_index_string_length, depth_stencil_backup->force_clear_index);
```

microdee:
how about reading it as int64 and clamping the result to the range of int32?

```Cpp
				if (s_preserve_depth_buffers)
				{
					int64_t clear_index_raw = 0;
					reshade::get_config_value(nullptr, "DEPTH", "DepthCopyAtClearIndex", clear_index_raw);
					depth_stencil_backup->force_clear_index = static_cast<int32_t>(std::clamp(
						clear_index_raw,
						static_cast<int64_t>(std::numeric_limits<int32_t>::min()),
						static_cast<int64_t>(std::numeric_limits<int32_t>::max())
					));
				}
				else
					depth_stencil_backup->force_clear_index = 0;
```

PR #436, opened 2026-09-04
fix(png): add option to exclude cICP chunk from HDR PNG screenshots

Jahbanny:
## Problem

Discord fails to render HDR PNG screenshots that contain a cICP chunk. This is a bug in Discord's PNG decoder, not a problem with ReShade's output.

## Approach

Instead of permanently removing the cICP chunk (which breaks other viewers that depend on it, such as Apple's image viewer), this PR:

- **Keeps cICP by default** — matching the original ReShade behavior
- **Adds an advanced checkbox** in Settings > Output to opt out for users hitting the Discord bug

## Changes

| File | Change |
|---|---|
| `deps/stb_image/stb_image_write_hdr_png.h` | Added `include_cicp` parameter; cICP only written when enabled; fixed `file_size` calculation |
| `source/runtime.hpp` | Added `bool _screenshot_include_cicp = true;` |
| `source/runtime.cpp` | Config load/save + pass flag to PNG writer |
| `source/runtime_gui.cpp` | Checkbox with tooltip: "Uncheck to fix image display issues in Discord" |


Jahbanny:
﻿Good call. I reworked the approach: cICP is now kept by default (matching original behavior) and added an advanced checkbox in the GUI to opt out for Discord compatibility. The ICC profile data from the original PR is still in the file but unused - happy to remove it if you prefer.

Jahbanny:
Checked (Defaulted behavior)

image

Jahbanny:
Unchecked (Workaround for discord bug)

image

crosire:
This feels like a thing that should be done with ReShade's screenshot post-command feature, rather than part of core ReShade. Considering it's a Discord bug, not a ReShade bug. Just need a simple tool/script that adjusts the png, enter that as a post-command, and it's applied automatically to every screenshot. (The current implementation here will also throw asserts when using any language other than English.)

Jahbanny:
> This feels like a thing that should be done with ReShade's screenshot post-command feature, rather than part of core ReShade. Considering it's a Discord bug, not a ReShade bug. Just need a simple tool/script that adjusts the png, enter that as a post-command, and it's applied automatically to every screenshot. (The current implementation here will also throw asserts when using any language other than English.)

Makes sense. I will close the PR and maybe just do a separate addon for RHI.

PR #437, opened 2026-09-05
dxgi: serialize D3D11 immediate context access during swap chain creation

elzuzu:
Disclosure up front: the investigation, the instrumentation and this patch were done by an AI assistant (Claude Fable 5.1, through Claude Code) running on my machine; I provided the game, the hardware and the decisions, and I ran all the measurements below myself, so I can reproduce anything on request. If AI-assisted contributions are not wanted here, feel free to close this, no hard feelings.

Mass Effect: Andromeda (Frostbite, D3D11) loses its D3D11 device in roughly one launch out of four as soon as ReShade is loaded, showing the game's own "DirectX Error … DXGI_ERROR_DEVICE_REMOVED" dialog a few seconds after start. Without ReShade it never fails. The failure does not depend on add-ons (ReShade alone reproduces it), on the proxy name (`dxgi.dll` or `d3d11.dll`), on NVIDIA Smooth Motion, on Windows Auto HDR, on the windowed swap effect upgrade, and it survives a reboot.

An instrumented build (`GetDeviceRemovedReason` logged after creation and on failure, a 1 ms watchdog, and optionally `D3D11_CREATE_DEVICE_DEBUG` with the info queue streamed to the log) shows the device is alive right after `D3D11CreateDevice` and gets removed inside `IDXGIFactory::CreateSwapChain` (6 ms into the call in one capture) or on the first resource the effect runtime creates right after it, with `DXGI_ERROR_DEVICE_REMOVED` or `DXGI_ERROR_DRIVER_INTERNAL_ERROR`. With the debug layer on, the runtime emits a CORRUPTION message about 450 ms after device creation: `ID3D11DeviceContext::GetData: Two threads were found to be executing functions associated with the same Device[Context] at the same time.` The game polls queries on the immediate context from a worker thread while its main thread is still creating the swap chain; on its own that stays harmless in practice, but ReShade initializes its runtime on the same immediate context during swap chain creation, and the NVIDIA driver then corrupts and removes the device.

This change turns on `ID3D11Multithread` protection on the immediate context only for the duration of the swap chain creation (the DXGI call plus the runtime initialization it triggers) and restores the previous state afterwards, on the four creation paths. Applications that already enable the protection are untouched, the others pay the critical section for one call only. A variant that enabled the protection permanently at device creation fixed the crash as well, but this scoped version keeps the per-call cost unchanged after start-up for games that never enable it.

This deliberately mirrors `unique_direct3d_device_lock` in `source/dxgi/dxgi_swapchain.cpp`, which already does the same `SetMultithreadProtected` save/restore dance on the present and resize paths. I did not factor the two together because that one also takes a `std::recursive_mutex` and reaches through the `D3D10Device`/`D3D11Device` proxy types, neither of which exists yet at the point where the swap chain is being created. If you would rather have a single shared helper, or a different placement entirely, say the word and I will rework it.

Measured on the same machine, back-to-back launches: 8 device removals in 36 launches without the patch (across all the configurations listed above), 0 in 24 launches with the permanent variant, 0 in 24 launches with this scoped version. Windows 11 26200, GeForce RTX 5070, driver 616.64, ReShade 6.8.0.2155 add-on build, Mass Effect: Andromeda 1.10 (EA app), 3440x1440 borderless. Logs from the instrumented runs are available if useful.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


crosire:
That's a great root cause analysis, the implementation of the fix isn't optimal though. Will push a proper fix taking advantage of the existing lock mechanism, which also extends to D3D12.

elzuzu:
Confirming your fix on the original repro (Mass Effect: Andromeda, RTX 5070, driver 616.64, ReShade add-on build), 24 launches each, same harness:

- `main` @ a33de92 — **0 device removals / 24**
- released 6.8.0 — **7 / 24**

Thanks for picking it up.

PR #438, opened 2026-09-09
Drop [fastopt] in favor of [loop] always.

draeath:
Running in proton, shader compilation fails if [fastopt] is used. Since
the shader model check is insufficient in this environment, just drop
it. Apparently on modern GPUs, the performance difference is negligable?
It apparently just affects compilation speed.


crosire:
`fastopt` was intentionally used to reduce compile time with d3dcompiler (where this can make a dramatic difference). I'm a bit hesitant to move away from it again and cause an invisible regression. Ideally Wine would just not hard error when encountering it (considering it only warns for other attributes it doesn't recognize, instead of failing) .....: https://gitlab.winehq.org/wine/vkd3d/-/blob/master/libs/vkd3d-shader/hlsl.y?ref_type=heads#L751

draeath:
Is that just a compile time difference? How dramatic is dramatic, roughly speaking?

Am I wrong in thinking that, except when a user is actively changing their settings or changed their hardware, compilation is done once and is kept in cache? If we're talking about a 1-5 second initialization once in a while vs 1-5 seconds every launch, that's two different stories for sure.

I'm thinking getting it changed on the Wine side would be an uphill battle, and wouldn't help anyone using a currently-existing proton release.

Can this be made dynamic at all, perhaps with preprocessor defs? I'm largely talking out of my rear when it comes to C/C++, but I imagine two different code paths could be gated by such a means of configuration.

crosire:
Eh, let's just make it loop. If effects start failing to compile due to d3dcompiler optimizer choking, can always revert.

PR #439, opened 2026-09-09
Fix D3D12 descriptor heap destruction callback lifetime

Dragon31337:
## Description

Keep the D3D12 device proxy alive until the native descriptor heap destruction callback executes. Capture the original GPU descriptor handle by value so the callback does not dereference a released heap proxy, and release the native heap before deleting its proxy.

Without this, a descriptor heap retained beyond the final device proxy release can invoke its callback after the device implementation and descriptor maps have been destroyed. The callback then accesses the destroyed map, potentially causing process heap corruption during shutdown.

This was observed when Streamline shutdown released a remaining D3D12 descriptor heap after ReShade device teardown.

## Testing

- Built ReShade x64 Release successfully:
  `MSBuild ReShade.sln /m /t:ReShade /p:Configuration=Release /p:Platform=64-bit`
- Verified that the change applies cleanly to current upstream `main`.


crosire:
The main problem is really that the destruction callback mechanism is triggered for the original heap object, not the proxy heap object. So I think the cleaner solution is to not rely on that at all here and instead trigger registration/unregistration explicitly in the `D3D12DescriptorHeap` constructor/destructor, to ensure it's called at the right time. I'll go ahead and restructure the code to that end.

crosire:
Should be fixed now.

